PR #49, opened 2020-06-26
[RAPTOR-2904] Add flag and start dummy server to propagate imports/pipeline failures

rvorobii:

## Rationale
In server mode, we should have an option to have the server up even when errors occur on model loading etc.
This is required so that k8s doesn't restart container as it won't help.
Instead, the server should allow getting the errors via API.

As we've discussed, fail points are
1. Imports
2. Model loading
There was an idea to check imports and model loading before running drum.

However, imports and model loading is coupled in drum: depending on model artifacts present predictors are selected, and each predictor has its own set of imports.
Additionally, having model loading _before_ drum leads to logic/code duplication.

To me, it's better to wrap _everything_ with error handling logic, that can start "error server" in case of exceptions.

Benefit 1:
we don't have to gather all imports in one place to see if we can load modules - just run drum as normal.
During pipeline initialisation phase: predictor is selected, predictor load model, all the components are traversed to `configure` them.
When we wrap all the pipeline initialisation - we won't miss any import + we cover initialisation off all pipeline components with error handling logic.

Benefit 2:
The solution is simple
- drum logic is not remains the same.
- it's easy to see what code paths are affected.

## Summary
1. Add flag `--force-start-internal`
2. Wrap all the code with error handling logic (`with DrumRuntime() as runtime: ...`)
3. Pass `DrumRuntime` to `CMRunner` 
3.1. When pipeline initialisation is done - set `initialization_succeeded` in `DrumRuntime`.
3.2. When exception occurs, check if dummy error server should be started.
It should be started if pipeline initialisation hasn't succeeded (`initialization_succeeded` is False) and `--force-start-internal` was passed when starting drum.
3.2.1. Common code of dummy error server and prediction server is separated into a separate blueprint
4. Error server propagates errors via `/predict/` route, to keep the current protocol.
To distinguish between failed pipeline initialisation and other "predict" failures (like missing pred dataset), 503 Service Unavailable is used to define pipeline initialisation errors. 
5. Tests
5.1. Refactor tests: create context manager for starting drum server
5.2. Add unit and e2e tests for `DrumRuntime`

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/main.py
+++ b/custom_model_runner/datarobot_drum/drum/main.py
@@ -29,40 +29,46 @@ Examples:
     # Run regression user model in fit mode.
     drum fit --code-dir <custom code dir> --input <input.csv> --output <output_dir> --target <target feature> --verbose
 """
-
-import argcomplete
-import os
-import sys
-import signal
-import requests
-from datarobot_drum.drum.args_parser import CMRunnerArgsRegistry
-from datarobot_drum.drum.drum import CMRunner
+from datarobot_drum.drum.common import DrumContext
+from datarobot_drum.drum.error_handling import DrumErrorHandler
 
 
 def main():
-    options = None
+    ctx = DrumContext()
+    with DrumErrorHandler(ctx):
+        import argcomplete
+        import os
+        import sys
+        import signal
+        import requests
+        from datarobot_drum.drum.args_parser import CMRunnerArgsRegistry
+        from datarobot_drum.drum.drum import CMRunner
 
-    def signal_handler(sig, frame):
-        # The signal is assigned so the stacktrace is not presented when Ctrl-C is pressed.
-        # The cleanup itself is done only if we are NOT running in performance test mode which
-        # has its own cleanup
-        print("\nCtrl+C pressed, aborting drum")
+        arg_parser = CMRunnerArgsRegistry.get_arg_parser()
+        # argcomplete call should be as close to the beginning as possible
+        argcomplete.autocomplete(arg_parser)
+        options = arg_parser.parse_args()
+        CMRunnerArgsRegistry.verify_options(options)
 
-        if options and options.docker and not options.in_perf_mode_internal:
-            url = "http://{}/shutdown/".format(options.address)
-            print("Sending shutdown to server: {}".format(url))
-            requests.post(url, timeout=2)
-        os.system("tput init")
-        os._exit(130)
+        ctx.options = options
 
-    signal.signal(signal.SIGINT, signal_handler)
+        options = None
 
-    arg_parser = CMRunnerArgsRegistry.get_arg_parser()
-    # argcomplete call should be as close to the beginning as possible
-    argcomplete.autocomplete(arg_parser)
-    options = arg_parser.parse_args()
-    CMRunnerArgsRegistry.verify_options(options)
-    CMRunner(options).run()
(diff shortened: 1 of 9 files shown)


tests/drum/test_custom_model.py
```diff
@@ -63,6 +75,78 @@
...
+        self, framework, problem, language, docker=None, custom_model_dir=None, force_start=False,
+    ):
+        self._custom_model_dir_to_remove = None
+        if custom_model_dir is None:
+            # If custom model directory is not set explicitly - it's is created for the class.
+            # In such case it should also be removed.
+            custom_model_dir = TestCMRunner._create_custom_model_dir(framework, problem, language)
+            self._custom_model_dir_to_remove = custom_model_dir
+
+        port = 6799
```
yakov-g, line 100: Maybe we should use `CMRunnerUtils.find_free_port()` here?

custom_model_runner/datarobot_drum/drum/main.py
```diff
@@ -29,38 +29,52 @@
...
+            # The cleanup itself is done only if we are NOT running in performance test mode which
+            # has its own cleanup
+            print("Ctrl+C pressed, aborting cmrun")
+
+            if options.docker and not options.in_perf_mode_internal:
+                url = "http://{}/shutdown/".format(options.address)
+                print("Sending shutdown to server: {}".format(url))
+                requests.post(url, timeout=2)
+            os.system("tput init")
+            sys.exit(0)
```
yakov-g, line 74: @4dahalibut changed this place, rebase on master

tests/drum/test_custom_model.py
rvorobii: I'll leave this as is for this PR - it works and has been working for quite some time.

rvorobii: we can improve this separately

custom_model_runner/datarobot_drum/drum/main.py
rvorobii: done

rvorobii:
Performed manual testing, drum is working as expected (both standalone and from within DR execution environment).
After PR is merged, several follow-ups will follow:
1. Change status code used by "dummy error server" from 503 to something else - currently 503 signals issues with pods, and DR retries prediction requests in such case.
2. Add support for starting "dummy error server" in Docker
3. Refactor tests to simplify working with temp files and directories

I'll merge PR after it passes review.

custom_model_runner/datarobot_drum/resource/components/Python/prediction_server/prediction_server.py
```diff
@@ -115,27 +115,8 @@ def stats():
...
-            if func is None:
-                raise RuntimeError("Not running with the Werkzeug Server")
-            func()
-
-        @app.route("{}/shutdown/".format(url_prefix), methods=["POST"])
-        def shutdown():
-            _shutdown_server()
-            return "Server shutting down...", HTTP_200_OK
+        app = get_flask_app(model_api)
+        logging.getLogger("werkzeug").setLevel(logger.getEffectiveLevel())
```
yakov-g, line 119: Eventually, looks like, this affects only on printing log messages in terminal.
Like: > received, response 200
you know what i mean...

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**7.36** USD.

PR #59, opened 2020-07-08
[RAPTOR-3025] Alpha Testing Feedback

4dahalibut:

## Summary
* Default to full row amount so as not to confuse people
* Randomly sampling from CSV rows for when people choose to use a subset
* Improve Ctrl-C Behavior
* Improve messages
* Fix error if target column is non-existent.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/drum.py
+++ b/custom_model_runner/datarobot_drum/drum/drum.py
@@ -187,11 +187,13 @@ class CMRunner(object):
                 self.run_test_predict()
             if remove_temp_output:
                 print(
-                    "Validation Complete: Your model can be fit to your data, "
-                    "and predictions can be made on the fit model! "
+                    "Validation Complete 🎉 Your model can be fit to your data, "
+                    "and predictions can be made on the fit model! \n"
                     "You're ready to add it to DataRobot. "
                 )
                 shutil.rmtree(remove_temp_output)
+            else:
+                print("Success 🎉")
         elif self.run_mode == RunMode.PERF_TEST:
             CMRunTests(self.options, self.run_mode).performance_test()
         elif self.run_mode == RunMode.VALIDATION:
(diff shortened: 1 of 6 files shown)


custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -187,11 +187,13 @@ def run(self):
                 self.run_test_predict()
             if remove_temp_output:
                 print(
-                    "Validation Complete: Your model can be fit to your data, "
-                    "and predictions can be made on the fit model! "
+                    "Validation Complete 🎉 Your model can be fit to your data, "
```
yakov-g, line 190: Is the `tada` emoji intentional? :)

custom_model_runner/datarobot_drum/drum/main.py
```diff
@@ -40,26 +40,28 @@
...
+        print("\nCtrl+C pressed, aborting drum")
 
-        if options.docker and not options.in_perf_mode_internal:
+        if options and options.docker and not options.in_perf_mode_internal:
             url = "http://{}/shutdown/".format(options.address)
             print("Sending shutdown to server: {}".format(url))
             requests.post(url, timeout=2)
         os.system("tput init")
-        sys.exit(0)
+        os._exit(130)
```
yakov-g, line 56: Why doing this?

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -269,8 +268,13 @@ def fit(self, X, y, output_dir, class_order=None, row_weights=None):
...
         else:
+            hooks = [
+                "{}: {}".format(hook, fn is not None) for hook, fn in self._custom_hooks.items()
+            ]
             raise DrumCommonException(
-                "fit() method must be implemented in the 'custom.py' in the provided code_dir: '{}'".format(
-                    self._model_dir
+                "\nfit() method must be implemented in a file named 'custom.py' in the provided code_dir: '{}' \n"
+                "Here is a list of files in this dir. {}\n"
+                "Here are the hooks your model has: {}".format(
```
yakov-g, line 277: Maybe better: `Here are the hooks your code implements`
as I understand, we can not use word `model` for training, (for the same reason we switched form `model_dir` to `code_dir`)

custom_model_runner/datarobot_drum/drum/drum.py
4dahalibut: no it just wandered in there

custom_model_runner/datarobot_drum/drum/main.py
4dahalibut: sys exit throws an exception, which we don't want in signal handlers. 130 is the interrupt exit code.

custom_model_runner/datarobot_drum/drum/model_adapter.py
4dahalibut: I thought about that and chose this way on purpose, I think its good for people to see the full range of hooks. I can fix the word model in a followup if thats all I need to change.

yakov-g: List of hooks is ok. I'm talking only about `model` word. And anyway you'll need to rebase on master I think?

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -187,11 +187,13 @@ def run(self):
...
+                    "and predictions can be made on the fit model! \n"
                     "You're ready to add it to DataRobot. "
                 )
                 shutil.rmtree(remove_temp_output)
+            else:
+                print("Success 🎉")
```
yakov-g, line 196: And `tada` is here as well.

4dahalibut: I like :tada:

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**2.69** USD.

PR #65, opened 2020-07-10
[RAPTOR-3088] Use custom status code to propagate drum pipeline initialization failure

rvorobii:

## Rationale & Summary
503 status was used to signal that drum server initialization failed.
However, 503 is also used to signal issues with pods. In such a case, DR retries requests to drum server (as it expects drum server to be restarted and working).

To signal drum server initialization failure, we need a dedicated http status.
I used a custom one for this purpose - 513.

Here's draft PR how this logic will be used on DR side: https://github.com/datarobot/DataRobot/pull/65690

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/runtime.py
+++ b/custom_model_runner/datarobot_drum/drum/runtime.py
@@ -1,7 +1,7 @@
 from datarobot_drum.drum.server import (
     base_api_blueprint,
     get_flask_app,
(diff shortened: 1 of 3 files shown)


rvorobii:
@liororama @meltzg the PR is ready for review

@klichukb you might also want to have a look

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**1.55** USD.

PR #69, opened 2020-07-13
Functional test now uses client for custom training models

4dahalibut:


datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**3.34** USD.

PR #70, opened 2020-07-13
[MODEL-4201] New Simple Custom Training Models Hook And Test

4dahalibut:

## Rationale
Zach wants the simplest possible setup for demos. 

## Summary
Created a weird function thing which will modify an object on import so that it is detectable by drum.

### DIFF
--- a/custom_model_runner/datarobot_drum/__init__.py
+++ b/custom_model_runner/datarobot_drum/__init__.py
@@ -1,1 +1,1 @@
-from .drum.custom_fit_wrapper import custom
+from .drum.custom_fit_wrapper import drum_autofit
--- a/custom_model_runner/datarobot_drum/__init__.py
+++ b/custom_model_runner/datarobot_drum/__init__.py
@@ -1,0 +1,1 @@
+from .drum.custom_fit_wrapper import custom
--- a/custom_model_runner/datarobot_drum/drum/custom_fit_wrapper.py
+++ b/custom_model_runner/datarobot_drum/drum/custom_fit_wrapper.py
@@ -1,0 +1,9 @@
+def custom(estimator):
+    try:
+        from sklearn.base import BaseEstimator
+    except ImportError:
+        raise ValueError("You don't have scikit-learn installed, so you can't use this hook")
+    if not isinstance(estimator, BaseEstimator):
+        raise ValueError("The object passed in does not inherit from BaseEstimator")
+    estimator.is_custom = True
+    return estimator
--- a/custom_model_runner/datarobot_drum/drum/custom_fit_wrapper.py
+++ b/custom_model_runner/datarobot_drum/drum/custom_fit_wrapper.py
@@ -1,9 +1,13 @@
-def custom(estimator):
+MAGIC_MARKER = "__drum_auto_fit__"
+
+
+def drum_autofit(estimator):
     try:
         from sklearn.base import BaseEstimator
     except ImportError:
         raise ValueError("You don't have scikit-learn installed, so you can't use this hook")
     if not isinstance(estimator, BaseEstimator):
         raise ValueError("The object passed in does not inherit from BaseEstimator")
+    setattr(estimator, MAGIC_MARKER, True)
     estimator.is_custom = True
     return estimator
--- a/custom_model_runner/datarobot_drum/drum/custom_fit_wrapper.py
+++ b/custom_model_runner/datarobot_drum/drum/custom_fit_wrapper.py
@@ -9,5 +9,4 @@ def drum_autofit(estimator):
     if not isinstance(estimator, BaseEstimator):
         raise ValueError("The object passed in does not inherit from BaseEstimator")
     setattr(estimator, MAGIC_MARKER, True)
-    estimator.is_custom = True
     return estimator
--- a/custom_model_runner/datarobot_drum/drum/model_adapter.py
+++ b/custom_model_runner/datarobot_drum/drum/model_adapter.py
@@ -1,27 +1,29 @@
+import glob
+import logging
 import os
+import pickle
+import sys
+import textwrap
+from pathlib import Path
 
 import numpy as np
 import pandas as pd
-import logging
-import sys
-import textwrap
-
-from datarobot_drum.drum.common import (
-    CustomHooks,
-    CUSTOM_FILE_NAME,
-    LOGGER_NAME_PREFIX,
-    REGRESSION_PRED_COLUMN,
-    POSITIVE_CLASS_LABEL_ARG_KEYWORD,
-    NEGATIVE_CLASS_LABEL_ARG_KEYWORD,
-)
-
-from datarobot_drum.drum.exceptions import DrumCommonException
+import sklearn
 
 from datarobot_drum.drum.artifact_predictors.keras_predictor import KerasPredictor
+from datarobot_drum.drum.artifact_predictors.pmml_predictor import PMMLPredictor
 from datarobot_drum.drum.artifact_predictors.sklearn_predictor import SKLearnPredictor
 from datarobot_drum.drum.artifact_predictors.torch_predictor import PyTorchPredictor
-from datarobot_drum.drum.artifact_predictors.pmml_predictor import PMMLPredictor
 from datarobot_drum.drum.artifact_predictors.xgboost_predictor import XGBoostPredictor
+from datarobot_drum.drum.common import (
+    CUSTOM_FILE_NAME,
+    CustomHooks,
+    LOGGER_NAME_PREFIX,
+    NEGATIVE_CLASS_LABEL_ARG_KEYWORD,
+    POSITIVE_CLASS_LABEL_ARG_KEYWORD,
+    REGRESSION_PRED_COLUMN,
+)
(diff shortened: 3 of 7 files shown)


custom_model_runner/datarobot_drum/__init__.py
```diff
@@ -0,0 +1 @@
+from .drum.custom_fit_wrapper import custom
```
yakov-g, line 1: As you mentioned, it really looks weird, but maybe we could clean it a bit.

You are doing this to be able to do: `from datarobot_drum import custom`, right? So a couple of thoughts:

1. As I can recall, importing smth by default in __init__ is not a very good practice.
2. custom() method is not a very descriptive name; imagine how you are going to write documentation.

I'd suggest remove this import from here and:
Create a package and a module to be able to do: `from datarobot_drum.training_support.sklern my_magic_wrapper for estimator`
I think this could look nice and user friendly - we provide some helpers to make it easier to use drum and custom models in general.

tests/drum/test_custom_model.py
```diff
@@ -144,34 +137,15 @@ def setup_class(cls):
         cls.tests_fixtures_path = os.path.join(cls.tests_root_path, "fixtures")
         cls.tests_artifacts_path = os.path.join(cls.tests_fixtures_path, "drop_in_model_artifacts")
         cls.tests_data_path = os.path.join(cls.tests_root_path, "testdata")
-        cls.model_templates_path = os.path.join(cls.tests_root_path, "..", "model_templates")
+        cls.training_templates_path = os.path.join(
+            cls.tests_root_path, "..", "model_templates", "training"
+        )
 
         cls.paths_to_real_models = {
```
yakov-g, line 144: Please rename this to `path_to_training_templates`

tests/drum/test_custom_model.py
```diff
@@ -303,13 +277,6 @@ def _get_artifact_filename(cls, framework, problem):
 
     @classmethod
     def _get_template_dir(cls, language, framework, is_training):
```
yakov-g, line 279: Remove unused `is_training` from here

custom_model_runner/datarobot_drum/__init__.py
4dahalibut: 1. This is actually pythonic https://mikegrouchy.com/blog/be-pythonic-__init__py its pretty much the only thing that goes into the __init__.py file
2. I think we should defer naming to a third party. I think that requiring CFDSes in demos from writing that all out is pretty terrible UX, but I accept it could optimally be slightly more verbose than what I have. I passed what I have by Zach M and he said it looked good.

4dahalibut: Actually I think I'd like to keep it as `custom` because its short and sweet.

yakov-g: 1. Maybe here it's ok. I was referring to how it is done in datarobot-client, where __init__ in the root package is used to import anything, and another package can not be installed into the `datarobot` namespace.
2. custom() name doesn't make sense. It doesn't reflect any action what this function does. We are going to have it in examples - this will confuse CFDSs. Once demoed this will confuse customers. If we think no one reads our docs, we want our code do be self descriptive. That's why, going back to the item 1, I prefer imports to be done explicitly using absolute path.

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -265,11 +267,68 @@ def predict(self, data, model=None, **kwargs):
 
         return predictions
 
+    def try_to_fit_on_custom_wrapper(self, X, y, output_dir):
+        """
+        A user can surround an sklearn pipeline or estimator with the custom() function, importable
+        from drum, which will tag the object that is passed in with a to_upload variable.
```
yakov-g, line 273: Probably not `to_upload` variable, but `is_custom`?

4dahalibut: everything in the file is custom.

custom_model_runner/datarobot_drum/__init__.py
4dahalibut: what about to_upload()

custom_model_runner/datarobot_drum/drum/model_adapter.py
yakov-g: Quote: `which will tag the object that is passed in with a to_upload variable.`

What is **to_upload** variable?

4dahalibut: the to_upload variable is tagged by the custom() function which lets drum know to upload this estimator instead of another one in the file.

yakov-g: Where is this variable in the code?

4dahalibut: ah right, I meant is_custom

4dahalibut: on third thought, I kinda want to change it back.

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -265,11 +267,68 @@ def predict(self, data, model=None, **kwargs):
...
+                module = __import__(filename[:-3])
+            except ImportError as e:
+                self._logger.warning(
+                    "File at path {} could not be imported: {}".format(filepath, str(e))
+                )
+                continue
+            for object_name in dir(module):
+                _object = getattr(module, object_name)
+                if isinstance(_object, sklearn.base.BaseEstimator):
+                    if hasattr(_object, "is_custom"):
```
yakov-g, line 307: If we use `hasattr(_object, "is_custom"):` can we do it symmetrical when we set it:
Not use:  `estimator.is_custom = True`, but `estimator(object, "is_custom", True)`.
And then define `is_custom` as a constant and choose safer name, like `__drum_is_custom__`

4dahalibut: sure

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -265,11 +268,68 @@ def predict(self, data, model=None, **kwargs):
 
         return predictions
 
+    def try_to_fit_on_custom_wrapper(self, X, y, output_dir):
```
yakov-g, line 271: And just the last effort...
rename this to: `_drum_autofit`

4dahalibut: why

yakov-g: 1. because `try_to_fit_on_custom_wrapper` name doesn't reflect the actions done inside...`_drum_autofit_internal` would be even better
2. I'd also rename `wrapped_object` to `marked_object`
3. replace `custom()` in the comment with `drum_autofit()`
4. names `drum_autofit()` and `_drum_autofit_internal` form together a pair, so it is easy to understand they are used together.

custom_model_runner/datarobot_drum/drum/custom_fit_wrapper.py
```diff
@@ -0,0 +1,13 @@
...
+
+def drum_autofit(estimator):
+    try:
+        from sklearn.base import BaseEstimator
+    except ImportError:
+        raise ValueError("You don't have scikit-learn installed, so you can't use this hook")
+    if not isinstance(estimator, BaseEstimator):
+        raise ValueError("The object passed in does not inherit from BaseEstimator")
+    setattr(estimator, MAGIC_MARKER, True)
+    estimator.is_custom = True
```
yakov-g, line 12: remove this line

model_templates/training/simple/my_first_pipeline.py
```diff
@@ -0,0 +1,23 @@
...
+            "imputer",
+            SimpleImputer(strategy="median", add_indicator=True),
+            make_column_selector(dtype_include=np.number),
+        )
+    ]
+)
+pipeline = Pipeline(steps=[("numeric", numeric_transformer), ("model", Ridge())])
+
+
+# The custom function will tag the pipeline object so that DRUM knows that this object
```
yakov-g, line 21: change `custom` to `drum_autofit`

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -265,11 +268,68 @@ def predict(self, data, model=None, **kwargs):
...
+                _object = getattr(module, object_name)
+                if isinstance(_object, sklearn.base.BaseEstimator):
+                    if hasattr(_object, MAGIC_MARKER):
+                        wrapped_object = _object
+                        break
+
+        if wrapped_object is not None:
+            wrapped_object.fit(X, y)
+            output_dir_path = Path(output_dir)
+            if output_dir_path.exists() and output_dir_path.is_dir():
```
yakov-g, line 315: This check is not needed; we check this in the very beginning

4dahalibut:

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**6.45** USD.

PR #73, opened 2020-07-15
[RAPTOR-3088] Hide error server output (similarly to pred server output)

rvorobii:

## Rationale
This PR hides error server output
```
 * Serving Flask app "datarobot_drum.drum.server" (lazy loading)
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off

```
similarly for how it's done for prediction server.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/runtime.py
+++ b/custom_model_runner/datarobot_drum/drum/runtime.py
@@ -1,5 +1,3 @@
-import logging
-
 from datarobot_drum.drum.server import (
     base_api_blueprint,
     get_flask_app,
@@ -57,8 +55,6 @@ class DrumRuntime:
 
 
 def run_error_server(host, port, exc_value):
-    logging.getLogger("werkzeug").setLevel(logging.ERROR)
-
     model_api = base_api_blueprint()
 
     @model_api.route("/predict/", methods=["POST"])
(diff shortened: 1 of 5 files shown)


custom_model_runner/datarobot_drum/drum/runtime.py
```diff
@@ -48,12 +50,15 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
...
 
-        run_error_server(host, port, exc_value)
+        with verbose_stdout(self.options.verbose):
+            run_error_server(host, port, exc_value)
 
         return False  # propagate exception further
 
 
 def run_error_server(host, port, exc_value):
+    logging.getLogger("werkzeug").setLevel(logging.ERROR)
```
meltzg, line 60: I'm not sure we should set it to error. If the user uses logger for their own logs, this'll hide them

rvorobii: You are right.
For drum pipeline user can specify whatever log level is wanted.

So keep default INFO log level here won't harm I guess.

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**1.61** USD.

PR #74, opened 2020-07-16
[DOC-2145] update docs

yakov-g:

## Summary
More docs clean ups.
Mostly training models related.

I have one concern - I don't know what to do with Custom Model Templates section, naming seems wrong to me.
These are not really templates, but  examples.

Maybe we should add smth like this: 
These are examples which you can use as templates for your models
or
These are examples which you can base your models on.

The same regarding environments.
@nate-goudreault  @meltzg @4dahalibut  what do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ The model folder must contain a `custom.py` file which defines a `fit` method.
 
 
 ## Custom Model Templates <a name="custom_model_templates"></a>
-The [model templates](model_templates) folder provides templates for building and deploying custom models in DataRobot.
+The [model templates](model_templates) folder provides templates for building and deploying custom models in DataRobot. Use the templates as an example structure for your own custom models.
 
 ### DataRobot User Model Runner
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -125,13 +128,27 @@ The **drum** tool currently supports models with DataRobot-generated Scoring Cod
...
+    - `output_dir` is the path to write the model artifact to.
+    - `kwargs` additional keyword arguments to the method;
+        - `class_order: List[str]` a two element long list dictating the order of classes which should be used for modeling.
+        - `row_weights: np.ndarray` an array of non-negative numeric values which can be used to dictate how important a row is.
+
+> Note: Training and inference hooks can be defined in the same file.
+
+
+## Custom Model Templates <a name="custom_model_templates"></a>
+The [model templates](model_templates) folder provides templates for building and deploying custom models in DataRobot.
```
nate-goudreault, line 151:
```suggestion
The [model templates](model_templates) folder provides templates for building and deploying custom models in DataRobot. Use the templates as an example structure for your own custom models.
```

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**2.03** USD.

PR #75, opened 2020-07-16
Publish drum v1.1.3, switch staging envs to use drum v1.1.3

yakov-g:

## Summary
Switching envs to test with drum v1.1.3rc1, 
**_If everything is ok, I'll publish drum v 1.1.3 and update deps in this PR_**
Then we can update Staging envs.


rvorobii:
@yakov-g can we also start drum server with `--force-start-internal` for all environments in scope of this PR?

With this flag, if error occurs on model loading drum will start "dummy" server. 
The "dummy" server exposes "/" and "predict/" routes. The error is propagated via "predict/" with 513 status.

The logic to handle this 513 response on DataRobot side is already both on staging and prod.

Once we update all environments to start drum server with `--force-start-internal` - we'll get faster error response.
For instance, in case of absent model artifact - we'll get error feedback in 1-2min instead of 10min.
Additionally, we'll get a nice error displaying with the new CM testing UI:
image

yakov-g:
jarvist test functional_drop_in

yakov-g:

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**5.83** USD.

PR #77, opened 2020-07-17
[RAPTOR-2091] Raise exceptions instead of exiting program so that DrumRuntime can do post-processing

rvorobii:


datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**0.4** USD.

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**2.16** USD.

PR #79, opened 2020-07-17
[RAPTOR-3153] install requirements file from the jarvis triggered script

yakov-g:


datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**1.71** USD.

PR #81, opened 2020-07-20
[RAPTOR-3132] custom.py files can be found in nested directories

4dahalibut:

## Summary

Sometimes users like messing up in this way, so instead of failing, we'll let them mess up and adapt.

### DIFF
--- a/tests/drum/test_custom_model.py
+++ b/tests/drum/test_custom_model.py
@@ -251,8 +251,10 @@ class TestCMRunner:
 
     @classmethod
     def _create_custom_model_dir(
-        cls, custom_model_dir, framework, problem, language, is_training=False
+        cls, custom_model_dir, framework, problem, language, is_training=False, nested=False
     ):
+        if nested:
+            custom_model_dir = custom_model_dir.joinpath("nested_dir")
         custom_model_dir.mkdir(parents=True, exist_ok=True)
         if is_training:
             model_template_dir = cls.paths_to_training_models[(language, framework)]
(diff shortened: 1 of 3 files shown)


tests/drum/test_custom_model.py
```diff
@@ -480,30 +482,32 @@ def wait_for_server(url, timeout, process_holder):
...
+            (XGB, BINARY, PYTHON, None, False),
+            (PYTORCH, REGRESSION, PYTHON, None, False),
+            (PYTORCH, BINARY, PYTHON, None, False),
+            (RDS, REGRESSION, R, None, True),
+            (RDS, BINARY, R, None, False),
+            (CODEGEN, REGRESSION, NO_CUSTOM, None, False),
+            (CODEGEN, BINARY, NO_CUSTOM, None, False),
+            (MULTI_ARTIFACT, REGRESSION, PYTHON_LOAD_MODEL, None, False),
+            (PYPMML, REGRESSION, NO_CUSTOM, None, True),
+            (PYPMML, BINARY, NO_CUSTOM, None, False),
```
meltzg, line 501: what's the logic for what is False and what is True?

4dahalibut: it should work for the true ones. I'm kinda just letting it spin, but I don't think nested dirs is gon work for java and R for now...

4dahalibut:
@yakov-g I just wanted to toss that in here so that people could run inference against vizai models without having to muck with pandas.

yakov-g:
> Is dataset change intentional?

> @yakov-g I just wanted to toss that in here so that people could run inference against vizai models without having to muck with pandas.

But what is the problem? I can run inference

tests/drum/test_custom_model.py
yakov-g: > what's the logic for what is False and what is True?

+1 : create a separate usecase, THen you won't need to put these dummy False.

And then, ode day, we could add R.

4dahalibut:
I didn't implement the feature for R.

4dahalibut:
We just don't have R custom training models yet.

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**2.49** USD.

PR #82, opened 2020-07-21
R fit template for DRUM

jlpolit:

## Summary

Create a new R fit component in DRUM, and corresponding R fit user template.

## Rationale

Allow users to upload custom training models using the R language

## Follow-up

We'll need to bump the drum version for the R env so the new fit method can be used (which I believe is also required to get a working functional test for this template)


model_templates/training/r_lang/custom.R
```diff
@@ -0,0 +1,191 @@
...
+  model_recipe <- recipe(target ~ ., data = train_df) %>%
+    # Drop constant columns
+    step_zv(all_predictors()) %>%
+    # Numeric preprocessing
+    step_medianimpute(all_numeric()) %>%
+    step_normalize(all_numeric(), -all_outcomes()) %>%
+    # Categorical preprocessing
+    step_other(all_nominal(), -all_outcomes()) %>%
+    step_dummy(all_nominal(), -all_outcomes())
+
```
4dahalibut, line 52: would you mind following the pattern of the other training models, and move this into another file, just to show users how imports work?

model_templates/training/r_lang/custom.R
```diff
@@ -0,0 +1,191 @@
...
+  # set up dataframe for modeling
+  train_df <- X
+  train_df$target <- unlist(y)
+
+  if (!is.null(class_order)){
+    # maybe add more processing to y if needed? R might not care
+    outfile <- 'r_classif.rds'
+  }else{
+    outfile <-'r_reg.rds'
+  }
```
4dahalibut, line 40: is there any reason to not always call the outfile artifact.rds?

model_templates/training/r_lang/custom.R
```diff
@@ -0,0 +1,191 @@
...
+  if(
+    substr(output_dir,
+          nchar(output_dir),
+          nchar(output_dir)) == '/'
+    ) {
+    seperator = ''
+  } else {
+    seperator = '/'
+  }
+
```
4dahalibut, line 66: Could you move this validation into python?

tests/drum/test_custom_model.py
```diff
@@ -681,6 +692,32 @@ def test_fit(self, framework, problem, language, docker, weights, use_output, tm
             cmd, "Failed in {} command line! {}".format(ArgumentsOptions.MAIN_COMMAND, cmd)
         )
 
+    @pytest.mark.parametrize("problem", [BINARY, REGRESSION])
+    @pytest.mark.parametrize("use_output", [True, False])
+    def test_r_fit(self, problem, use_output, tmp_path):
```
4dahalibut, line 697: Why not just add a parameter to test_fit?

jlpolit: As those params currently stand, I believe we're testing every combination of each param, which won't work if R params are introduced for language or framework.

It may be that there's an easy way I don't know about to test all the desired python combinations without testing impossible things like language=python and framework=rds?

model_templates/training/r_lang/custom.R
jlpolit: I don't see why not

tests/drum/test_custom_model.py
4dahalibut: I think you should get rid of the language param, and just make R a framework, and then have some dumb logic to choose language within the test.

jlpolit: that makes sense

model_templates/training/r_lang/custom.R
jlpolit: I think if I'm following the lead of the python templates, this logic would have to stay in R (since I believe the user is responsible for providing the code that will save their model rather than that happening on our end?)

I suppose if I am following the python templates' lead, I can just assume there's no trailing slash and let the users figure it out...

4dahalibut: The user is responsible for saving the model, but we can make sure theres always a slash inside drum.

jlpolit: fair!

model_templates/training/r_lang/custom.R
jlpolit: so far the only foolproof way I've found to do this involves moving the actual fitting step into the other file along with the "recipe" code- if that's cool, I'll push that change. otherwise I may have to confer with an R expert on best-practices

model_templates/training/r_lang/custom.R
```diff
@@ -0,0 +1,158 @@
...
+    #' @param ...: Added for forwards compatibility
+    #' @return Nothing
+
+  model <- create_pipeline(X, y)
+  # Save model
+  model_path <- file.path(
+    paste(
+      output_dir, 'artifact.rds', sep=''
+    )
+  )
```
jessezwu, line 38: The `paste` is unnecessary, `file.path(output_dir, 'artifact.rds')` is enough.

model_templates/training/r_lang/custom.R
```diff
@@ -0,0 +1,158 @@
+init <- function(code_dir) {
+  # custom init function to load required libraries
+  library(tidyverse)
+  library(caret)
+  library(recipes)
+  library(e1071)
+  library(gbm)
+  source(paste(code_dir, '/create_pipeline.R', sep=''))
```
jessezwu, line 8: `file.path(code_dir, 'create_pipeline.R')` is potentially cleaner here

custom_model_runner/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - error server is started in case of imports/pipeline failures
 - drum_autofit() helper for sklearn estimator
+- r_lang fit component, pipeline, and template
```
yakov-g, line 15: Please move this line under [1.1.4] section

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -145,6 +147,65 @@ def _check_artifacts_and_get_run_language(self):
...
+            )
+            all_files_message = "\n\nFiles(100 first) found in {}:\n{}\n".format(
+                code_dir_abspath, "\n".join(sorted(os.listdir(code_dir_abspath))[0:100])
+            )
+
+            error_mes += all_files_message
+            self.logger.error(error_mes)
+            raise DrumCommonException(error_mes)
+
+        # Get code dir's abs path and add it to the python path
```
yakov-g, line 167: you can remove this comment

model_templates/training/r_lang/custom.R
```diff
@@ -0,0 +1,158 @@
...
+##' code_dir - code folder passed in --code_dir argument
+##'
+##' @return NA, this method is intended for side effects only
+##' @export
+##'
+##' @examples
+#init <- function(...) {
+#}
+#
+##' This hook can be implemented to adjust logic in the scoring mode.
```
yakov-g, line 66: I suggest either :
- remove inference hooks in order not to confuse potential users
or
- comment that those hooks are INFERENCE...

but removing is better.

model_templates/training/r_lang/README.md
```diff
@@ -0,0 +1,14 @@
+## R Fit Template
```
yakov-g, line 1: As long as you already have this model template, could you add a functional test case?
https://github.com/datarobot/datarobot-user-models/blob/jlpolit/r_fit_draft/tests/functional/test_training_model_templates.py

custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
```diff
@@ -0,0 +1,99 @@
+# Needed libraries
```
yakov-g, line 1: In order not to duplicate code in score.R and fit.R
I suggest the following.
1. Create a folder `r_common_code`, (need to add `__init__.py` for the sake of setup). 
2. put all the code from `score.R` and `fit.R` into `common.R`
3. and import would look like
```
CUR_DIR = os.path.dirname(os.path.abspath(__file__))
R_COMMON_PATH = os.path.abspath(os.path.join(CUR_DIR, "..", "r_common_code", "common.R"))
```

model_templates/training/r_lang/README.md
jlpolit: I may be wrong here, but I thought we needed to publish the new version of DRUM and update that import in the R drop-in env for the functional test case to work?

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -145,6 +147,64 @@ def _check_artifacts_and_get_run_language(self):
...
+            raise DrumCommonException(error_mes)
+
+        code_dir_abspath = os.path.abspath(self.options.code_dir)
+
+        custom_language = None
+        run_language = None
+        is_py = False
+
+        # check which custom code files present in the code dir
+        custom_py_paths = list(Path(code_dir_abspath).rglob("{}.py".format("custom")))
```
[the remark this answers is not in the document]
jlpolit: for sure! good call

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -145,6 +147,64 @@ def _check_artifacts_and_get_run_language(self):
...
+            other_py = list(Path(code_dir_abspath).rglob("*.py"))
+
+            other_r = list(Path(code_dir_abspath).rglob("*.r")) + list(
+                Path(code_dir_abspath).rglob("*.R")
+            )
+
+            # if we find any py files and no R files set python, otherwise raise
+            if len(other_py) and not len(other_r):
+                is_py = True
+            else:
```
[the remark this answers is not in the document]
jlpolit: I don't believe we currently support non-"custom" files for R which is why I didn't put that in- I can try to figure out how to make an equivalent wrapper for R files (if that's a thing we can do, that is)

yakov-g:
jarvis run cmrunner_integration

yakov-g:

yakov-g:

model_templates/training/r_lang/README.md
yakov-g: Yes, you are right.
We can do the following: I take this branch,
1. Compile and publish drum 1.1.4rc1,
2. Update all the envs in current branch to use new drum and push into this PR. Functional tests then will use this new drum.
3. Once everything passes, we merge PR and install new envs on staging.

Just tell me when it is ready, so i can take it and publish drum.

yakov-g:

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**7.72** USD.

model_templates/training/r_lang/create_pipeline.R
```diff
@@ -0,0 +1,21 @@
+create_pipeline<-function(X, y) {
```
zachmayer, line 1: This is a lint comment, but traditionally you put spaces around operators like `<-`.

E.g. `create_pipeline <- function(X, y) {`

model_templates/training/r_lang/create_pipeline.R
```diff
@@ -0,0 +1,21 @@
...
+    step_normalize(all_numeric(), -all_outcomes()) %>%
+    # Categorical preprocessing
+    step_other(all_nominal(), -all_outcomes()) %>%
+    step_dummy(all_nominal(), -all_outcomes())
+
+  # Run the model using caret
+  model <- train(model_recipe, train_df, method = "gbm", trControl = trainControl(method = "cv", number = 3))
+
+  return(model)
+}
```
zachmayer, line 21: You should add a blank line at the end of the file.

@4dahalibut @meltzg There's a linter package for R: https://github.com/jimhester/lintr

Could we run it on the R files in this repo?

meltzg: we definitely should https://datarobot.atlassian.net/browse/RAPTOR-3229

PR #86, opened 2020-07-22
Update drum_integration_tests_base docker

yakov-g:

## Summary
Drum integration tests run in a docker container.
To compile scala code JDK should be installed into this container.

Dockerhub image has been already updated


datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**1.24** USD.

PR #87, opened 2020-07-22
Drum push + metadata file

4dahalibut:

## Summary

Lets people add a training or inference model from the CLI. These models are configured with a yaml file called model-metadata.yaml. 
## Rationale
This will make it easier for users to iterate on new models by not necessitating going through the UI to add a new model. It will also allow us to sneak in some validation if we centralize users to using this tool to push models into DataRobot.


model_templates/training/python3_sklearn/push.yaml
```diff
@@ -0,0 +1,7 @@
+model:
+    type: training
+    targetType: binary
+    environmentID: 5e8c889607389fe0f466c72d
+client:
+    endpoint: https://staging.datarobot.com/api/v2/
+    token: NWVhNzNlMDQ3OTM1OWUxODQ3ZWMxMWFiOk5EdGxPN1lEYWU0NUtFc3lZMjhmMzhHTFdPWGI5L2ovT1dRajA5Yk5XWGs9
```
meltzg, line 7: is this your token?

4dahalibut: of course not.

4dahalibut:
> why yaml? I think JSON might be more universally understood among non-engineers

Yaml is less ugly.

model_templates/training/python3_sklearn/push.yaml
```diff
@@ -0,0 +1,8 @@
+model:
+    name: drumpush-regression
+    type: training
+    targetType: regression
+    environmentID: 5e8c889607389fe0f466c72d
+client:
+    endpoint: https://staging.datarobot.com/api/v2/
+    token: NWVhNzNlMDQ3OTM1OWUxODQ3ZWMxMWFiOk5EdGxPN1lEYWU0NUtFc3lZMjhmMzhHTFdPWGI5L2ovT1dRajA5Yk5XWGs9
```
liororama: Hi, 
Wait a sec.. lets have a discussion about metdata in DRUM models before we add anything.. 
This addition requires a PBMP or at least a design doc.. 
Please lets not do this  in  a PR without  a design discussion.
@yakov-g

model_templates/training/python3_sklearn/push.yaml
```diff
@@ -0,0 +1,8 @@
+model:
+    name: drumpush-regression
+    type: training
+    targetType: regression
+    environmentID: 5e8c889607389fe0f466c72d
+client:
```
liororama, line 6: Client part should not be part of the metadata since this is not related to the model  itself.
Probably the endpoint and  token should be  provided via separate arguments to the drum or env

drum push -d /opt/models/my_model_1 --endpoint XXXX --token YYYY

model_templates/training/python3_sklearn/push.yaml
```diff
@@ -0,0 +1,8 @@
+model:
+    name: drumpush-regression
+    type: training
+    targetType: regression
+    environmentID: 5e8c889607389fe0f466c72d
```
liororama, line 5: Using an ID is not really user facing.. do we have other options? Like using a name?

model_templates/training/python3_sklearn/push.yaml
yakov-g: and env ID also should go just to the CLI param

yakov-g: drum push will create a model?
What will happen if I call `drum push` again? model will be existing model will be updated or another model will be  created?
Maybe we need to add model Id? And probably it should be just a string?

model_templates/training/python3_sklearn/push.yaml
yakov-g: Although we discussed some aspects with @4dahalibut , but looks like @liororama is right and it is just too big for a PR.

4dahalibut: I wanted this PR to be the design discussion

4dahalibut: I could schedule a meeting if you'd like

model_templates/training/python3_sklearn/push.yaml
4dahalibut: We can use a name but then people have to spell it right which is risky

4dahalibut: We could probably do some interesting regex to accomplish this though

klichukb: Often in such cases slugs are used as user-facing identifiers.
This repo already has unique slugs internally ("python3_keras", "python3_sklearn"), which could be easy to refer to.
And then if you look, push.yaml already lives under "python3_sklearn" env identifier, so its more or less aware.
Downside: I havent seen similar patterns in DR except maybe layout/overlay names in DRCA, + maintaining constraints within DR.

klichukb: (Remembered that we slugify organization names but use them afaik internally)

model_templates/training/python3_sklearn/push.yaml
klichukb: > > why yaml? I think JSON might be more universally understood among non-engineers
> 
> Yaml is less ugly.

agree that yaml is more user friendly and readable

model_templates/inference/python3_sklearn/model-metadata.yaml
```diff
@@ -0,0 +1,10 @@
+name: drumpush-regression
+type: inference
+targetType: regression
+modelID: 5f1f15a4d6111f01cb7f91fd
+environmentID: 5e8c889607389fe0f466c72d
+inferenceModel:
+  targetName: MEDV
+validation:
+  inputData: hi
```
yakov-g, line 9: Is this expected to be a path?

yakov-g: Could you add an example with class labels? Maybe even in regression, but null values.

tests/drum/run-drum-tests-in-container.sh
```diff
@@ -38,7 +38,7 @@ source $GIT_ROOT/tests/drum/integration-helpers.sh
 
 cd $GIT_ROOT || exit 1
 
-CMRUNNER_REQUIREMENT_PATH=$GIT_ROOT/custom_model_runner/requirements.txt
+CMRUNNER_REQUIREMENT_PATH=$GIT_ROOT/custom_model_runner/requirements_dev.txt
```
yakov-g, line 41: Why?

tests/drum/run-drum-tests-in-container.sh
```diff
@@ -47,9 +47,9 @@ build_docker_image_with_cmrun tests/fixtures/cmrun_docker_env \
                                $CMRUNNER_REQUIREMENT_PATH || exit 1
 
 echo
-echo "Installing the requirements for all tests:  $GIT_ROOT/requirements.txt"
+echo "Installing the requirements for all tests:  $GIT_ROOT/requirements_dev.txt"
```
yakov-g, line 50: Why?

tests/drum/run-drum-tests-in-container.sh
```diff
@@ -47,9 +47,9 @@ build_docker_image_with_cmrun tests/fixtures/cmrun_docker_env \
...
 cd $GIT_ROOT || exit 1
-pip install -r $GIT_ROOT/requirements.txt
+pip install -r $GIT_ROOT/requirements_dev.txt
```
yakov-g, line 52: Why?

tests/drum/locally-run-drum-tests.sh
```diff
@@ -41,7 +41,7 @@ function check_env_or_create() {
         exit 1
     fi
     . $ENV_ACTIVATE
-    pip install pytest -r ${GIT_ROOT}/requirements.txt
+    pip install pytest -r ${GIT_ROOT}/requirements_dev.txt
```
yakov-g, line 44: Why?

tests/drum/run-drum-tests-in-container.sh
4dahalibut: I added a dependency into requirements_test that I needed. I could change these to requirements_test as well...

model_templates/inference/python3_sklearn/model-metadata.yaml
4dahalibut: I'm going to follow this up with work to get the validation working.

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**8.58** USD.

PR #88, opened 2020-07-23
[RAPTOR-3220] unify all the pipelines into one pipeline template

yakov-g:

## Summary
I'll need to perform some refactoring to implement uwsgi.
This is a first step, which makes sense in any case: replace inference R/py/java pipeline templates with one template.


yakov-g:
jenkins test functional_drop_in

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**1.55** USD.

PR #90, opened 2020-07-25
[RAPTOR-3220] refactoring: separate predictors by language from mlpiper components

yakov-g:

## Summary
We have all prediction login in mlpiper components: python/java/R predictors.
1. I took out all the logic to separate classes: PythonPredictor/JavaPredictor/RPredictor whose main methods are configure(params) and predict(dataframe).
2. Now I can use these classes in a single batch `everything_predictor`  component and in a prediction server component.
3.  And now it will be easy to use it in a uwsgi component.

https://github.com/datarobot/datarobot-user-models/tree/master/custom_model_runner/datarobot_drum/resource/components/Python/python_predictor


custom_model_runner/datarobot_drum/resource/components/Python/everything_predictor/everything_predictor.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def configure(self, params):
+        super(EverythingPredictorComponent, self).configure(params)
+        self._run_language = RunLanguage(params.get("run_language"))
+
+        if self._run_language == RunLanguage.PYTHON:
+            self._predictor = PythonPredictor()
+        elif self._run_language == RunLanguage.JAVA:
+            self._predictor = JavaPredictor()
+        elif self._run_language == RunLanguage.R:
```
liororama, line 28: What if the language does not match java,python or R... ?

custom_model_runner/datarobot_drum/resource/components/Python/prediction_server/prediction_server.py
```diff
@@ -42,6 +51,17 @@ def configure(self, params):
...
         )
         self._memory_monitor = MemoryMonitor()
+        self._run_language = RunLanguage(params.get("run_language"))
+        if self._run_language == RunLanguage.PYTHON:
+            self._predictor = PythonPredictor()
+        elif self._run_language == RunLanguage.JAVA:
+            self._predictor = JavaPredictor()
+        elif self._run_language == RunLanguage.R:
+            from datarobot_drum.drum.language_predictors.r_predictor.r_predictor import RPredictor
+
```
liororama, line 61: No closing else?

custom_model_runner/datarobot_drum/resource/components/Python/everything_predictor/everything_predictor.py
yakov-g: I added else.
But this is internal code: we determine the language and pass further, so any error will be caught during development.

custom_model_runner/datarobot_drum/resource/components/Python/prediction_server/prediction_server.py
yakov-g: Added

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**5.0** USD.

PR #91, opened 2020-07-27
[RAPTOR-3195] Use a dedicated endpoint to check server health

rvorobii:

## Summary
Use a dedicated "/health/" route to check server state.
The reason to do this is that if "/predict/" is used - for healthy server we still can get 422 of dataset is not specified etc. and this error response is logged by flask.
As a result, even for healthy servers there's no easy way to get 200 response that'd mean "the server is healthy".

CM testing without this fix:
image

CM testing with this fix:
image

Note: errors are propagated via "/predict/" route as well - the current protocol is kept.


datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**0.82** USD.

PR #92, opened 2020-07-27
[RAPTOR-3220] refactoring: clean up after prev patch

yakov-g:

## Summary
Some clean up after the prev patch


datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**0.86** USD.

PR #93, opened 2020-07-28
[RAPTOR-3201] Add exception handler for prediction server

rvorobii:

## Summary
Currently exceptions raised during a prediction pipeline are not processed.
The result is very generic message:
image

This PR adds exception handler that would return error text along with 500:
image


rvorobii:
@meltzg could you have a look at this small PR please?

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**1.7** USD.

PR #95, opened 2020-07-28
[RAPTOR-3220] don't log exception if args were not parsed

yakov-g:

## Summary
Two fixes here:
1. Fix indentation for os._exit call in a SIGINT handler.
2. Showing error messages. Drum is used locally and in the env in the app.
Locally we want to show nice error message without stacktrace and in a server mode we want to see stacktrace.

Maybe we want to add a flag `--show-stacktrace` and use it in `start_server.sh` and `fit.sh`?

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/main.py
+++ b/custom_model_runner/datarobot_drum/drum/main.py
@@ -64,7 +64,7 @@ def main():
                     url = "http://{}/shutdown/".format(runtime.options.address)
                     print("Sending shutdown to server: {}".format(url))
                     requests.post(url, timeout=2)
-            os.system("tput init")
+
             os._exit(130)
 
         signal.signal(signal.SIGINT, signal_handler)
--- a/custom_model_runner/datarobot_drum/drum/main.py
+++ b/custom_model_runner/datarobot_drum/drum/main.py
@@ -64,8 +64,8 @@ def main():
                     url = "http://{}/shutdown/".format(runtime.options.address)
                     print("Sending shutdown to server: {}".format(url))
                     requests.post(url, timeout=2)
-                os.system("tput init")
-                os._exit(130)
+            os.system("tput init")
+            os._exit(130)
 
         signal.signal(signal.SIGINT, signal_handler)
 
--- a/custom_model_runner/datarobot_drum/drum/runtime.py
+++ b/custom_model_runner/datarobot_drum/drum/runtime.py
@@ -27,14 +27,14 @@ class DrumRuntime:
         if not exc_type:
             return True  # no exception, just return
 
+        if not self.options:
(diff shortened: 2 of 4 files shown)


custom_model_runner/datarobot_drum/drum/main.py
```diff
@@ -64,8 +64,8 @@ def signal_handler(sig, frame):
                     url = "http://{}/shutdown/".format(runtime.options.address)
                     print("Sending shutdown to server: {}".format(url))
                     requests.post(url, timeout=2)
-                os.system("tput init")
-                os._exit(130)
+            os.system("tput init")
```
rvorobii, line 67: :+1:

custom_model_runner/datarobot_drum/drum/runtime.py
```diff
@@ -27,14 +29,20 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
...
 
-        if not self.options:
-            # exception occurred before args were parsed
-            return False  # propagate exception further
+        if self.options.show_stacktrace:
+            logger.error(exc_msg)
+        else:
+            if exc_type == DrumCommonException:
+                print(exc_value)
+            exit(1)
```
rvorobii, line 45: why we exit here?
if user passes `--with-error-server`, if exception is not `DrumCommonException` - the server will not be started

IMO it's better to:
```
if run_mode == RunMode.SERVER or self.options.show_stacktrace:
    logger.error(exc_msg)
```
-- we check is drum is run in server mode or if user passes `--show-stacktrace`, and if so - log error

custom_model_runner/datarobot_drum/drum/runtime.py
```diff
@@ -37,14 +39,17 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
         exc_msg_lines = traceback.format_exception(exc_type, exc_value, exc_traceback)
         exc_msg = "".join(exc_msg_lines)
 
-        if self.options.show_stacktrace:
+        if (
+            self.options.show_stacktrace
+            or run_mode != RunMode.SERVER
```
rvorobii:
```suggestion
            or (run_mode == RunMode.SERVER
```

custom_model_runner/datarobot_drum/drum/runtime.py
```diff
@@ -37,14 +39,17 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
...
+            and self.options.with_error_server
```
rvorobii, line 45:
```suggestion
            and self.options.with_error_server)
```

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**7.58** USD.

PR #100, opened 2020-07-30
[RAPTOR-3249] don't use experimental methods that assume dependency management

meltzg:

## Summary
the blueprint creation method used in the test was changed to support unreleased behavior. I changed the test to use the more verbose, but more stable path

Also, since we're using the 2.22 beta, we have to use a base env id when creating model versions in drum push

### DIFF
--- a/tests/functional/test_training_model_templates.py
+++ b/tests/functional/test_training_model_templates.py
@@ -4,7 +4,7 @@ import os
 import pytest
 import datarobot as dr
 
-from datarobot._experimental import CustomTrainingBlueprint
+from datarobot._experimental import CustomTrainingBlueprint, CustomTrainingModel
 
 BASE_MODEL_TEMPLATES_DIR = "model_templates"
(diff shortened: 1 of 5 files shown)


tests/functional/test_training_model_templates.py
```diff
@@ -79,12 +79,18 @@ def test_training_model_templates(self, request, model_template, proj, env, targ
...
+            custom_model_id=model.id,
+            base_environment_id=env_id,
             folder_path=os.path.join(BASE_MODEL_TEMPLATES_DIR, model_template),
         )
+        blueprint = CustomTrainingBlueprint.create(
+            custom_model_id=model.id,
+            custom_model_version_id=model_version.id,
+            environment_id=env_id,
+            environment_version_id=env_version_id,
+        )
```
4dahalibut, line 93: Why are you changing the method which gets used?

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**2.14** USD.

PR #106, opened 2020-07-31
[RAPTOR-3258] small fixes for --docker param

yakov-g:

## Summary
When i provide dir as relative path with `..` , docker image get() method treats it as REST endpoint.

So first check if provided parameter is a directory, and work with dir; otherwise try to fetch image with provided name.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/drum.py
+++ b/custom_model_runner/datarobot_drum/drum/drum.py
@@ -437,7 +437,7 @@ class CMRunner(object):
         Returns: docker command line to run as a string
         """
 
-        options.docker = self.maybe_build_image(options.docker)
+        options.docker = self._maybe_build_image(options.docker)
         in_docker_model = "/opt/model"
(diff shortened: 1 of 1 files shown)


custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -540,10 +540,10 @@ def maybe_build_image(self, docker_image_or_directory):
         try:
             client.images.get(docker_image_or_directory)
             return docker_image_or_directory
-        except docker.errors.ImageNotFound:
+        except (docker.errors.ImageNotFound, docker.errors.APIError):
```
4dahalibut, line 543: Why are you adding this ?

4dahalibut: We should instead of doing this, check if its a directory first. IDK why I didn't do that.

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**3.16** USD.

PR #111, opened 2020-08-06
[RAPTOR-3246] changes for feedback from Jake Snyder

4dahalibut:

## Summary
* Enhanced some user facing communication
* Show user logs when verbose is not specified
* Memory profiling

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/common.py
+++ b/custom_model_runner/datarobot_drum/drum/common.py
@@ -131,6 +131,16 @@ class EnvVarNames:
 
 
 @contextmanager
+def reroute_stdout_to_stderr():
+    keep = sys.stdout
+    sys.stdout = sys.stderr
+    try:
+        yield
+    finally:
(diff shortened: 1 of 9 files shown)


custom_model_runner/datarobot_drum/drum/common.py
```diff
@@ -130,6 +130,16 @@ class EnvVarNames:
...
 
 
+@contextmanager
+def reroute_stdout_to_stderr():
+    keep = sys.stdout
+    sys.stdout = sys.stderr
+    try:
+        yield
+    finally:
+        sys.stdout = keep
```
meltzg, line 140: does this redirect stderr to stdout, or just replace stdout with stderr?

4dahalibut: redirects stdout to stderr

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**1.6** USD.

PR #113, opened 2020-08-10
[RAPTOR-3292] Pytorch template

jlpolit:

## Summary

Vanilla regression/classification template...I can add some preprocessing and/or a vizAI version as with the keras templates.


datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**0.96** USD.

PR #115, opened 2020-08-15
[RAPTOR-3319] allow to enforce run language

yakov-g:

## Summary
Allow to enforce run language with argument: --language [java, r, python]
This argument will be added into start_server.sh in the drop in environments

Added header "Running environment language: [R|Java|Python]" into error messages. Because of current code structure, I didn't find a nice way to add it into one common place.

## Rationale
This argument will be used in the drop in envs, to give env some control and ensure that drum will running proper framework, e.g. running Java in java env.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/args_parser.py
+++ b/custom_model_runner/datarobot_drum/drum/args_parser.py
@@ -314,14 +314,20 @@ class CMRunnerArgsRegistry(object):
 
     @staticmethod
     def _reg_arg_language(*parsers):
-        langs = [e.value for e in RunLanguage]
-        langs.remove(RunLanguage.JAVA.value)
         for parser in parsers:
+            langs = [e.value for e in RunLanguage]
+            prog_name_lst = CMRunnerArgsRegistry._tokenize_parser_prog(parser)
+            if prog_name_lst[1] == ArgumentsOptions.NEW:
+                langs.remove(RunLanguage.JAVA.value)
+                required_val = True
+            else:
+                required_val = False
+
             parser.add_argument(
                 ArgumentsOptions.LANGUAGE,
                 choices=langs,
                 default=None,
-                required=True,
+                required=required_val,
                 help="Language to use for the new model/env template to create",
             )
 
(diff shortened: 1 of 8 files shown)


yakov-g:

custom_model_runner/datarobot_drum/drum/args_parser.py
```diff
@@ -314,14 +314,20 @@ def _reg_arg_url(*parsers):
 
     @staticmethod
     def _reg_arg_language(*parsers):
-        langs = [e.value for e in RunLanguage]
-        langs.remove(RunLanguage.JAVA.value)
         for parser in parsers:
+            langs = [e.value for e in RunLanguage]
+            prog_name_lst = CMRunnerArgsRegistry._tokenize_parser_prog(parser)
```
klichukb, line 319: can you explain briefly what does this logic below means?

yakov-g:
```
@staticmethod
    def _tokenize_parser_prog(parser):
        # example:
        # - for score_parser prog is "drum score"
        # - for new_model_parser prog is "drum new model"
        return parser.prog.split(" ")
```
So i split prog by tokens to be able to customize arg's behavior for different subcommands.

klichukb: I see, but what about "if its a new model, remove java from langs"?

klichukb: so from DMs FYI, this means that this logic doesnt support creating a template for java for now.
As well we figured it out that we might start doing so for mojo/pojo.

tests/drum/test_custom_model.py
```diff
@@ -476,6 +480,78 @@ def test_detect_language(self, framework, problem, language, tmp_path):
...
+            ),  # java and sklearn artifacts, no custom.py
+            (
+                CODEGEN_AND_SKLEARN,
+                REGRESSION,
+                NO_CUSTOM,
+                "r",
+            ),  # java and sklearn artifacts, no custom.py
+        ],
+    )
+    def test_set_language(self, framework, problem, language, set_language, tmp_path):
```
klichukb, line 520: Very nice test

custom_model_runner/datarobot_drum/drum/args_parser.py
yakov-g: https://datarobot.atlassian.net/browse/RAPTOR-3330

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**4.45** USD.

PR #116, opened 2020-08-17
Skip feature impact test before FI API is available in the API client

klichukb:
This test will be unskipped when we implement API for calculating feature impact


datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**0.95** USD.

PR #117, opened 2020-08-18
[RAPTOR-3332] Fix & unskip feature impact test

klichukb:
This should trigger calculation of Feature Impact and so test might become valid again.


datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**1.96** USD.

PR #118, opened 2020-08-18
[RAPTOR-3342] relocate import sklearn

yakov-g:

## Summary
Relocate `import sklearn` into a method using sklearn.

## Rationale
Global `import sklearn`  in this file interferes when running inference models which don't require sklearn.


datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**1.03** USD.

PR #119, opened 2020-08-20
update to handle no-target models

jlpolit:

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/args_parser.py
+++ b/custom_model_runner/datarobot_drum/drum/args_parser.py
@@ -112,8 +112,8 @@ class CMRunnerArgsRegistry(object):
                 ArgumentsOptions.TARGET,
                 type=str,
                 required=False,
-                help="Which column to use as the target. Argument is mutually exclusive with {}".format(
-                    ArgumentsOptions.TARGET_FILENAME
+                help="Which column to use as the target. Argument is mutually exclusive with {} and {}".format(
+                    ArgumentsOptions.TARGET_FILENAME, ArgumentsOptions.UNSUPERVISED_ANOMALY
                 ),
             )
 
@@ -121,9 +121,20 @@ class CMRunnerArgsRegistry(object):
                 ArgumentsOptions.TARGET_FILENAME,
                 type=CMRunnerArgsRegistry._is_valid_file,
                 required=False,
-                help="A file containing the target values. Argument is mutually exclusive with {}".format(
-                    ArgumentsOptions.TARGET
+                help="A file containing the target values. Argument is mutually exclusive with {} and {}".format(
+                    ArgumentsOptions.TARGET, ArgumentsOptions.UNSUPERVISED_ANOMALY
                 ),
+            )
+
+            group.add_argument(
+                ArgumentsOptions.UNSUPERVISED_ANOMALY,
+                action='store_true',
+                required=False,
+                default=False,
+                help="If present, indicates that this is an unsupervised anomaly detection model."
+                     " Argument is mutually exclusive with {} and {}".format(
+                        ArgumentsOptions.TARGET, ArgumentsOptions.UNSUPERVISED_ANOMALY
+                     ),
             )
 
     @staticmethod
@@ -136,7 +147,7 @@ class CMRunnerArgsRegistry(object):
                 required=False,
                 default=None,
                 help="A column name of row weights in your training dataframe. "
-                "Argument is mutually exclusive with {}".format(ArgumentsOptions.WEIGHTS_CSV),
+                     "Argument is mutually exclusive with {}".format(ArgumentsOptions.WEIGHTS_CSV),
             )
             group.add_argument(
                 ArgumentsOptions.WEIGHTS_CSV,
@@ -144,7 +155,7 @@ class CMRunnerArgsRegistry(object):
                 required=False,
                 default=None,
                 help="A one column csv file to be parsed as row weights. "
-                "Argument is mutually exclusive with {}".format(ArgumentsOptions.WEIGHTS),
+                     "Argument is mutually exclusive with {}".format(ArgumentsOptions.WEIGHTS),
             )
 
     @staticmethod
@@ -156,7 +167,7 @@ class CMRunnerArgsRegistry(object):
                 default=False,
                 action="store_true",
                 help="By default we will attempt to predict using your model, but we give you the"
-                "option to turn this off",
+                     "option to turn this off",
             )
 
     @staticmethod
@@ -241,8 +252,8 @@ class CMRunnerArgsRegistry(object):
                 default=None,
                 required=False,
                 help="Docker image to use to run {} in the {} mode, "
-                "or a directory, containing a Dockerfile, "
-                "which can be built into a docker image. ".format(
+                     "or a directory, containing a Dockerfile, "
+                     "which can be built into a docker image. ".format(
                     ArgumentsOptions.MAIN_COMMAND, prog_name_lst[1]
                 ),
             )
@@ -338,7 +349,7 @@ class CMRunnerArgsRegistry(object):
                 ArgumentsOptions.NUM_ROWS,
                 default="ALL",
                 help="Number of rows to use for testing the fit functionality. "
-                "Set to ALL to use all rows. Default is 100",
+                     "Set to ALL to use all rows. Default is 100",
             )
 
     @staticmethod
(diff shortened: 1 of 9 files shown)


custom_model_runner/datarobot_drum/drum/args_parser.py
```diff
@@ -112,20 +112,31 @@ def _reg_arg_target_feature_and_filename(*parsers):
                 ArgumentsOptions.TARGET,
                 type=str,
                 required=False,
-                help="Which column to use as the target. Argument is mutually exclusive with {}".format(
-                    ArgumentsOptions.TARGET_FILENAME
+                help="Which column to use as the target. Argument is mutually exclusive with {} and {}".format(
+                    ArgumentsOptions.TARGET_FILENAME, ArgumentsOptions.UNSUPERVISED
```
4dahalibut, line 116: you can make arguments mutually exclusive using argparse https://stackoverflow.com/questions/17909294/python-argparse-mutual-exclusive-group

custom_model_runner/datarobot_drum/drum/common.py
```diff
@@ -92,6 +92,7 @@ class ArgumentsOptions(object):
     LOGGING_LEVEL = "--logging-level"
     LANGUAGE = "--language"
     NUM_ROWS = "--num-rows"
+    UNSUPERVISED = "--unsupervised"
```
4dahalibut, line 95: unsupervised or anomaly? you say one above and the other here.

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -366,9 +369,14 @@ def _drum_autofit_internal(self, X, y, output_dir):
     def fit(self, X, y, output_dir, class_order=None, row_weights=None):
         with reroute_stdout_to_stderr():
             if self._custom_hooks.get(CustomHooks.FIT):
-                self._custom_hooks[CustomHooks.FIT](
-                    X, y, output_dir, class_order=class_order, row_weights=row_weights
-                )
+                if y is not None:
```
4dahalibut, line 372: I think it might be better to pass y in as None explicitly.

4dahalibut: that way the args are always the same

custom_model_runner/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### [1.1.5rc1] - in progress
 ##### Added
 - optional **--language [python|r|java]** argument to enforce execution framework
+- optional **--anomaly-detection** argument and handling for unsupervised Fit in python
```
4dahalibut, line 10: this should change to --unsupervised

model_templates/training/python3_anomaly_detection/custom.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    This hook must be implemented with your fitting code, for running drum in the fit mode.
+
+    This hook MUST ALWAYS be implemented for custom training models.
+    For inference models, this hook can stick around unimplemented, and won’t be triggered.
+
+    Parameters
+    ----------
+    X: pd.DataFrame - training data to perform fit on
+    y: pd.Series - target data to perform fit on
```
4dahalibut, line 20: you'll want to change this

custom_model_runner/datarobot_drum/drum/args_parser.py
jlpolit: I think these three are mutually exclusive as things stand? all three are added to "group" which is defined: `group = parser.add_mutually_exclusive_group(required=True)` on line 110

public_dropin_environments/python3_sklearn/fit.sh
```diff
@@ -4,12 +4,19 @@ cd "${CODEPATH}" || exit 1
 export PYTHONPATH="${CODEPATH}":"${PYTHONPATH}"
```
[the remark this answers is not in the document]
jlpolit: (that is, of course, assuming we want this functionality supported for all dropin environments?)

custom_model_runner/datarobot_drum/drum/args_parser.py
4dahalibut: got it, I didn't notice that.

yakov-g:
LGTM
The only thing -  tests become super oversized and complicated, do we want to refactor them somehow? Split into multiple files?

jlpolit:
> LGTM
> The only thing - tests become super oversized and complicated, do we want to refactor them somehow? Split into multiple files?

good call - it's definitely only going to get more convoluted as more target types are added

yakov-g:
> good call - it's definitely only going to get more convoluted as more target types are added
But how do we split? 
1. Training/inference?
2. By  framework type per file: sklearn with training/inference/anomaly... Each framework defines it's datasets, and then training test case could go first, and inference test case works on the trained model..
3. Other suggestions?

yakov-g:
@jlpolit maybe you should update docs on anomaly?
btw, why it has dual name anomaly/unsupervised
@4dahalibut you probably want to review this

Let's merge it by mid day tomorrow

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -346,7 +346,8 @@ def _prepare_prediction_server_or_batch_pipeline(self, run_language):
         return functional_pipeline_str
 
     def _prepare_fit_pipeline(self, run_language):
-        if not self.options.negative_class_label:
+
+        if not self.options.negative_class_label and not self.options.unsupervised:
```
4dahalibut, line 350: does using class labels fail for unsupervised?

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -357,7 +357,10 @@ def _drum_autofit_internal(self, X, y, output_dir):
                         break
 
         if marked_object is not None:
-            marked_object.fit(X, y)
+            if y is not None:
+                marked_object.fit(X, y)
+            else:
+                marked_object.fit(X)
```
4dahalibut, line 363: do you think we should pass y=none here explicitly?

custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
```diff
@@ -46,3 +46,17 @@ outer_fit <- function(X, y, output_dir, class_order=NULL, row_weights=NULL) {
             stop(sprintf("No Fit method provided."))
         }
 }
+
+
+outer_fit_unsupervised <- function(X, output_dir, row_weights=NULL) {
+    if (!isFALSE(fit_hook)) {
```
4dahalibut, line 52: theres no isTRUE?

model_templates/training/python3_anomaly_detection/README.md
```diff
@@ -0,0 +1,11 @@
+## Python Unsupevised Anomaly Detection Training Model Template
```
4dahalibut, line 1: typo in unsupervised...

custom_model_runner/datarobot_drum/drum/drum.py
jlpolit: possibly_intuit_order will throw an error if there's no target- that's why this check is here

custom_model_runner/datarobot_drum/drum/model_adapter.py
jlpolit: yeah I guess there's no reason not to

model_templates/training/python3_anomaly_detection/README.md
jlpolit: nope, that's just how we spell it now

custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
jlpolit: did some testing - looks like `!isFALSE ` and `isTRUE` don't behave the same, because R is the worst

custom_model_runner/datarobot_drum/drum/drum.py
4dahalibut: what if you made it not throw an error though.

jlpolit: I can probably do that

yakov-g:
jarvis test black

jlpolit:
jarvis test black

jlpolit:
> Some minor things

lmk how this looks. also whether the fit.sh updates should be made to the other envs (in this or possibly a future update)? thanks

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**2.3** USD.

PR #122, opened 2020-08-28
[RAPTOR-3395] drop any fully missing columns in preprocessing for keras and xgboost

jlpolit:

## Summary

Our numeric preprocessing for sklearn and xgboost was causing an error if the users passed in data with a totally-missing numeric column. Arguably the user shouldn't be doing this, but it is also relatively straightforward to handle...maybe I should add some logging to indicate that we encountered a column with nothing but NaNs and had to drop it?

### DIFF
--- a/model_templates/training/python3_keras_joblib/README.md
+++ b/model_templates/training/python3_keras_joblib/README.md
@@ -5,7 +5,9 @@ The supplied h5 file is a keras + tensorflow model trained on [boston_housing.cs
 with a MEDV as the target (regression), though any binary or regression model trained using the libraries
 outlined in [Python 3 Keras Drop-In Environment](../../../public_dropin_environments/python3_keras/) will work.
 
-For this sample model, custom.py contains additional data pre-processing that the model itself lacks.
+For this sample model, custom.py contains additional data pre-processing that the model itself lacks.  In this case,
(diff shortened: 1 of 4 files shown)


yakov-g:
Just wondering:
Are these templates intended to work with any dataset?

jlpolit:
> Just wondering:
> Are these templates intended to work with any dataset?

I wasn't sure about that either...my assumption would be that they should serve as a jumping off point, and should definitely work with the datasets we provide but we can't guarantee they'll work for any arbitrary dataset. 

It could be that this bug is more of an indicator that we aren't making that sufficiently clear to users? But others should probably weigh in

yakov-g:
In this case, let's update READMEs for these templates to tell users what is going on.
Use sklearn training readme as an example:  https://github.com/datarobot/datarobot-user-models/blob/master/model_templates/training/python3_sklearn/README.md

jlpolit:
> In this case, let's update READMEs for these templates to tell users what is going on.
> Use sklearn training readme as an example: https://github.com/datarobot/datarobot-user-models/blob/master/model_templates/training/python3_sklearn/README.md

good call!

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**1.77** USD.

PR #123, opened 2020-08-28
Switch drop-in envs to work with uwsgi server with one worker

yakov-g:

## Summary
**### This will be merged after drum release and prod drop in envs update.**

Enable uwsgi based prediction server with one worker in drop in envs.
For now, these envs will be deployed only on staging.

--with-error-server param has been removed. Error server has been used to serve error messages, if model loading fails.
With uwsgi server, loading fails on worker, so main drum process won't fail. Instead worker will be serving error message.

### DIFF
--- a/dev_env/java_nginx_env/Dockerfile
+++ b/dev_env/java_nginx_env/Dockerfile
@@ -1,9 +1,7 @@
 # This is the default base image for use with user models and workflows.
 FROM datarobot/java-dropin-env-base
 
-RUN apt-get update && apt-get install -y vim nginx
-
-RUN chmod 707 /var/lib/nginx
+RUN apt-get update && apt-get install -y vim
 
 # Install the list of core requirements, e.g. sklearn, numpy, pandas, flask.
 # **Don't modify this file!**
--- a/public_dropin_environments/r_lang/start_server.sh
+++ b/public_dropin_environments/r_lang/start_server.sh
@@ -2,7 +2,7 @@
 cd /opt/code/ || exit 1
 export PYTHONPATH=/opt/code
 
-CMD="drum server -cd . --address 0.0.0.0:8080 --with-error-server"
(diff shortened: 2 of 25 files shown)


dev_env/java_nginx_env/Dockerfile
```diff
@@ -1,9 +1,7 @@
 # This is the default base image for use with user models and workflows.
 FROM datarobot/java-dropin-env-base
 
-RUN apt-get update && apt-get install -y vim nginx
-
-RUN chmod 707 /var/lib/nginx
+RUN apt-get update && apt-get install -y vim
```
meltzg, line 4: why do we install vim?

yakov-g: These are development/test/debug envs, so just to have it there.

public_dropin_environments/r_lang/start_server.sh
```diff
@@ -2,7 +2,7 @@
 cd /opt/code/ || exit 1
 export PYTHONPATH=/opt/code
 
-CMD="drum server -cd . --address 0.0.0.0:8080 --with-error-server"
+CMD="drum server -cd . --address 0.0.0.0:8080 --production --max-workers 1 --show-stacktrace"
```
meltzg, line 5: do we no longer need the error server?

yakov-g:
> environment version IDs need to be updated

Envs ID's have been updated before and installed on prod.
So envs are uninstallable on staging. Do you want to update them once again to differ from prod.?

public_dropin_environments/r_lang/start_server.sh
yakov-g: --with-error-server param has been removed. Error server has been used to serve error messages, if model loading fails.
With uwsgi server, loading fails on worker, so main drum process won't fail. Instead worker will be serving error message.

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**6.23** USD.

PR #126, opened 2020-09-01
Fix regression dimension mismatch in pytorch template

jlpolit:

## Summary

y_hat was being returned as a (batch_size, 1) shaped vector, while the y var is (batch_size) shaped, which caused pytorch to complain. This reshapes that target to the same dimension as the model's output.


4dahalibut (approved):

jmbannon (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**1.11** USD.

PR #128, opened 2020-09-01
[RAPTOR-3337]fix situation where the output dir is inside the code dir

scottp-ml:

## Summary
Bug fix that prevents errors if the user specifies an output directory within the code director.


yakov-g:
Looks ok.
Please add tests:
https://github.com/datarobot/datarobot-user-models/blob/4f740ac40528c4972056acaf896548715d10896f/tests/drum/test_custom_model.py#L847 copy this method, and change fixture to run only one use case - output into path inside code_dir.

4dahalibut:
Would you mind checking this behavior holds in a test?

4dahalibut:
just a unit test should be fine

yakov-g (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**2.95** USD.

PR #140, opened 2020-09-21
[RAPTOR-3503] pytorch template and handling updates

jlpolit:

## Summary

Pytorch template didn't have preprocessing to handle non-numeric data, so it was failing on 10k diabetes (and others). This PR adds processing to identify and keep only numeric data (further processing can be included at a future time to actually use text/categorical data for features) and functional tests for the pytorch template on 10k diabetes. 

It also adds 10k diabetes to the repo as a test dataset; expect some follow-up PRs to test a wider array of the current templates against this data, and possibly a more general set of "diversity" checks against a large number of small but varied datasets (i.e. ~100 randomly sampled rows from a yaml of commonly MBTested datasets).


custom_model_runner/datarobot_drum/drum/artifact_predictors/torch_predictor.py
```diff
@@ -66,7 +66,11 @@ def predict(self, data, model, **kwargs):
         # done in the base class
         super(PyTorchPredictor, self).predict(data, model, **kwargs)
 
-        data = Variable(torch.from_numpy(data.values).type(torch.FloatTensor))
+        data = Variable(
```
jlpolit, line 69: This isn't a deal breaker to include, but is probably nice to have if we want to allow users flexibility in what format they pass their data in after preprocessing. Currently we're assuming it will reach pytorch as a dataframe, but if tools like sklearn are used as part of the preprocessing flow this is not guaranteed to be the case.

model_templates/training/python3_pytorch/custom.py
```diff
@@ -47,15 +48,42 @@ def fit(
...
+    model: object, the deserialized model
+
+    Returns
+    -------
+    pd.DataFrame
+    """
+    # Execute any steps you need to do before scoring
+    # Remove target columns if  they're in the dataset
+    preprocessor, cols = build_preprocessor(data)
+    preprocessor.fit(data)
```
jmbannon, line 87: This doesn't seem right, the preprocessor shouldn't be rebuilt at transform time

jlpolit: oh good call - so we may need to wrap both the preprocessor and modeler in a separate (sklearn-style?) estimator object that we can pickle/unpickle as one unit...or, I guess, save two artifacts in fit and write another custom hook to unpack them?

jmbannon: An sklearn pipeline that contains all preprocessing and the estimator would be ideal

jlpolit: yeah- I'm not sure that there's a nice plug-and-play way to do that with pytorch

jmbannon: This has a good example: https://www.kaggle.com/graymant/pytorch-regression-with-sklearn-pipelines
Have to extend the sklearn BaseEstimator class

jlpolit: ....I'm going to punt on this task (making a new ticket to figure out a good framework for this) and just put in a subset transformer to keep only numerical features for now. WIP of the sklearn estimator is [here](https://github.com/datarobot/datarobot-user-models/tree/jlpolit/pytorch-sklearn-estimator-wip) but it still needs some work

meltzg (approved):

4dahalibut (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**3.55** USD.

PR #141, opened 2020-09-21
[RAPTOR-3555] init monitoring arguments in modes that implicitly use them

yakov-g:

## Summary
Monitoring arguments assignment should be available only for score and server modes.
But initialize those arguments to False/None, if they are needed for fit or other modes.


liororama (approved):

meltzg (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**2.25** USD.

PR #146, opened 2020-09-28
fix encoding r issue

jlpolit:

## Summary

Fix an rpy2 issue where nans in non-float fields confuse python and it tries to encode a float column. Still more annoying R bugs to go....

### DIFF
--- a/custom_model_runner/datarobot_drum/resource/components/Python/r_fit/r_fit.py
+++ b/custom_model_runner/datarobot_drum/resource/components/Python/r_fit/r_fit.py
@@ -81,9 +81,9 @@ class RFit(ConnectableComponent):
 
         # replace any errant NaN's in object cols with None
         X_cols = X.columns
-        X_obj = X.select_dtypes(include=['object'])
-        X_obj = X_obj.where(pd.notnull(X.select_dtypes(include=['object'])), None)
-        X = pd.concat([X.select_dtypes(exclude=['object']), X_obj], axis=1)
+        X_obj = X.select_dtypes(include=["object"])
+        X_obj = X_obj.where(pd.notnull(X.select_dtypes(include=["object"])), None)
+        X = pd.concat([X.select_dtypes(exclude=["object"]), X_obj], axis=1)
         X = X[X_cols]
 
         optional_args = {}
--- a/custom_model_runner/datarobot_drum/resource/components/Python/r_fit/r_fit.py
+++ b/custom_model_runner/datarobot_drum/resource/components/Python/r_fit/r_fit.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import pandas as pd
 
 from mlpiper.components.connectable_component import ConnectableComponent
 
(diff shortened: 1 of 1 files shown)


custom_model_runner/datarobot_drum/resource/components/Python/r_fit/r_fit.py
```diff
@@ -78,6 +79,13 @@ def _materialize(self, parent_data_objs, user_data):
 
         X, y, class_order, row_weights = shared_fit_preprocessing(self)
 
+        # replace any errant NaN's in object cols with None
```
yakov-g, line 82: Why does this happen?
Should we open datasets in R itself?
Or maybe data should be processed in the R fit code?

yakov-g (changes requested):

custom_model_runner/datarobot_drum/resource/components/Python/r_fit/r_fit.py
jlpolit: that's a great question.....it could be that this is an easier fix by messing with the pandas read_csv settings instead of fiddling around in r_fit- I can look into that

yakov-g: Once again, maybe `shared_fit_preprocessing` shouldn't be shared and opening data file should happen in `outer_fit` ?

Like I did it for inference. I had to move file opening into each language itself: Py/R/Java in order to add custom users's hook to open file. https://github.com/datarobot/datarobot-user-models/blob/c2df04ce6e95b5df84323b6fbeebd19964d60f82/custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/score.R#L174
May not be relevant for training, but maybe will help to solve other R-related issues you mentioned.

Also because as R - DS, I'd probably expect to see not processed data... Like NA, NaN values processed by R and not by Pandas.

Tell me what you think.

jlpolit: I think @4dahalibut and/or @meltzg  may want to weigh in on that question - I could imagine some advantages of moving the data loading into R itself, but it comes at the cost of messier debugging

meltzg: I might be on board the `move data loading into native language environment` boat.  It's probably worth floating the idea with some R people, but to me, it makes sense to parse the csv using the same tools as the model, especially since R and pandas treat dataframes differently

4dahalibut: We'd probably have to write out data then if the user supplies just one dataset file... I think thats probably going to avoid bugs though given rpy2 has already proven to be buggy.

jlpolit: alright, sounds like we move it to R. I'll close this PR and open a ticket to make that change- @4dahalibut are you talking about a case where we get a single file instead of a separate X and y file or something else?

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**1.14** USD.

PR #147, opened 2020-09-29
[RAPTOR-3643] import keras from TF in drum keras predictor

yakov-g:

## Summary
Add a check for keras model is instance of keras included into tensorflow.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/artifact_predictors/keras_predictor.py
+++ b/custom_model_runner/datarobot_drum/drum/artifact_predictors/keras_predictor.py
@@ -47,12 +47,13 @@ class KerasPredictor(ArtifactPredictor):
         try:
             import keras
             from sklearn.pipeline import Pipeline
+            from tensorflow import keras as keras_tf
 
             if isinstance(model, Pipeline):
                 # check the final estimator in the pipeline is Keras
-                if isinstance(model[-1], keras.Model):
+                if isinstance(model[-1], (keras.Model, keras_tf.Model)):
                     return True
-            elif isinstance(model, keras.Model):
+            elif isinstance(model, (keras.Model, keras_tf.Model)):
                 return True
         except Exception as e:
             self._logger.debug("Exception: {}".format(e))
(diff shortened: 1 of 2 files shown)


4dahalibut (approved):

custom_model_runner/datarobot_drum/drum/artifact_predictors/keras_predictor.py
```diff
@@ -47,12 +47,13 @@ def can_use_model(self, model):
         try:
             import keras
             from sklearn.pipeline import Pipeline
+            from tensorflow import keras as keras_tf
 
             if isinstance(model, Pipeline):
                 # check the final estimator in the pipeline is Keras
-                if isinstance(model[-1], keras.Model):
+                if isinstance(model[-1], (keras.Model, keras_tf.Model)):
```
4dahalibut, line 54: Is there any parent class they both depend on?

yakov-g: No, looks like there is no.
As I understand, Keras is API standard which can use different backends.

Starting tensorflow 2 - tensorflow implements keras.
https://blog.tensorflow.org/2018/12/standardizing-on-keras-guidance.html

So technically, tensorflow >= 2 doesn't require keras installation.

meltzg: That depends on the version of keras.  In the latest version of standalone keras, they're the same class. In the version included with the public env, the only common parent is object

meltzg (approved):

custom_model_runner/datarobot_drum/drum/artifact_predictors/keras_predictor.py
yakov-g: I see, other backends has been discontinued: https://github.com/keras-team/keras/blob/master/README.md

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**1.23** USD.

PR #148, opened 2020-09-30
fix torch template & update variety data key

jlpolit:

## Summary

Doing a dumb zero-imputation step so torch template doesn't fail on datasets w/ missing data. Also fixed the targets for a dataset in the variety samples, which should be floats not int.


yakov-g (approved):

4dahalibut (approved):

scottp-ml (approved):
LGTM

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**1.25** USD.

PR #150, opened 2020-09-30
[RAPTOR-3644] Move R fit data handling into R language

jlpolit:

## Summary

Moving all of the data handling from python into R for the r_fit functionality in DRUM.

### DIFF
--- a/custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
+++ b/custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
@@ -5,6 +5,7 @@ library(devtools)
 init_hook <- FALSE
 fit_hook <- FALSE
 
+set.seed(1)
 
 init <- function(code_dir) {
   if (file.exists(file.path(code_dir, 'custom.R'))){
@@ -26,20 +27,18 @@ init <- function(code_dir) {
 
 load_data <- function(input_filename){
     tmp = readChar(input_filename, file.info(input_filename)$size)
-    data <- read.csv(text=gsub("\r","", tmp, fixed=TRUE))
-    data
+    read.csv(text=gsub("\r","", tmp, fixed=TRUE))
 }
 
 
 process_data <- function(input_filename, target_filename, target_name, num_rows){
-    set.seed(1)
     # read X
     df <- load_data(input_filename)
     # set num_rows
     if (num_rows == 'ALL'){
         num_rows <- nrow(df)
     } else {
-        num_rows <- as.integer(a)
+        num_rows <- as.integer(num_rows)
     }
     # handle target
     if (!is.null(target_filename) || !is.null(target_name)){
@@ -71,7 +70,6 @@ process_data <- function(input_filename, target_filename, target_name, num_rows)
 
 process_weights <- function(X, weights_filename, weights, num_rows){
   if (!is.null(weights_filename)){
-    set.seed(1)
     row_weights <- load_data(weights_filename)
     row_weights <- row_weights[sample(nrow(row_weights), size=num_rows, replace=TRUE ), ]
   } else if(!is.null(weights)){
--- a/custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
+++ b/custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
@@ -24,6 +24,68 @@ init <- function(code_dir) {
 }
 
 
+load_data <- function(input_filename){
+    tmp = readChar(input_filename, file.info(input_filename)$size)
+    data <- read.csv(text=gsub("\r","", tmp, fixed=TRUE))
+    data
+}
+
+
+process_data <- function(input_filename, target_filename, target_name, num_rows){
+    set.seed(1)
+    # read X
+    df <- load_data(input_filename)
+    # set num_rows
+    if (num_rows == 'ALL'){
+        num_rows <- nrow(df)
+    } else {
+        num_rows <- as.integer(a)
+    }
+    # handle target
+    if (!is.null(target_filename) || !is.null(target_name)){
+        if (!is.null(target_filename)){
+            y_unsampled = load_data(target_filename)
+            assert( length(colnames(y_unsampled)) == 1)
+            assert( nrow(df) == nrow(y_unsampled))
+            df <- cbind(df, y)
+            target_name <- colnames(y)
+        }
+    df <-  df[!(is.na(df[target_name])) , ]
+    X <-  df[,!(names(df) %in% c(target_name))]
+    X <- X[sample(nrow(X), size=num_rows, replace=TRUE ), ]
+
+    y <- X[,target_name, drop=FALSE]
+    y <- y[sample(nrow(y), size=num_rows, replace=TRUE),]
+
+    } else {
+
+        y <- NULL
+        X <- df[sample(nrow(df), size=num_rows, replace=TRUE ), ]
+
+    }
+
+    return (list('X' = X, 'y' = y, 'num_rows' = num_rows))
+
+}
+
+
+process_weights <- function(X, weights_filename, weights, num_rows){
+  if (!is.null(weights_filename)){
+    row_weights <- load_data(weights_filename)
+    row_weights <- row_weights[sample(nrow(row_weights), size=num_rows, replace=TRUE ), ]
+  } else if(!is.null(weights)){
+    if (! weights %in% colnames(X)){
+      stop( paste("The column name",
+                  weights,
+                  "is not one of the columns in your training data", sep=' '))
+    }
+    row_weights <- X[, weights, drop=FALSE]
+  } else {
+    row_weights = NULL
+  }
+  row_weights
+}
+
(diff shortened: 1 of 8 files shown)


custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
```diff
@@ -24,6 +24,69 @@ init <- function(code_dir) {
...
+        if (!is.null(target_filename)){
+            y_unsampled = load_data(target_filename)
+            assert( length(colnames(y_unsampled)) == 1)
+            assert( nrow(df) == nrow(y_unsampled))
+            df <- cbind(df, y)
+            target_name <- colnames(y)
+        }
+    df <-  df[!(is.na(df[target_name])) , ]
+    X <-  df[,!(names(df) %in% c(target_name))]
+    X <- X[sample(nrow(X), size=num_rows, replace=TRUE ), ]
```
jlpolit, line 55: am I understanding correctly that we always resample the data with replacement, even if we use the full dataset? is there a reason for this?

4dahalibut: In python, we sample with replacement so that if a user asks for 100 rows we don't fail, but instead draw from the data that does exist.

custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
```diff
@@ -24,6 +24,69 @@ init <- function(code_dir) {
...
+  if (!is.null(weights_filename)){
+    set.seed(1)
+    row_weights <- load_data(weights_filename)
+    row_weights <- row_weights[sample(nrow(row_weights), size=num_rows, replace=TRUE ), ]
+  } else if(!is.null(weights)){
+    if (! weights %in% colnames(X)){
+      stop( paste("The column name",
+                  weights,
+                  "is not one of the columns in your training data", sep=' '))
+    }
```
4dahalibut, line 82: did you verify this works?

custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
```diff
@@ -24,6 +24,69 @@ init <- function(code_dir) {
...
+
+    return (list('X' = X, 'y' = y, 'num_rows' = num_rows))
+
+}
+
+
+process_weights <- function(X, weights_filename, weights, num_rows){
+  if (!is.null(weights_filename)){
+    set.seed(1)
```
4dahalibut, line 74: whats this do?

custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
```diff
@@ -24,6 +24,69 @@ init <- function(code_dir) {
 }
 
 
+load_data <- function(input_filename){
+    tmp = readChar(input_filename, file.info(input_filename)$size)
+    data <- read.csv(text=gsub("\r","", tmp, fixed=TRUE))
+    data
```
4dahalibut, line 30: you could just return the read.csv line right

custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
```diff
@@ -24,6 +24,69 @@ init <- function(code_dir) {
...
+
+process_data <- function(input_filename, target_filename, target_name, num_rows){
+    set.seed(1)
+    # read X
+    df <- load_data(input_filename)
+    # set num_rows
+    if (num_rows == 'ALL'){
+        num_rows <- nrow(df)
+    } else {
+        num_rows <- as.integer(a)
```
4dahalibut, line 42: what is a?

custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
jlpolit: it looks like rather than passing the random seed to individual functions, you set.seed in its own call in R - I think actually I can put this once at the top...will confirm that

custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
jlpolit: yeah I think you're right - will confirm

custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
jlpolit: lol great question - will fix

custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
```diff
@@ -24,6 +24,69 @@ init <- function(code_dir) {
...
+    }
+    # handle target
+    if (!is.null(target_filename) || !is.null(target_name)){
```
4dahalibut, line 45: I love how this code is just like, an absolute port of the python code.

custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
jlpolit: yep, tested locally by passing --row-weights fakeColumn and got "The column name fakeColumn is not one of the columns in your training data"

custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
jlpolit: I mean....I'm happy to try and get it to be more R-ish, but also I'm not 100% sure what making it more Rish even looks like?

jlpolit: maybe @yakov-g  has some thoughts?

custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
```diff
@@ -24,6 +24,69 @@ init <- function(code_dir) {
...
+        if (!is.null(target_filename)){
+            y_unsampled = load_data(target_filename)
```
yakov-g, line 47: replace = to <-

custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
```diff
@@ -24,6 +24,69 @@ init <- function(code_dir) {
...
+            stopifnot( length(colnames(y_unsampled)) == 1)
+            stopifnot( nrow(df) == nrow(y_unsampled))
```
yakov-g, line 49: remove extra whitespaces

custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
```diff
@@ -24,6 +24,69 @@ init <- function(code_dir) {
...
+            df <- cbind(df, y)
+            target_name <- colnames(y)
+        }
+    df <-  df[!(is.na(df[target_name])) , ]
```
yakov-g, line 53: remove extra whitespaces

custom_model_runner/datarobot_drum/resource/components/Python/r_fit/fit.R
```diff
@@ -24,6 +25,66 @@ init <- function(code_dir) {
 }
 
 
+load_data <- function(input_filename){
+    tmp = readChar(input_filename, file.info(input_filename)$size)
+    read.csv(text=gsub("\r","", tmp, fixed=TRUE), na.strings = c("NA", ""))
```
jlpolit, line 30: I don't like this- R by default just looks for 'NA' as a null value...and nothing else

jlpolit:
( @yakov-g @4dahalibut @meltzg any thoughts welcome- we can also discuss at standup tomorrow)

ok, so here's a problem....when we run a predict test after drum fit, we pull the data in using pandas, take out the target, then dump to a temp location in a csv (`custom_model_runner/datarobot_drum/drum/drum.py` line 291.

The issue is that pandas and R deal with missing column names differently; pandas does something like "Unnamed: 0", and R basically fills it with X (along with other wacky behavior for other "invalid" column names).

Which means that any dataset with a missing colname (like, say, telecoms churn) is going to lead to an issue where the original R model was trained on our X column (or to omit the X column, etc), but our predict method will fail because it's receiving an Unnamed: 0 column post-pandas read/save.

So...yeah, that's not the best. I guess we can do that intermediate step in R as well? Or something else....

yakov-g:
And unnamed column in churn is index only.
How does DR App handle unnamed columns?
Should we enforce users to name all the features?

jlpolit:
@yakov-g  I'm guessing those failing java tests don't actually have to do with my changes?

yakov-g:
I don't know. It's just failing on master, trying to figure out.

4dahalibut (approved):

yakov-g:

yakov-g (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**8.34** USD.

PR #154, opened 2020-10-07
Lior/raptor 3665

liororama:

## Summary
* Adding --memory  
* Adding r2d2 main for ease of use

### DIFF
--- a/tests/drum/drum_server_utils.py
+++ b/tests/drum/drum_server_utils.py
@@ -62,22 +62,27 @@ class DrumServerRun:
         with_error_server=False,
         show_stacktrace=True,
         nginx=False,
+        memory=None,
+        fail_on_shutdown_error=True,
     ):
         port = CMRunnerUtils.find_free_port()
-        server_address = "localhost:{}".format(port)
+        self.server_address = "localhost:{}".format(port)
         url_host = os.environ.get("TEST_URL_HOST", "localhost")
+
         if docker:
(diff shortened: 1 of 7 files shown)


yakov-g:

yakov-g:

tests/drum/drum_server_utils.py
```diff
@@ -62,22 +62,27 @@ def __init__(
         with_error_server=False,
         show_stacktrace=True,
         nginx=False,
+        memory=None,
+        fail_on_shutdown_error=True,
```
yakov-g, line 66: Shouldn't   `fail_on_shutdown_error` var be called smth like `shutdown_on_fail` ?

yakov-g (changes requested):

tests/drum/drum_server_utils.py
liororama: The idea here is not to fail on a shutdown error.. 
In the   test I added I am actually killing the container while inside the the scope of the context manager. The issue is that when going out of the scope of the context manager we go into the __exit__ function and then we try to send a shutdown request to the server in the docker (in case of nginx=None). Since the docker is down (we just killed it by asking r2d2 to allocate 1000mb where --memory was 500).. the request is throwing an exception..

yakov-g (approved):

meltzg (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**2.47** USD.

PR #157, opened 2020-10-08
[RAPTOR-3574] unstructured mode - drop 2

yakov-g:

## Summary
Unstructured mode drop -2 .
Implement logic according to the specs
Main points:
1. Implement `score_unstructured(model, **kwargs)`  hook according to specs.
kwargs == {"data": data(bytes/str), "mimetype": mimetype, "charset": charset, ....any other params....}

return {"data": data(bytes/str), "mimetype", "charset":charset}

> Note: according to the spec we should be passing data in either 'data' field or "text". But last minute insight opened our with @klichukb eyes and implemented it in one param

2. Make --target-type arg optional. Target type must be provided in either --target-type or model_config file.


yakov-g:

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
```diff
@@ -104,3 +103,56 @@ def predict(self, input_filename):
                 logger.error(error_message)
                 raise DrumCommonException(error_message)
         return py_data_object
+
+    def predict_unstructured(self, **kwargs):
```
klichukb, line 107: As discussed, for now we'll go with directly passing **kwargs to the python's R proxy function.
Would be great to avoid this transformation alltogehter because string builders are hard when it comes to various data. For example this one lacks quote escaping. We'll see if we can find a more durable solution.

For now this transformation would be used for arbitrary data, for example query args or web headers.

yakov-g:

meltzg (approved):

custom_model_runner/datarobot_drum/drum/args_parser.py
```diff
@@ -428,13 +428,23 @@ def _reg_args_monitoring(*parsers):
...
+            )
+
     @staticmethod
     def _reg_arg_target_type(*parsers):
         target_types = [e.value for e in TargetType]
         for parser in parsers:
             parser.add_argument(
                 ArgumentsOptions.TARGET_TYPE,
-                required=True,
+                required=False,
```
klichukb, line 447: If its not required anymore, what means None then?

custom_model_runner/datarobot_drum/drum/language_predictors/python_predictor/python_predictor.py
```diff
@@ -63,3 +50,17 @@ def predict(self, input_filename):
         self.monitor(input_filename, predictions, execution_time_ms)
 
         return predictions
+
+    def predict_unstructured(self, **kwargs):
+        # validating_input
+        data_text = kwargs.get("text", None)
```
klichukb, line 56: Line of code that I referred to that has both text/data. Commenting since we agreed to only use data.

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
```diff
@@ -104,3 +104,42 @@ def predict(self, input_filename):
...
+
+        # if data_binary_or_text is str it will be auto converted into R character type;
+        # otherwise if it is bytes, manually convert it into byte vector (raw)
+        r_data_binary_or_text = data_binary_or_text
+        if isinstance(data_binary_or_text, bytes):
+            r_data_binary_or_text = ro.rinterface.ByteSexpVector(data_binary_or_text)
+
+        kwargs_filtered = {k: v for k, v in kwargs.items() if v is not None}
+        kwargs_filtered[UnstructuredDtoKeys.DATA] = r_data_binary_or_text
+        ret_rlist = r_handler.predict_unstructured(model=self._model, **kwargs_filtered)
```
klichukb, line 142: To be honest since data is the only key now, i would probably move it out of kwargs since most users will want to use it right away. What do you think?

I mean I would have then the main interface be score_unstructured(model, body, **kwargs) in both Python and R

custom_model_runner/datarobot_drum/drum/language_predictors/python_predictor/python_predictor.py
yakov-g: fixed

custom_model_runner/datarobot_drum/drum/utils.py
```diff
@@ -171,3 +171,21 @@ def handle_missing_colnames(df):
         missing_lookup = {pycol: rcol for pycol, rcol in zip(missing_cols, r_vals)}
         return df.rename(columns=missing_lookup)
     return df
+
+
+def split_params_to_dict(params):
```
klichukb, line 176: Is this parsing content type header? lets use `from werkzeug.http import parse_options_header`

klichukb: Ah, its using command line params. I dont really see point of reinventing parser inside of parser, we can just make them separate parameters and dont parse anything manually

yakov-g: it parses both: params from cli and params from content-type

custom_model_runner/datarobot_drum/resource/components/Python/prediction_server/prediction_server.py
```diff
@@ -92,6 +86,19 @@ def predict():
...
             return response, response_status
 
+        @model_api.route("/predictUnstructured/", methods=["POST"])
+        def predict_unstructured():
+            logger.debug("Entering predict() endpoint")
+
+            self._stats_collector.enable()
+            self._stats_collector.mark("start")
+
+            response, response_status = self.do_predict_unstructured(logger=logger)
```
klichukb, line 96: I would organise this as a context manager or at least wrap the do_predict_unstructured into try/finally to handle stats.

```
with stats_collector():
    response, response_status = self.do_predict_unstructured(logger=logger)
```

klichukb: it is a good pattern to avoid cases when you called enable/start but didnt call finish/disable because of exception

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
yakov-g: **DONE**

custom_model_runner/datarobot_drum/drum/args_parser.py
yakov-g: Target type itself is required, but it can be obtained from either cli or model config file.
So default None is needed to correctly resolve final value

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
yakov-g: ***DONE***

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
```diff
@@ -104,3 +104,68 @@ def predict(self, input_filename):
...
+    def predict_unstructured(self, data, **kwargs):
+        def _r_is_character(r_val):
+            _is_character = ro.r("is.character")
+            return bool(_is_character(r_val))
+
+        def _r_is_raw(r_val):
+            _is_raw = ro.r("is.raw")
+            return bool(_is_raw(r_val))
+
+        def _r_is_null(r_val):
```
klichukb, line 117: doesnt rpy2 just provide us with robjects.NULL object? we can just check the object instead of running R expression..

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
```diff
@@ -104,3 +104,68 @@ def predict(self, input_filename):
...
+                return str(r_val[0])
+            else:
+                raise DrumCommonException(
+                    "Can not convert R value {} type {}".format(r_val, type(r_val))
+                )
+
+        def _rlist_to_dict(rlist):
+            if _r_is_null(rlist):
+                return None
+            return {str(k): _cast_r_to_py(v) for k, v in rlist.items()}
```
klichukb, line 138: Is it even possible that R will return a list where you get explicit NULL value? i dont think so

Whenever you assign NULL to a list's field, it becomes unavailable, I dont think you'll get something like 
```
list(mimetype=None, charset='utf-8')
``` 
here, instead
```
list(charset='utf-8')
```

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
```diff
@@ -104,3 +104,68 @@ def predict(self, input_filename):
                 logger.error(error_message)
                 raise DrumCommonException(error_message)
         return py_data_object
+
+    def predict_unstructured(self, data, **kwargs):
+        def _r_is_character(r_val):
```
klichukb, line 109: we'll need to make sure those internal functions are covered with tests currently its hard to (they are nested), only end-to-end calling predict_unstructured.

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
```diff
@@ -104,3 +104,68 @@ def predict(self, input_filename):
...
+            _is_character = ro.r("is.character")
+            return bool(_is_character(r_val))
+
+        def _r_is_raw(r_val):
```
klichukb, line 113: we dont need to run R interpreteter if value is bytes or string, its returned back as a typed rpy2 proxy, like StrVector or ByteVector, we check pythonic type of the proxy instead of running interpreter..

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
```diff
@@ -104,3 +104,68 @@ def predict(self, input_filename):
...
+            return {str(k): _cast_r_to_py(v) for k, v in rlist.items()}
+
+        data_binary_or_text = data
+
+        if UnstructuredDtoKeys.QUERY in kwargs:
+            if isinstance(kwargs[UnstructuredDtoKeys.QUERY], dict):
+                query_kwargs = kwargs[UnstructuredDtoKeys.QUERY]
+                for k, v in query_kwargs.items():
+                    if v is None:
+                        query_kwargs[k] = ro.rinterface.NULL
```
klichukb, line 147: there's no point passing Null values to R since they are unavailable as keys at all there, they should be skipped afaik

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
```diff
@@ -104,3 +104,68 @@ def predict(self, input_filename):
...
+        # otherwise if it is bytes, manually convert it into byte vector (raw)
+        r_data_binary_or_text = data_binary_or_text
+        if isinstance(data_binary_or_text, bytes):
+            r_data_binary_or_text = ro.vectors.ByteVector(data_binary_or_text)
+
+        kwargs_filtered = {k: v for k, v in kwargs.items() if v is not None}
+        one_or_list = r_handler.predict_unstructured(
+            model=self._model, data=r_data_binary_or_text, **kwargs_filtered
+        )
+        if isinstance(one_or_list, ro.vectors.ListVector):
```
klichukb, line 160: this is the type of check that should be done instead of is_raw/is_character and so on

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/score.R
```diff
@@ -232,3 +232,36 @@ outer_predict <- function(input_filename, target_type, model=NULL, positive_clas
...
+        .is_instance_character_raw_null <- function(v) {
+            is.character(v) || is.raw(v) || is.null(v)
+        }
+
+        .is_instance_list_none <- function(v) {
+            is.list(v) || is.null(v)
+        }
+
+
+        if (.is_instance_character_raw_null(to_validate)) {
```
klichukb, line 247: this IF is empty

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
klichukb: also query arg value likely cant be None

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
```diff
@@ -104,3 +104,68 @@ def predict(self, input_filename):
...
+            first = _cast_r_to_py(one_or_list[0])
+            second = _rlist_to_dict(one_or_list[1])
+            ret = first, second
+        elif _r_is_null(one_or_list) or _r_is_character(one_or_list) or _r_is_raw(one_or_list):
```
klichukb, line 164:
```
else:
   ret = _case_r_to_py(one_or_list), None
```
let it crash
or wrap with try/except and throw a different error message

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/score.R
```diff
@@ -232,3 +232,36 @@ outer_predict <- function(input_filename, target_type, model=NULL, positive_clas
...
+
+        } else if (is.list(to_validate)) {
+            if (length(to_validate) != 2){
+                stop(sprintf("In unstructured mode predictions of type list must have length = 2, got %s", length(to_validate)))
+            }
+            par1 <- to_validate[[1]]
+            par2 <- to_validate[[2]]
+            if (!.is_instance_character_raw_null(par1) || !.is_instance_list_none(par2)) {
```
klichukb, line 255: can you even have a list where one of the items is null?

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -338,11 +342,45 @@ def predict(self, input_filename, model=None, **kwargs):
...
+
+    @staticmethod
+    def _validate_unstructured_predictions(unstructured_response):
+        def _is_instance_str_bytes_none(data):
+            return isinstance(data, (bytes, str, type(None)))
+
+        def _is_instance_dict_none(data):
+            return isinstance(data, (dict, type(None)))
+
+        if _is_instance_str_bytes_none(unstructured_response):
```
klichukb, line 357: pass??

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -338,11 +342,45 @@ def predict(self, input_filename, model=None, **kwargs):
...
+                raise ValueError(
+                    "In unstructured mode predictions of type tuple must have length = 2, got {}".format(
+                        len(unstructured_response)
+                    )
+                )
+            par1 = unstructured_response[0]
+            par2 = unstructured_response[1]
+            if not _is_instance_str_bytes_none(par1) or not _is_instance_dict_none(par2):
+                raise ValueError(
+                    "In unstructured mode predictions tuple must be of type (str, dict) but received ({}, {})".format(
```
klichukb, line 370: code says that message should say "str/bytes, dict"

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -338,11 +342,45 @@ def predict(self, input_filename, model=None, **kwargs):
...
+                        type(par1), type(par2)
+                    )
+                )
         else:
-            self._validate_predictions(predictions, positive_class_label, negative_class_label)
+            raise ValueError(
+                "In unstructured mode predictions must be of type str or tuple; but received {}".format(
```
klichukb, line 376: str ->  str/bytes

custom_model_runner/datarobot_drum/resource/components/Python/generic_predictor/generic_predictor.py
```diff
@@ -53,11 +58,43 @@ def configure(self, params):
...
+        additional_params = self._params.get("query_params")
+
         if self._target_type == TargetType.UNSTRUCTURED:
-            with open(output_filename, "w") as f:
-                f.write(predictions)
+            kwargs_params = {}
+
+            # trick in order not to write own parser for params using
+            # "name1=value1;name2=value2" schema
+            _, params = werkzeug.http.parse_options_header("_;" + additional_params)
```
klichukb, line 70: we should not be using header parser for query.

either param per query value or a real query parser, consistent with web browser.

```
[ins] In [4]: from urllib.parse import parse_qs
[ins] In [5]: parse_qs("search=quint&tags=python")
Out[5]: {'search': ['quint'], 'tags': ['python']}
```

custom_model_runner/datarobot_drum/resource/components/Python/generic_predictor/generic_predictor.py
```diff
@@ -53,11 +58,43 @@ def configure(self, params):
...
+            )
+
+            # only for screen printout convenience we take pred data directly from unstructured_response
+            if isinstance(ret_data, bytes):
+                with open(output_filename, "wb") as f:
+                    f.write(ret_data)
+            else:
+                if ret_data is None:
+                    ret_data = "Return value from prediction is: None (NULL in R)"
+                with open(output_filename, "w") as f:
```
klichukb, line 94: be aware that default encoding here is platform dependant. utf-8 on linux

custom_model_runner/datarobot_drum/drum/model_adapter.py
klichukb, line 357:
```
# common denominator, normalize to tuple
if not isinstance(unstructured_response, tuple):
    unstructured_response = (unstructured_response, None)
if len(unstructured_response) != 2:
    raise SpecificError()
body, kwargs = unstructured_response

# this is much easier to read inline than functions that have names almost the same as the code in them
if not isinstance(body, (bytes, str, type(None))) or not isinstance(kwargs, (dict, type(None))):
    raise SpecificError
```

yakov-g:

tests/drum/test_other_cases.py
```diff
@@ -239,37 +241,63 @@ def test_set_language(
...
         )
-        preds = pd.read_csv(output)
-        assert all(
-            val for val in (preds["Predictions"] == len(CustomHooks.ALL_PREDICT)).values
-        ), preds
+        if hooks_list == CustomHooks.ALL_PREDICT_STRUCTURED:
+            preds = pd.read_csv(output)
+            assert all(val for val in (preds["Predictions"] == len(hooks_list)).values), preds
+        elif hooks_list == CustomHooks.ALL_PREDICT_UNSTRUCTURED:
+            with open(output) as f:
```
klichukb, line 298: this uses a default encoding which is platform dependant, which is utf-8.

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
yakov-g: I thought about this, but rpy2 proxy for character object is `ndarray`, that's why I prefer R way.

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/score.R
yakov-g: yes, as we allow valuer to be None, str, bytes, tuple; it's more clean that to do:
```
if not Null, str, bytes tuple:
   error
else if tuple
else if str or bytes
....
etc
```

klichukb (approved):
I feel like there is code for Py-R translation that is not needed, can be simplified, when it comes to examining converted types and working with null.
Advice for the future: the PR is too big, it is hard to verify, size problem is partially because of the feature "query" that
1) is not highest priority
2) does have suboptimal implementation
"query" implementation creates a pressure to merge PR because its urgent, so "query" has to tag along.

Did not test personally.
I have to approve, but would be good if improvements are made before this goes live.

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/score.R
yakov-g:
```
> list("a", NULL)
[[1]]
[1] "a"

[[2]]
NULL
```

custom_model_runner/datarobot_drum/drum/model_adapter.py
yakov-g: The same flow as in score.R

custom_model_runner/datarobot_drum/drum/model_adapter.py
yakov-g: This error is shown when tuple (str, int) is returned.
So I mean if you try to return tuple, it has to be (str, dict) tuple

custom_model_runner/datarobot_drum/drum/model_adapter.py
yakov-g: ***DONE***

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
yakov-g: ***Done***

tests/drum/test_other_cases.py
yakov-g: The test saves this file in utf8.

custom_model_runner/datarobot_drum/resource/components/Python/generic_predictor/generic_predictor.py
yakov-g: FIxed: save file with user's default encoding

custom_model_runner/datarobot_drum/resource/components/Python/generic_predictor/generic_predictor.py
yakov-g: switched to
```
>>> dict(parse_qsl("name1=val1;name2=[val2,val3]"))
{'name1': 'val1', 'name2': '[val2,val3]'}
```

custom_model_runner/datarobot_drum/drum/model_adapter.py
yakov-g: ***DONE***

custom_model_runner/datarobot_drum/drum/model_adapter.py
yakov-g: ***Fixed as you suggested***

custom_model_runner/datarobot_drum/drum/model_adapter.py
yakov-g: ***Fixed as you suggested***

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/score.R
yakov-g: ***Fixed as you suggested***

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/score.R
yakov-g: ***Fixed as you suggested***

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
yakov-g: You are right.
Not relevant, after I changed parser.
This code is removed.

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
yakov-g: changed this code, so comment not relevant

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
yakov-g: added toTODO

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
yakov-g: added TODO

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
yakov-g: added TODO

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**21.1** USD.

PR #160, opened 2020-10-10
[RAPTOR-3554] show error if uwsgi is not available in production mode

yakov-g:

## Summary
Show error if uwsgi is not available in production mode, which is the case on Windows machines.
Tested manually

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/args_parser.py
+++ b/custom_model_runner/datarobot_drum/drum/args_parser.py
@@ -3,6 +3,7 @@ from argparse import RawTextHelpFormatter
 import os
 from datarobot_drum.drum.push import HELP_TEXT
 import sys
+import subprocess
 
 from datarobot_drum.drum.description import version
 from datarobot_drum.drum.common import LOG_LEVELS, ArgumentsOptions, RunLanguage, TargetType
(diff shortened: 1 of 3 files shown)


meltzg (approved):

custom_model_runner/datarobot_drum/drum/args_parser.py
```diff
@@ -599,5 +600,15 @@ def verify_options(options):
             if not options.new_mode:
                 CMRunnerArgsRegistry._parsers[ArgumentsOptions.NEW].print_help()
                 exit(1)
+        elif options.subparser_name in [ArgumentsOptions.SERVER, ArgumentsOptions.PERF_TEST]:
+            if options.production:
+                ret_code = subprocess.run(["python", "-m", "pip", "show", "uwsgi"]).returncode
```
liororama, line 605: You might be using a different python then the currently used python.. the "python argument should be the exact currently used python.. .. can you just try to load uwsgi module in the current context?

liororama (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**3.28** USD.

PR #163, opened 2020-10-15
Flag potential improvement

GlobalMin:
### Hoping to cut down on command line typing if that would get used
**Current situation:**
DRUM fit requires or suggests relative path inputs that could easily be removed and it looks like we are already doing this standard path/file checking on the testing side, so if this would be useful i can make a real tweak to the path input requirements.

> Paths are relative to `datarobot-user-models` root:  
> `drum fit  --code-dir model_templates/training/python3_xgboost --input tests/testdata/boston_housing.csv --target MEDV`  
> Check that `artifact.pkl` file was created.

**Idea for improved user experience:**
Would a makefile or another route be the most logical to turn the above into this? Regardless that should be straightforward I'm guessing but the new syntax could be this instead.

> 
> Paths are relative to `datarobot-user-models` root:  
> `drum fit  --code-dir python3_xgboost --boston_housing.csv --target MEDV`  
> Check that `artifact.pkl` file was created.

### DIFF
--- a/model_templates/training/python3_xgboost/utils.py
+++ b/model_templates/training/python3_xgboost/utils.py
@@ -1,0 +1,80 @@
+import os
+import glob
+import shutil
+import subprocess
+
+
+from .constants import PYTHON, R, R_ALL_PREDICT_STRUCTURED_HOOKS, R_FIT
+
+
+def _create_custom_model_dir(
+    resources,
+    tmp_dir,
+    framework,
+    problem,
+    language,
+    is_training=False,
+    nested=False,
+):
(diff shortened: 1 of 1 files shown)


GlobalMin:
@meltzg @yakov-g @liororama if an Issues object here is more suitable given the lack of code implementation for my idea, I can do that instead. If worth looking at I can also keep this open to add simple idea to once someone shares feedback.

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**0.01** USD.

PR #167, opened 2020-10-18
Switch R envs to flask because of OOM

yakov-g:

## Summary
R nginx started to fail with OOM
Switch to Flask

### DIFF
--- a/dev_env/rlang_nginx_env/start_server.sh
+++ b/dev_env/rlang_nginx_env/start_server.sh
@@ -4,7 +4,7 @@ export PYTHONPATH=/opt/code
 
 TARGET_TYPE="regression"
 
-CMD="drum server -cd . --address 0.0.0.0:8080 --production --max-workers 1 --show-stacktrace"
+CMD="drum server -cd . --address 0.0.0.0:8080 --with-error-server"
 
 if [ ! -z "${POSITIVE_CLASS_LABEL}" ]; then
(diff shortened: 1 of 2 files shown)


meltzg (approved):
why were we installing vim in the first place?

yakov-g:
> why were we installing vim in the first place?

This is dev env, so I wanted to install vim

yakov-g:
jarvis please test functional_drop_in

meltzg (approved):

liororama:
What kind of failure we are talking about ? if we are using a single worker then NGINKS is not talking much more memory then the flask? And I prefer our default environments to use nginx and not flask server.

yakov-g:
> What kind of failure we are talking about ? if we are using a single worker then NGINKS is not talking much more memory then the flask? And I prefer our default environments to use nginx and not flask server.

I don't know what has happened: 
the flow was.
1. Smth has changed in R repos, that required to rebuild R base image. Docker file hasn't been changed, I only recreated the image.
2. Tests started to fail.
3. Tests in R nginx env fail both on jenkins and locally
4. Tests in R flask env fail locally, but do not fail in jenkins.
5. I want to merge this to unblock development and investigate.

liororama (approved):
As long as this is for unblocking development I am ok with this.

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**2.49** USD.

PR #174, opened 2020-10-20
Deps up: pandas==1.1.0, numpy==1.19.0

yakov-g:

## Summary
Bumping deps version in a couple of places i missed.

### DIFF
--- a/public_dropin_environments/python3_pmml/requirements.txt
+++ b/public_dropin_environments/python3_pmml/requirements.txt
@@ -1,3 +1,3 @@
 pypmml==0.9.6
 numpy==1.19.0
-pandas==1.0.3
+pandas==1.1.0
--- a/tests/fixtures/cmrun_docker_env/requirements.txt
+++ b/tests/fixtures/cmrun_docker_env/requirements.txt
@@ -1,4 +1,4 @@
-numpy==1.16.4
-pandas==0.24.2
+numpy==1.19.0
+pandas==1.1.0
 scikit-learn==0.23.1
 sagemaker-scikit-learn-extension==1.1.0


klichukb:
I believe as we go further this would be considered a really breaking change. Not sure about now :)

klichukb:
Might requalify it to "major", not "breaking" since it would affect only new models.

yakov-g:
To tell the truth, I look at these envs only as envs for our model templates - examples.
If anyone needs his specific requirements - provide them with your model.

A couple of hours ago I merged a PR where I bumped pandas in several places, because I wanted to upgrade rpy2 and unify dependencies.

Having different deps across envs also doesn't look clear clear to me.
So this PR is only bump in two places as I bumped in 20 other places before.

meltzg:
@yakov-g these envs are not templates/examples. they're part of the product. these are the exact environments that are installed as public envs

klichukb:
Kinda of agree that its only template, but at the same time, I meant that they are then ending up in dR cloud as default envs to go. 

No problem at this point, would be interesting to see if anything break

yakov-g:
I meant these are envs for our model templates at first place.
I agree with you both, they are part of the product, but in this case we shouldn't touch them at all. And all our newly introduced examples (model templates) shouldn't change these envs. Thus we need to create separate envs where our examples run.

Maybe we even should go further:
have 3 public envs py/java/r and move requirements into models. This will also remind CFDSs and customers to provide proper requirements with the model and not to rely on env. (I'm not sure if dependency management is available for self service)

yakov-g:
@meltzg @klichukb 
And could you approve?

meltzg (approved):

rvorobii (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**2.47** USD.

PR #176, opened 2020-10-20
[RAPTOR-3576] handle the case where standalone keras is not installed

yakov-g:

## Summary
Handle a case when standalone keras package is not installed.

A couple of days ago I removed support for standalone keras. So drum integration tests are supposed to work in the env without standalone keras, but tensorflow>=2.2.1. Tests didn't fail because old test container was in use.
Today I updated integration tests container, so there is no keras package in it.

This code is still there only for keras vizai case in functional tests which doesn't work with tf.keras (@rsugumar is expected to help).


meltzg (approved):

liororama (approved):
do you want to print anything for each case?

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**1.45** USD.

PR #179, opened 2020-10-21
[RAPTOR-3574] don't assign default charset if mimetype is binary

yakov-g:

## Summary
Some cleanup for `score_unstructured` hook.

Request (and related batch functionality):
In unstructured mode we treat incoming mimetype that startswith `text/` or `application/json` or "" or missing as "textual"

If user sends request with "textual" mimetype, but charset is missing, we set default charset to `utf8`.
If user sends request with non textual mimetype, but charset is missing, we don't pass charset param into the hook.

Response (and related batch functionality):
Outgoing
If user returns string, but doesn't provide mimetype/charset, we set them to `text/plain` and `utf8`
If user returns binary, but doesn't provide mimetype/charset, we set them to `application/octet-stream` and `None`

### DIFF
--- a/custom_model_runner/datarobot_drum/resource/unstructured_helpers.py
+++ b/custom_model_runner/datarobot_drum/resource/unstructured_helpers.py
@@ -11,15 +11,19 @@ def _is_text_mimetype(mimetype):
 
 
 def _resolve_incoming_unstructured_data(in_data, mimetype, charset):
-    ret_mimetype = mimetype if mimetype is not None and mimetype != "" else MIMETYPE_TEXT_DEFAULT
-    ret_charset = charset if charset is not None else CHARSET_DEFAULT
-
     if not isinstance(in_data, bytes):
         raise DrumCommonException("bytes data is expected, received {}".format(type(in_data)))
 
+    # Incoming mimetype that startswith `text/` or `application/json` or "" or missing is considered "textual".
+    # If user sends request with "textual" mimetype, but charset is missing, we set default charset to `utf8`.
+    # If user sends request with non textual mimetype, but charset is missing, we don't pass charset param into the hook.
+    ret_mimetype = mimetype if mimetype is not None and mimetype != "" else MIMETYPE_TEXT_DEFAULT
+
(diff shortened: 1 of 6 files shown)


custom_model_runner/datarobot_drum/resource/unstructured_helpers.py
```diff
@@ -28,14 +32,17 @@ def _resolve_incoming_unstructured_data(in_data, mimetype, charset):
...
-    ret_mimetype = ret_kwargs.get(
-        UnstructuredDtoKeys.MIMETYPE,
-        MIMETYPE_TEXT_DEFAULT
-        if isinstance(ret_data, (str, type(None)))
-        else MIMETYPE_BINARY_DEFAULT,
-    )
-    if isinstance(ret_data, str):
-        ret_data = ret_data.encode(ret_charset)
+
+    # in this function incoming ret_data expected to be either str or bytes
```
rvorobii, line 36:
```suggestion
    # In this function incoming ret_data expected to be either str or bytes
```

tests/drum/test_unstructured_mode.py
```diff
@@ -227,25 +228,37 @@ def test_response_one_var_return(
             data = None
             headers = {"Content-Type": "text/plain; charset=UTF-8"}
             response = requests.post(url=url, data=data, params=params, headers=headers)
-            content_type_header = response.headers["Content-Type"]
+
```
rvorobii, line 231: you can also add tests to
1. send text data with no charset
2. send text data with non-utf-8 charset (since utf-8 is a default one)
3. send binary data with encoding

rvorobii (approved):

tests/drum/test_unstructured_mode.py
yakov-g: Added more test cases, not sure all you mention. :)

meltzg (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**3.95** USD.

PR #180, opened 2020-10-21
[RAPTOR-3700] Sparse Drum

4dahalibut:

## Summary
We're passing matrix market files now from DataRobot, so this adds support and a new Sklearn template that can receive sparse data.


custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/score.R
```diff
@@ -211,6 +211,8 @@ outer_predict <- function(input_filename, target_type, model=NULL, positive_clas
 
     if (!isFALSE(read_input_data_hook)) {
         data <- read_input_data_hook(input_filename)
+    } else if (grepl(".mtx", input_filename, fixed=TRUE)) {
```
klichukb, line 214: I would always check for "endswith()" instead of "includes()"

```
grepl("\\.mtx$", input_filename)
```

klichukb: we dont want "hello.mtxfoo" or "hello.mtx.txt" match

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -265,10 +266,11 @@ def _validate_predictions(self, to_validate, class_labels):
     @staticmethod
     def _read_structured_input(filename):
         try:
-            df = pd.read_csv(filename, lineterminator="\n")
+            if ".mtx" in filename:
```
klichukb, line 269: Same here. Would go with `filename.endswith(".mtx")`

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/score.R
4dahalibut: good catch, I'll change this.

custom_model_runner/datarobot_drum/drum/model_adapter.py
4dahalibut: will do

klichukb (commented):
Looks good, 1 comment to address (file extension match)

model_templates/training/python3_sparse/custom.py
```diff
@@ -0,0 +1,148 @@
...
+):
+    """
+    This hook must be implemented with your fitting code, for running drum in the fit mode.
+
+    This hook MUST ALWAYS be implemented for custom training models.
+    For inference models, this hook can stick around unimplemented, and won’t be triggered.
+
+    Parameters
+    ----------
+    X: pd.DataFrame - training data to perform fit on
```
jlpolit, line 26: nit- a couple of comments/docstrings here could use updating (I think just X's expect type and the comment below about "feel free to delete...")

jlpolit (approved):
LGTM

model_templates/training/python3_sparse/custom.py
4dahalibut: I don't understand what you mean by this.

jlpolit: the param for X in the docstring says we're expecting a pd.DataFrame - not sure if that's true since the fit call says it's a csr_matrix? and the line above estimator=Ridge() that says "feel free to delete whichever of these you aren't using" when there's only mod estimator path...I'm pretty sure our templates are full of little stuff like that

4dahalibut: good comments.

klichukb (approved):
thanks for addressing comments

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**8.17** USD.

PR #181, opened 2020-10-21
[RAPTOR-3768] fix process detection for memory monitoring

yakov-g:

## Summary
`psutil.proc.name()` behaves differently on Mac and Linux, so `if current_proc.name() == ArgumentsOptions.MAIN_COMMAND` statement didn't work as expected on Mac.

Also add a "label" environment variable to a drum server subprocess to be able to kill it in case `perf-test` fails


meltzg (approved):

rvorobii (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**1.27** USD.

PR #183, opened 2020-10-22
[RAPTOR-3554] hide 'pip show uwsgi' output

yakov-g:

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/args_parser.py
+++ b/custom_model_runner/datarobot_drum/drum/args_parser.py
@@ -711,7 +711,9 @@ class CMRunnerArgsRegistry(object):
                 exit(1)
         elif options.subparser_name in [ArgumentsOptions.SERVER, ArgumentsOptions.PERF_TEST]:
             if options.production:
-                ret_code = subprocess.run([sys.executable, "-m", "pip", "show", "uwsgi"]).returncode
+                ret_code = subprocess.run(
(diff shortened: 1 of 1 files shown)


meltzg (approved):

custom_model_runner/datarobot_drum/drum/args_parser.py
```diff
@@ -711,7 +711,9 @@ def verify_options(options):
                 exit(1)
         elif options.subparser_name in [ArgumentsOptions.SERVER, ArgumentsOptions.PERF_TEST]:
             if options.production:
-                ret_code = subprocess.run([sys.executable, "-m", "pip", "show", "uwsgi"]).returncode
+                ret_code = subprocess.run(
```
liororama, line 714: I prefer if we printed the output in case it is  failing.. so something like:
                if options.verbose:
                    print("Checking if uwsgi is installed")
                result = subprocess.run([sys.executable, "-m", "pip", "show", "uwsgi"],
                                        stdout=subprocess.PIPE, stderr=subprocess.PIPE)
                if result.returncode != 0:
                    print(
                        "Looks like 'uwsgi` package is missing."
                        " Don't use '{}' option when running drum server or try to install"
                        " 'uwsgi'.".format(ArgumentsOptions.PRODUCTION))
                    print(result.stdout)
                    print(result.stderr)
                    exit(1)

liororama (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**2.24** USD.

PR #186, opened 2020-10-23
[RAPTOR-3569] change start_server.sh/fit.sh to take target_type from …

yakov-g:
change start_server.sh/fit.sh to take target_type from LRS env var


rvorobii:
we need to have these environments updated to on prod only when https://github.com/datarobot/DataRobot/pull/71133 os deployed there

rvorobii (approved):
we need to have these environments updated to on prod only when datarobot/DataRobot#71133 os deployed there

yakov-g:
> we need to have these environments updated to on prod only when [datarobot/DataRobot#71133](https://github.com/datarobot/DataRobot/pull/71133) os deployed there

Good point. When does this happen?

rvorobii:
> > we need to have these environments updated to on prod only when [datarobot/DataRobot#71133](https://github.com/datarobot/DataRobot/pull/71133) os deployed there
> 
> Good point. When does this happen?

we have cloud deploys weekly, used to be on Wed
we can ask supersully from time to time in Slack `@supersully pr RAPTOR-3569`

meltzg (changes requested):
env version ids should be updated.  Also, we shouldn't merge this until the change to DR makes its way to prod

meltzg (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**4.07** USD.

PR #190, opened 2020-10-27
[RAPTOR-3835] Add more realistic multiclass examples

meltzg:

## Summary
Adding multiclass examples with requirements and multiple model artifacts

### DIFF
--- a/model_templates/inference/python3_pytorch_multiclass/custom.py
+++ b/model_templates/inference/python3_pytorch_multiclass/custom.py
@@ -32,7 +32,7 @@ def load_model(code_dir: str) -> Any:
 
     model = torch.load(os.path.join(code_dir, "artifact.pth"))
     model.eval()
-    return model    
+    return model
 
 
 def transform(data: pd.DataFrame, model: Any) -> pd.DataFrame:
@@ -92,7 +92,7 @@ def fit(
     -------
     Nothing
     """
-    
+
     print("Fitting Preprocessing pipeline")
     preprocessor = dense_preprocessing_pipeline.fit(X)
     lb = LabelEncoder().fit(y)
--- a/model_templates/inference/python3_pytorch_multiclass/custom.py
+++ b/model_templates/inference/python3_pytorch_multiclass/custom.py
@@ -11,7 +11,6 @@ from preprocessing import dense_preprocessing_pipeline
 from model_utils import build_classifier, train_classifier, save_torch_model
 
 preprocessor = None
-expected_columns = None
 
 
 def load_model(code_dir: str) -> Any:
@@ -28,15 +27,12 @@ def load_model(code_dir: str) -> Any:
     If used, this hook must return a non-None value
     """
     global preprocessor
-    global expected_columns
     with open(os.path.join(code_dir, "preprocessor.pkl"), mode="rb") as f:
         preprocessor = pickle.load(f)
-    with open(os.path.join(code_dir, "expected_columns.pkl"), mode="rb") as f:
-        expected_columns = pickle.load(f)
 
     model = torch.load(os.path.join(code_dir, "artifact.pth"))
     model.eval()
-    return model
+    return model    
 
 
 def transform(data: pd.DataFrame, model: Any) -> pd.DataFrame:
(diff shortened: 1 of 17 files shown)


model_templates/inference/python3_pytorch_multiclass/custom.py
```diff
@@ -0,0 +1,115 @@
...
+    -------
+    Transformed data
+    """
+    if preprocessor is None:
+        raise ValueError("Preprocessor not loaded")
+
+    return preprocessor.transform(data)
+
+
+def fit(
```
4dahalibut, line 59: why is there a fit method in the inference template though.

model_templates/training/python3_pytorch_multiclass/custom.py
```diff
@@ -0,0 +1,115 @@
...
+    Parameters
+    ----------
+    data : is the dataframe given to **drum** to make predictions on
+    model : is the deserialized model loaded by **drum** or by `load_model`, if supplied
+
+    Returns
+    -------
+    Transformed data
+    """
+    if preprocessor is None:
```
4dahalibut, line 53: oof that is notgreat

model_templates/inference/python3_pytorch_multiclass/custom.py
meltzg: the model was fit with drum. figured it made sense to keep

model_templates/training/python3_pytorch_multiclass/custom.py
meltzg: ?

4dahalibut: cant you pass it through the return for load_model.

meltzg: That'd require me to then implement my own hook for `score`

meltzg: which to me feels like overkill in this case

4dahalibut: alright. I still dont like that we're advertising this as best practice though right.

model_templates/inference/python3_pytorch_multiclass/custom.py
4dahalibut: But what about all the other inference templates

meltzg: none of them were fit with drum. they predate drum even being able to fit. I think most of them straight up predate drum

model_templates/training/python3_pytorch_multiclass/custom.py
meltzg: i get that. This seemed like the least awful approach. Other option I thought of was to load the model every time transform is called. Pros: no globals. Cons: expensive and slow

timsetsfire (approved):

model_templates/inference/python3_pytorch_multiclass/custom.py
4dahalibut: So why do we have inference and training models templates then?

model_templates/training/python3_pytorch_multiclass/custom.py
4dahalibut: I feel like this might have been a design mistake on our part to not have users build python classes instead of calling python methods.

model_templates/inference/python3_pytorch_multiclass/custom.py
meltzg: the other inference templates were not made with drum, but work with drum for scoring.

model_templates/training/python3_pytorch_multiclass/custom.py
meltzg: that may be, but is outside the scope of this PR. perhaps standalone hooks can deprecated in favor of a new class based approach in a major version of drum

alexromanyuk (approved):
:green_apple:

4dahalibut (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**4.66** USD.

PR #193, opened 2020-10-29
Change dep from numpy>=1.19.0 to numpy<1.19.0

yakov-g:

## Summary
Only tensorflow==2.2.1 requires numpy<1.19.0; but change it everywhere for consistancy.

## Rationale
Starting pip 20.3 new stricter dependency resolver planned to be enabled.
It won't allow to install packages whose version requirements don't match.
We use tensorflow==2.2.1 which requires numpy<1.19.0
    
In order to prevent problems in future, I suggest to reduce numpy version.

That's how new feature is going to behave:
`pip install tensorflow==2.2.1 numpy==1.19.0 --use-feature=2020-resolver`

```
ERROR: Cannot install numpy==1.19.0 and tensorflow 2.2.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested numpy==1.19.0
    tensorflow 2.2.1 depends on numpy<1.19.0 and >=1.16.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
```


public_dropin_environments/python3_keras/requirements.txt
```diff
@@ -1,6 +1,6 @@
 tensorflow==2.2.1
 Keras==2.3.1
-numpy==1.19.0
+numpy<1.19.0
```
klichukb, line 3: this is not a correct rule, as far as I see the rule is `numpy<1.19.0 and >=1.16.0`

Current rule would be fine with numpy 0.1 technically.

klichukb:
```
numpy>=1.16.0,<1.19.0
```

yakov-g: I suppose it will install 1.18.5 anyhow, but I'll change

yakov-g: **DONE**

klichukb (approved):

yakov-g:

meltzg (changes requested):
update env version IDs please

meltzg (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**5.8** USD.

PR #198, opened 2020-11-02
Move drum push test; add unsupervised case

jlpolit:

### DIFF
--- a/tests/drum/test_drum_push.py
+++ b/tests/drum/test_drum_push.py
@@ -6,7 +6,7 @@ import yaml
 
 from datarobot_drum.drum.common import ArgumentsOptions
 from tests.conftest import PYTHON, SKLEARN, REGRESSION, BINARY
-from ..drum.utils import _create_custom_model_dir, _exec_shell_cmd
+from tests.drum.utils import _create_custom_model_dir, _exec_shell_cmd
 
 BASE_MODEL_TEMPLATES_DIR = "model_templates"
 BASE_DATASET_DIR = "tests/testdata"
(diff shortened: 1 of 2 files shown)


yakov-g:
There is another `push` test  case in tests/drum/test_units.py
Could you please move into test_push.py

tests/drum/test_drum_push.py
```diff
@@ -40,17 +49,12 @@ def test_drum_push_training(
...
-            resources,
-            tmp_path,
-            framework,
-            problem,
-            language,
+            resources, tmp_path, framework, problem, language, is_training=True
         )
 
-        env_id, _ = sklearn_drop_in_env
+        env_id = "5e8c889607389fe0f466c72d"
```
4dahalibut, line 57: Why is this done?

jlpolit: @yakov-g requested that I move this out of the functional tests, so this was to avoid moving the sklearn_drop_in_env in the functional conftest up a level....but maybe there's no reason not to do that?

4dahalibut: if this is hitting DR it should be inside of functional tests tho

jlpolit: calling push from the command line will always hit DR, correct? in which case @yakov-g I think Josh is right that this does belong in functional tests after all....I can open a new PR to just take care of the unsupervised part of this

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**2.99** USD.

jlpolit:
closing in favor of https://github.com/datarobot/datarobot-user-models/pull/207

PR #199, opened 2020-11-03
[RAPTOR-3866] Catch exception from user's code, log it and return proper response

zohar-mizrahi:
Two major changes:
* Catch user's exceptions that might occur during predictions, log them and return a proper response
* Pump up mlpiper version dependency to mlpiper v2.4.0

**Note:** the tests for this PR will not pass until the `mlpiper v2.4.0` will be pushed to DataRobot artifactory

### DIFF
--- a/custom_model_runner/datarobot_drum/resource/components/Python/uwsgi_component/uwsgi_serving.py
+++ b/custom_model_runner/datarobot_drum/resource/components/Python/uwsgi_component/uwsgi_serving.py
@@ -1,10 +1,6 @@
 import logging
 import os
-import pandas as pd
 import sys
-import tempfile
-
-from flask import request
 
 from mlpiper.components.restful.flask_route import FlaskRoute
 from mlpiper.components.restful_component import RESTfulComponent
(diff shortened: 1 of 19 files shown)


alexromanyuk (approved):
:+1:

yakov-g:

rvorobii (approved):

yakov-g:
@rvorobii 
COuld you take a look?
With these changes  `test_e2e_predict_fails` fails: with error:
```
>           assert response.status_code == 500  # error occurs
E           assert 513 == 500
E            +  where 513 = <Response [513]>.status_code
```

Do we want to update tests?

rvorobii:
> @rvorobii
> COuld you take a look?
> With these changes `test_e2e_predict_fails` fails: with error:
> 
> ```
> >           assert response.status_code == 500  # error occurs
> E           assert 513 == 500
> E            +  where 513 = <Response [513]>.status_code
> ```
> 
> Do we want to update tests?

The tests is correct.
513 should be only returned if exception happened before prediction server started.
513 return drum initilization errors, but if prediction server is already running - initilization was fine, and there should be no 513.

custom_model_runner/datarobot_drum/resource/components/Python/uwsgi_component/uwsgi_serving.py
```diff
@@ -145,6 +145,10 @@ def predict(self, url_params, form_params):
                 # this counter is managed by uwsgi
                 self._total_predict_requests.increase()
                 self._predict_calls_count += 1
+        except Exception as ex:
+            self._logger.error(ex)
+            response_status = HTTP_513_DRUM_PIPELINE_ERROR
```
rvorobii, line 150: @yakov-g raised a good point - why 513 and not 500?
513 is designed only for errors happened during drum initilization.
prediction errors should be 500

rvorobii (changes requested):
looks like there should be an error with 500, not with 513

custom_model_runner/datarobot_drum/resource/components/Python/uwsgi_component/uwsgi_serving.py
rvorobii: also, looks like this ticket is about this issue
could you please have a look at close it if it's fixed https://datarobot.atlassian.net/browse/RAPTOR-3751?

yakov-g: thanks
I'll take a look, if @zohar-mizrahi won't

rvorobii (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**5.87** USD.

PR #200, opened 2020-11-03
[RAPTOR-3842] fix for pickling with joblib creates issue with TF2 upgrading

rsugumar:
## Summary
This PR moves the final step in the pipeline to `inference` code.

## Rationale
The `feature_extraction` step in the pipeline is moved to `inference` code, coz while pickling the artifact during the `drum fit` stage, the upgraded libraries is causing the error `TypeError: can't pickle _thread.RLock objects` (root cause of the issue is not very clear yet), however moving this step to inference code resolves this issue (as we're _skipping the pickling_ that step basically)


meltzg (approved):

yakov-g (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**5.57** USD.

PR #206, opened 2020-11-04
[RAPTOR-3899] add anomaly inference template

jlpolit:

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,6 +267,7 @@ The [model_templates](model_templates) folder contains sample models that work w
 
 ##### Inference Models
 * [Scikit-Learn sample model](model_templates/inference/python3_sklearn)
+* [Scikit-Learn sample unsupervised anomaly detection model](model_templates/inference/python3_sklearn_anomaly)
 * [PyTorch sample model](model_templates/inference/python3_pytorch)
 * [XGBoost sample model](model_templates/inference/python3_xgboost)
 * [Keras sample model](model_templates/inference/python3_keras)
--- a/model_templates/inference/python3_sklearn_anomaly/README.md
+++ b/model_templates/inference/python3_sklearn_anomaly/README.md
@@ -7,4 +7,4 @@ with no target.
 
 ### To run locally using 'drum'
 Paths are relative to `./datarobot-user-models`:   
-`drum score --code-dir model_templates/inference/python3_sklearn_anomaly -input tests/testdata/boston_housing.csv`
+`drum score --code-dir model_templates/inference/python3_sklearn_anomaly --target-type anomaly --input tests/testdata/boston_housing.csv`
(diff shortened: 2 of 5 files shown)


yakov-g:
We should a functional test as well.
And add ANOMALY target type into public API.

yakov-g:
Could you add a new `anomaly` section somewhere here
https://github.com/datarobot/datarobot-user-models/blob/master/README.md#training-models

README.md
```diff
@@ -267,6 +267,7 @@ The [model_templates](model_templates) folder contains sample models that work w
 
 ##### Inference Models
 * [Scikit-Learn sample model](model_templates/inference/python3_sklearn)
+* [Scikit-Learn sample unsupervised anomaly detection model](model_templates/inference/python3_sklearn_anomaly)
```
yakov-g, line 270: I thought as a different section:
- inference
- training
- anomaly

but up to you

yakov-g (approved):

jlpolit:
> We should a functional test as well.
> And add ANOMALY target type into public API.

the latest master of the public api should work, but I don't think that code has made it into a published version yet- will definitely need to bump api version in requirements to get a working functional test for this

yakov-g:
So does it make sense to add a test for this this template here:
https://github.com/datarobot/datarobot-user-models/blob/master/tests/functional/test_inference_model_templates.py
?

jlpolit:
> So does it make sense to add a test for this this template here:
> https://github.com/datarobot/datarobot-user-models/blob/master/tests/functional/test_inference_model_templates.py
> ?

yeah, that's what I've done locally but it won't pass until we're using an updated dr public api client (since the last published version 2.22 doesn't account for anomaly target type)

meltzg (approved):

README.md
4dahalibut: I disagree

4dahalibut (approved):

README.md
yakov-g: why

yakov-g:
According to this: https://github.com/datarobot/datarobot-user-models/blob/master/jenkins/test3_drop_in_envs.sh
and `requirements_test.txt` datarobot==2.23.0b0 expected to be used

README.md
4dahalibut: anomaly models aren't a third category, they fit the same way as multiclass within the inference training dichotomy.

jlpolit:
> According to this: https://github.com/datarobot/datarobot-user-models/blob/master/jenkins/test3_drop_in_envs.sh
> and `requirements_test.txt` datarobot==2.23.0b0 expected to be used

interesting - seemingly that version doesn't match up with what I'm seeing on the public api master: https://github.com/datarobot/public_api_client/blob/master/datarobot/models/custom_model.py#L134  

not sure when that should be expected to be published/versioned

README.md
yakov-g: you are right

jlpolit:
will follow up with a functional test once this goes in https://github.com/datarobot/public_api_client/pull/1344

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**4.34** USD.

PR #207, opened 2020-11-04
[RAPTOR-3813] test for unsupervised drum push

jlpolit:


yakov-g:
There is push test case in `tests/drum/test_units.py` 
Could you move them into test_push.py?

yakov-g:
The thing is that this test file has never been run.
Added a commit to run only this test, once it passes, I'll bring back all the others.

yakov-g (approved):

meltzg (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**3.73** USD.

PR #214, opened 2020-11-09
R boolean target 

jlpolit:

## Summary

So R has a fun thing where it thinks TRUE/FALSE is boolean while python thinks True/False is. And pandas will automatically convert TRUE to True when reading unless you tell it to read that column as a string. Which means in something like 10k diabetes, where our target values are TRUE and FALSE, pandas thinks they're a different value from R, which makes the r scoring code angry. 

I'm not convinced that this is the best solution to that problem, but it does seem to be a workable one: basically when we try to guess at the class labels, it adds an extra check for R: if pandas read a target in that's boolean, we tell it to go back and read it in as a string. 

Very open to other suggestions, since this feels like a pretty hacked-together approach to a sort of obnoxious edge case.

## Update

Turns out it makes sense to just always cast to string when we read the target column for the purpose of intuiting the labels; made that change which ends up fixing a bunch of other stuff that was breaking for non-R templates and non-boolean class labels.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/drum.py
+++ b/custom_model_runner/datarobot_drum/drum/drum.py
@@ -439,7 +439,6 @@ class CMRunner:
                 self.options.target_csv,
                 self.options.target,
                 self.options.unsupervised,
-                r_fit=run_language == RunLanguage.R,
             )
             if possible_class_labels is not None:
                 if self.target_type == TargetType.BINARY:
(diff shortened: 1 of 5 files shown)


yakov-g:
What if go the inference way and read dataset and guess labels in Python/R code itself. So we won't need to touch pandas at allw when working with R?

jlpolit:
seems weird to me that a multiclass drop-in test is failing for these changes

4dahalibut (approved):

yakov-g (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**3.02** USD.

PR #223, opened 2020-11-16
[RAPTOR-3777] 10K diabetes file with windows line terminators breaks. 

4dahalibut:

## Summary
It was breaking because of new changes having to do with target label casting, and because of my lineterminator issues. I fixed those things

## Testing
I tested manually, because we have a variety test that will hopefully come to fruition soon and will test this.


yakov-g (approved):

yakov-g:
please rebase on master to pass functional tests

custom_model_runner/datarobot_drum/resource/predict_mixin.py
```diff
@@ -31,14 +31,15 @@ class PredictMixin:
...
         response_status = HTTP_200_OK
-        response = None
 
         file_key = "X"
         filestorage = request.files.get(file_key)
 
         if not filestorage:
             wrong_key_error_message = (
-                "Samples should be provided as a csv file under `{}` key.".format(file_key)
+                "Samples should be provided as a csv, mtx, or arrow file under `{}` key.".format(
```
rvorobii, line 40: this is tricky, as not all predictors support Arrow right now (only python ones)
same for mtx
the only common format is csv. other formats can checked via /capabilities/ endpoint.
I think we should say that

rvorobii (approved):
LGTM

regarding error message - saying that X should be csv or mtx or arrow is not correct, because Arrow and mtx is not supported by all predictors.
probably we should say something like
"check /capabilities/ endpoint"
"should be csv or format listed in /capabilities/"

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**6.92** USD.

PR #224, opened 2020-11-16
[RAPTOR-3977] Let drum run transformations without downstream model

jlpolit:

## Summary

First step in getting custom transformers working in DRUM. Bypasses everything after the transform hook in predict method, so we just return the transformed dataset. For now, we are requiring the user to provide their own transform hook, but will hopefully relax that requirement in the future. 

This implementation will require the user to specify either a target or the --unsupervised flag; this is by design, as we still need a way to differentiate X from y for the fit method.

## To Do

I think it may be a little tricky to package both an X and y transformer into one model object (I'm guessing we can accomplish this by inheriting from the sklearn base estimator class), so I punted on putting that into the template for now.

This just makes drum fit happy, but more thought needs to go into what actually happens to the dataset once it's transformed (especially on the DR side)

The way drum is currently set up, transform only acts on the features (as far as I can tell), and not the target....we'll want a way to allow users to transform both if they so desire, which will take a bit more refactoring


yakov-g:
please rebase on master to pass functional tests

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -117,6 +117,9 @@ def load_model_from_artifact(self):
         if (
             self._target_type != TargetType.UNSTRUCTURED
             and not self._custom_hooks[CustomHooks.SCORE]
+        ) or (
+            self._target_type == TargetType.TRANSFORM
+            and not self._custom_hooks[CustomHooks.TRANSFORM]
```
yakov-g, line 122: So if target type is transform, but there is no hook, we expect artifact to do the transformation, right?
Then we should validate output, am I right?

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -320,12 +334,18 @@ def predict(self, input_filename, model=None, **kwargs):
...
                 data = self._custom_hooks[CustomHooks.TRANSFORM](data, model)
+
             except Exception as exc:
                 raise type(exc)(
                     "Model transform hook failed to transform dataset: {}".format(exc)
                 ).with_traceback(sys.exc_info()[2]) from None
             self._validate_data(data, CustomHooks.TRANSFORM)
+            if self._target_type == TargetType.TRANSFORM:
+                self._validate_transform_rows(nrow_original, data)
+                return data
```
yakov-g, line 348: I don't like this early return here

custom_model_runner/datarobot_drum/drum/model_adapter.py
jlpolit: actually now that I think of it, for now this probably isn't appropriate- down the road, we don't want to require that hook for all cases, but as of now it is necessary for transform type. will change to error for now if no hook is provided

custom_model_runner/datarobot_drum/drum/model_adapter.py
jlpolit: I'm open to that- just bypass predict entirely if it's a transform target type...let me take a look

custom_model_runner/datarobot_drum/drum/language_predictors/python_predictor/python_predictor.py
```diff
@@ -58,6 +58,17 @@ def predict(self, input_filename):
 
         return predictions
 
+    def transform(self, input_filename):
+        start_predict = time.time()
+        transformed_df = self._model_adapter.transform(input_filename, model=self._model)
+        end_predict = time.time()
+        execution_time_ms = (end_predict - start_predict) * 1000
+
+        # TODO: call monitor only if we are in structured mode
```
4dahalibut, line 67: needs ticket

jlpolit: (@yakov-g or someone? this todo was already there when I got here....)

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -219,6 +226,7 @@ def _load_via_predictors(self, model_artifact_file):
         return model
 
     def _find_predictor_to_use(self):
+        # TODO: update for transform?
```
4dahalibut, line 229: link  to the right ticket

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -241,10 +249,20 @@ def _validate_data(to_validate, hook):
...
+        if to_validate.shape[0] != nrow_original:
+            raise ValueError(
+                "Transformation resulted in different number of rows than original data"
+            )
+
     def _validate_predictions(self, to_validate, class_labels):
         self._validate_data(to_validate, "Predictions")
         columns_to_validate = set(str(label) for label in to_validate.columns)
-        if class_labels:
+        if self._target_type == TargetType.TRANSFORM:
```
4dahalibut, line 262: take it out or fail

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -320,12 +332,56 @@ def predict(self, input_filename, model=None, **kwargs):
         if self._custom_hooks.get(CustomHooks.TRANSFORM):
             try:
                 # noinspection PyCallingNonCallable
+                if self._target_type == TargetType.TRANSFORM:
+                    nrow_original = data.shape[0]
                 data = self._custom_hooks[CustomHooks.TRANSFORM](data, model)
```
4dahalibut, line 337: rename output_data from data

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -320,12 +332,56 @@ def predict(self, input_filename, model=None, **kwargs):
...
+        ----------
+        input_filename: str
+            Path to the feature csv
+        model: Any
+            The model
+        Returns
+        -------
+        pd.DataFrame
+        """
+        return self._load_and_transform_data(input_filename, model)
```
4dahalibut, line 363: rename inner method to transform

model_templates/training/python3_sklearn_transform/custom.py
```diff
@@ -0,0 +1,45 @@
...
+def fit(
+    X: pd.DataFrame,
+    y: pd.Series,
+    output_dir: str,
+    **kwargs,
+):
+    """
+    This hook must be implemented with your fitting code, for running drum in the fit mode.
+
+    This hook MUST ALWAYS be implemented for custom training models.
```
4dahalibut, line 16: and transforms

model_templates/training/python3_sklearn_transform/custom.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    transformer = make_pipeline()
+    transformer.fit(X, y)
+
+    # You must serialize out your model to the output_dir given, however if you wish to change this
+    # code, you will probably have to add a load_model method to read the serialized model back in
+    # When prediction is done.
+    # Check out this doc for more information on serialization https://github.com/datarobot/custom-\
+    # model-templates/tree/master/custom_model_runner#python
+    # NOTE: We currently set a 10GB limit to the size of the serialized model
```
4dahalibut, line 39: change language here

model_templates/training/python3_sklearn_transform/custom.py
```diff
@@ -0,0 +1,45 @@
...
+    with open("{}/artifact.pkl".format(output_dir), "wb") as fp:
+        pickle.dump(transformer, fp)
+
+
+def transform(data, model):
```
4dahalibut, line 44: you're not transforming a model

please leave docstring

tests/drum/test_fit.py
```diff
@@ -175,6 +179,8 @@ def test_fit(
             target_type = BINARY
         elif problem == MULTICLASS_NUM_LABELS:
             target_type = MULTICLASS
+        elif framework == SKLEARN_TRANSFORM:
+            target_type = TRANSFORM
         else:
             target_type = problem
```
4dahalibut, line 186: in a separate PR, add a transform server test.

custom_model_runner/datarobot_drum/drum/language_predictors/python_predictor/python_predictor.py
yakov-g: Please remove TODO; we have separate methods for predict and predict_unstructured

yakov-g: So, I think monitoring is not needed here.
MLOps monitoring is used to report predictions metrics, so we should report predictions only from predict().
Please remove self.monitor() call from here.
Also please remove TODO from the predict() call

jlpolit: sounds good

tests/drum/test_fit.py
yakov-g: In the server mode:
- do we use `predict` endpoint or we want to add `transform`?
- when transformer is used as a step in blueprint - how steps communicate to each other?

jlpolit: the plan will probably be to stand up a transform endpoint (I'm looking into that right now as a follow-up). I need to do a little more reading up on customizable blueprints, but my understanding is that there's a convention of a schema for all blueprint components, so a given step has expected input and output type that are validated. In terms of the actual mechanics of how they communicate- let me chat w/ some folks and get back to you

custom_model_runner/datarobot_drum/drum/language_predictors/python_predictor/python_predictor.py
liororama: I agree.

tests/drum/test_fit.py
jlpolit: ...so it sounds like once the transformer is being used as a step in a blueprint, it's basically going to function like any other DR task in terms of how it communicates with up/downstream tasks (at least for now).

yakov-g: and what does it mean: `function like any other DR task ` ? how does it communicate?

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -219,6 +226,7 @@ def _load_via_predictors(self, model_artifact_file):
         return model
 
     def _find_predictor_to_use(self):
+        # TODO: RAPTOR-4014
```
4dahalibut, line 229: this comment could be fleshed out a little lol

model_templates/training/python3_sklearn_transform/custom.py
```diff
@@ -0,0 +1,57 @@
...
+def fit(
+    X: pd.DataFrame,
+    y: pd.Series,
+    output_dir: str,
+    **kwargs,
+):
+    """
+    This hook must be implemented with your fitting code, for running drum in the fit mode.
+
+    This hook MUST ALWAYS be implemented for custom training models. For custom transform models, the
```
4dahalibut, line 16: custom transform models... interesting wording.

4dahalibut (approved):
:ship:

model_templates/training/python3_sklearn_transform/custom.py
jlpolit: yeah ok, maybe "custom trainsformers"

custom_model_runner/datarobot_drum/drum/model_adapter.py
jlpolit: never

yakov-g:
Need to add a couple of things:
1. Update README with transform feature
2. Update custom_model_runner/CHANGELOG:  add `1.4.4 - in progress` with proper section, t.e. Added
3. In custom_model_runner/datarobot_drum/drum/description.py set version to 1.4.4

yakov-g (approved):
Please fix doc, changelog and version

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**17.99** USD.

PR #228, opened 2020-11-17
[YOLO] Fix drum push

4dahalibut:

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/push.py
+++ b/custom_model_runner/datarobot_drum/drum/push.py
@@ -29,7 +29,7 @@ def _convert_target_type(unconverted_target_type):
     elif unconverted_target_type == TargetType.BINARY.value:
         return dr_client.TARGET_TYPE.BINARY
     elif unconverted_target_type == TargetType.ANOMALY.value:
-        return dr_client.CUSTOM_MODEL_TARGET_TYPE.ANOMALY
+        return dr_client.enums.CUSTOM_MODEL_TARGET_TYPE.ANOMALY
     raise DrumCommonException("Unsupported target type {}".format(unconverted_target_type))
 
 
@@ -149,7 +149,7 @@ def _setup_training_validation(config, options):
     options.negative_class_label = None
     options.positive_class_label = None
     options.target_csv = None
-    options.target = config["validation"]["targetName"]
+    options.target = config["validation"].get("targetName")
     options.unsupervised = None
     options.row_weights = None
     options.row_weights_csv = None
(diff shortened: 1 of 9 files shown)


4dahalibut:
jarvis test failed

4dahalibut:

custom_model_runner/datarobot_drum/drum/push.py
```diff
@@ -29,7 +29,7 @@ def _convert_target_type(unconverted_target_type):
     elif unconverted_target_type == TargetType.BINARY.value:
         return dr_client.TARGET_TYPE.BINARY
     elif unconverted_target_type == TargetType.ANOMALY.value:
-        return dr_client.CUSTOM_MODEL_TARGET_TYPE.ANOMALY
+        return dr_client.enums.CUSTOM_MODEL_TARGET_TYPE.ANOMALY
```
yakov-g, line 32: Could we unify enums being used in this block?
`dr_client.TARGET_TYPE` vs `dr_client.enums.CUSTOM_MODEL_TARGET_TYPE`

yakov-g (changes requested):

custom_model_runner/datarobot_drum/drum/push.py
4dahalibut: @jlpolit there was a reason why no. Anomaly doesn't show up in the normal modeling target types.

yakov-g (approved):

custom_model_runner/datarobot_drum/drum/push.py
jlpolit: yeah I think for now we need to check the custom model enums

custom_model_runner/datarobot_drum/drum/push.py
```diff
@@ -149,7 +149,7 @@ def _setup_training_validation(config, options):
     options.negative_class_label = None
     options.positive_class_label = None
     options.target_csv = None
-    options.target = config["validation"]["targetName"]
+    options.target = config["validation"].get("targetName")
```
jlpolit, line 152: what happens to the `raw_args_for_docker` command when target here is None/empty?

jlpolit: ...I guess it must work, since there seems to be an anomaly test case below

tests/drum/utils.py
```diff
@@ -64,11 +64,10 @@ def _exec_shell_cmd(cmd, err_msg, assert_if_fail=True, process_obj_holder=None,
         process_obj_holder.out_stream = stdout
         process_obj_holder.err_stream = stderr
 
-    if p.returncode != 0:
-        print("stdout: {}".format(stdout))
-        print("stderr: {}".format(stderr))
-        if assert_if_fail:
-            assert p.returncode == 0, err_msg
+    print("stdout: {}".format(stdout))
```
jlpolit, line 67: :+1:  :+1:

jlpolit (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**10.65** USD.

PR #231, opened 2020-11-21
[RAPTOR-4030] add support for receiving regular binary data for /predict endpoint

yakov-g:

## Summary
* Add endpoints: `predictions`/`predictionsUnstructured` - to match PPS endpoints
* Add `ping endpoint` - to match PPS endpoints
* Allow to pass data not only as form-data param, but just as binary data; when sending mtx/arrow as binary data - mimetype should be set to `text/mtx`/`text/arrow`.
* when data is passed as binary data, we won't have saved file, so `read_input_data` receives data as bytes.
* update tests to send data in all formats to all endpoints

## Rationale
* Most of this is required for PPS.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/common.py
+++ b/custom_model_runner/datarobot_drum/drum/common.py
@@ -84,7 +84,7 @@ class PredictionServerMimetypes:
     APPLICATION_JSON = "application/json"
     APPLICATION_OCTET_STREAM = "application/octet-stream"
     TEXT_PLAIN = "text/plain"
-    TEXT_ARROW = "text/arrow"
+    APPLICATION_X_APACHE_ARROW_STREAM = "application/x-apache-arrow-stream"
     TEXT_MTX = "text/mtx"
 
 
@@ -95,7 +95,7 @@ class InputFormatExtension:
 
 InputFormatToMimetype = {
     InputFormatExtension.MTX: PredictionServerMimetypes.TEXT_MTX,
-    InputFormatExtension.ARROW: PredictionServerMimetypes.TEXT_ARROW,
+    InputFormatExtension.ARROW: PredictionServerMimetypes.APPLICATION_X_APACHE_ARROW_STREAM,
 }
 
 
--- a/custom_model_runner/datarobot_drum/drum/common.py
+++ b/custom_model_runner/datarobot_drum/drum/common.py
@@ -73,10 +73,30 @@ class UnstructuredDtoKeys:
     CHARSET = "charset"
 
 
+class StructuredDtoKeys:
+    FILENAME = "filename"
+    BINARY_DATA = "binary_data"
+    MIMETYPE = "mimetype"
+    CHARSET = "charset"
+
+
 class PredictionServerMimetypes:
     APPLICATION_JSON = "application/json"
     APPLICATION_OCTET_STREAM = "application/octet-stream"
     TEXT_PLAIN = "text/plain"
+    TEXT_ARROW = "text/arrow"
+    TEXT_MTX = "text/mtx"
+
+
+class InputFormatExtension:
+    MTX = ".mtx"
+    ARROW = ".arrow"
+
+
+InputFormatToMimetype = {
+    InputFormatExtension.MTX: PredictionServerMimetypes.TEXT_MTX,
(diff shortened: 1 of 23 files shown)


yakov-g:

yakov-g:

yakov-g:

yakov-g:

yakov-g:

yakov-g:

yakov-g:

yakov-g:

yakov-g:

yakov-g:

yakov-g:

custom_model_runner/datarobot_drum/drum/common.py
```diff
@@ -73,10 +73,30 @@ class UnstructuredDtoKeys:
...
+    MIMETYPE = "mimetype"
+    CHARSET = "charset"
+
+
 class PredictionServerMimetypes:
     APPLICATION_JSON = "application/json"
     APPLICATION_OCTET_STREAM = "application/octet-stream"
     TEXT_PLAIN = "text/plain"
+    TEXT_ARROW = "text/arrow"
+    TEXT_MTX = "text/mtx"
```
meltzg, line 88: aren't these binary formats? would application/[mtx, arrow] make more sense? text is intended for human readable formats

yakov-g: Agree.
We talked with @rvorobii and @klichukb and decided that as long as there is not official mimetypes for these formats we'll stick with something.
Let's wait for others to comment and fix it.

meltzg (approved):

custom_model_runner/datarobot_drum/drum/common.py
meltzg: no need. if it's been discussed and agreed to, i don't feel strongly

rvorobii: this is definitely not "text".

I propose to change it to "application/x-apache-arrow-stream".
- "application" because it's a binary format
- "x-" means it's not standardized (check https://en.wikipedia.org/wiki/Media_type)
- "-stream" because it's a stream in Arrow terms

check out this thread where people are discussing mime types for Arrow https://lists.apache.org/thread.html/b15726d0c0da2223ba1b45a226ef86263f688b20532a30535cd5e267%40%3Cdev.arrow.apache.org%3E

rvorobii (changes requested):
let's change mime types
binary formats should be "application/..."

also, at least for Arrow IMO better option is `application/x-apache-arrow-stream` (see my comment for details)

custom_model_runner/datarobot_drum/drum/common.py
klichukb: This is _not_ a text format. When we discussed sticking to some I mean one of those discussed in arrow threads. ^ What Roman says.

klichukb: Also mtx is probably not text as well, these ought to be fixed before merging.

klichukb: Mtx seems to be a text format, maybe that one is fine with 'text/' unless you know better.

custom_model_runner/datarobot_drum/drum/common.py
```diff
@@ -73,10 +73,30 @@ class UnstructuredDtoKeys:
...
+    MIMETYPE = "mimetype"
+    CHARSET = "charset"
+
+
 class PredictionServerMimetypes:
     APPLICATION_JSON = "application/json"
     APPLICATION_OCTET_STREAM = "application/octet-stream"
     TEXT_PLAIN = "text/plain"
+    APPLICATION_X_APACHE_ARROW_STREAM = "application/x-apache-arrow-stream"
+    TEXT_MTX = "text/mtx"
```
rvorobii, line 88: mtx is not standardized
should we have it "text/x-mtx"?

rvorobii (approved):
not sure if we need to have mtx mimetype as "text/x-mtx", since mtx is not standardized

also fyi there'll be conflicts with PR https://github.com/datarobot/datarobot-user-models/pull/242 (either me or you will have to resolve conflicts)

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**6.9** USD.

PR #232, opened 2020-11-23
[RAPTOR-3979][RAPTOR-4017] Transform endpoint for prediction server

jlpolit:

## Summary

Set up Transform endpoint in prediction server. Also changed transform pipeline to omit custom SVD method because that was causing issues.


custom_model_runner/datarobot_drum/resource/predict_mixin.py
```diff
@@ -115,3 +116,53 @@ def _validate_content_type_header(header):
...
+            if logger is not None:
+                logger.debug("Filename provided under X key: {}".format(filestorage.filename))
+
+        _, file_ext = os.path.splitext(filestorage.filename)
+        with tempfile.NamedTemporaryFile(suffix=file_ext) as f:
+            filestorage.save(f)
+            f.flush()
+            out_data = self._predictor.transform(f.name)
+
+        # float32 is not JSON serializable, so cast to float, which is float64
```
zohar-mizrahi, line 155: It would be better, IMHO, to use the `default_handler` argument of the `to_json` function?

jlpolit: I'm reading the description of that arg but still not totally clear on how to use it- got an example handy? thanks!

custom_model_runner/datarobot_drum/resource/predict_mixin.py
```diff
@@ -29,8 +30,19 @@ class PredictMixin:
 
     """
 
-    def do_predict(self, logger=None):
+    def _predict_or_transform(self, logger=None, predict=True):
         response_status = HTTP_200_OK
+        if predict is False and self._target_type != TargetType.TRANSFORM.value:
```
yakov-g, line 35: As I understand, we should also check here for the case:
predict==True and target_type == TRANSFORM

Instead of this we could check for proper target_type in `do_transform` and `do_predict` methods and then use `_predict_or_transform` without additional param `predict`. And do resolution on target_type as we already checked it.

Or you just don't need `do_transform` and `do_predict` and can use `_predict_or_transform` directly from the server code.

yakov-g (changes requested):

custom_model_runner/datarobot_drum/resource/predict_mixin.py
```diff
@@ -29,8 +30,19 @@ class PredictMixin:
...
+            wrong_target_type_error_message = (
```
yakov-g, line 36: As I understand, this error message is in case wrong endpoint is used.
Please change the message to contain "endpoint" terminology, e.t. "Your project is transform, use /transform/ endpoint"
"Your project is binary, use /predict/ endpoint"
just handle the case if target_type is unstructured - use 'predictUnstructured' endpoint

custom_model_runner/datarobot_drum/resource/components/Python/prediction_server/prediction_server.py
```diff
@@ -91,6 +91,20 @@ def predict():
                 self._stats_collector.disable()
             return response, response_status
 
+        @model_api.route("/transform/", methods=["POST"])
```
yakov-g, line 94: You also need to add new endpoint into `uwsgi_component`

yakov-g (changes requested):

custom_model_runner/datarobot_drum/resource/predict_mixin.py
jlpolit: good call!

custom_model_runner/datarobot_drum/resource/components/Python/prediction_server/prediction_server.py
jlpolit: thanks! what's the right place to add test cases for that one?

yakov-g: test_custom_models_with_drum_nginx_prediction_server
notice, these tests run inside docker container to make sure we install nginx properly.  And this container only has deps to run sklearn.
Ping me in slack if need help.

jlpolit: thanks - I think I got it up, but having trouble running the test locally (presumably for docker image reasons)

custom_model_runner/datarobot_drum/resource/predict_mixin.py
```diff
@@ -53,10 +54,21 @@ def do_predict(self, logger=None):
...
             filestorage.save(f)
             f.flush()
-            out_data = self._predictor.predict(f.name)
+            if self._target_type == TargetType.TRANSFORM.value:
+                out_data = self._predictor.transform(f.name)
+            else:
+                out_data = self._predictor.predict(f.name)
 
-        if self._target_type == TargetType.UNSTRUCTURED:
+        if self._target_type == TargetType.UNSTRUCTURED.value:
```
yakov-g, line 62: TargetType.UNSTRUCTURED should not take value.
self._target_type is enum, not str, check it's assignment in `BaseLanguagePredictor`

And the same for TRANSFORM further in the code

yakov-g (changes requested):

custom_model_runner/datarobot_drum/resource/predict_mixin.py
jlpolit: I must be doing something wrong in my test then- I get a mismatch when I don't check the .value for some reason

yakov-g:
LGTM, just need to update README with the link to the training transform template.
I'm appriving, you can add update the readme in the next PR you have

yakov-g (approved):

jlpolit:
closing in favor of https://github.com/datarobot/datarobot-user-models/pull/236

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**10.36** USD.

PR #233, opened 2020-11-24
[RAPTOR-4033] pin pyarrow version only for DR envs

yakov-g:

## Summary
Currently DRUM is pinned to pyarrow==0.14.1 which is the version supported by DR app, but available only for python 3.7. Which makes it impossible to install DRUM locally in other Py versions.

Solution: unpin drum pyarrow version, pin public envs pyarrow version


rvorobii (approved):

4dahalibut (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**3.77** USD.

PR #236, opened 2020-11-30
[RAPTOR-3979][RAPTOR-4016][RAPTOR-4017] Transform endpoint using arrow/mtx

jlpolit:

## Summary

Includes everything from here (https://github.com/datarobot/datarobot-user-models/pull/232) but pass output as arrow/mtx rather than json.


tests/fixtures/transform_custom_dense.py
```diff
@@ -0,0 +1,14 @@
+import pandas as pd
```
4dahalibut, line 1: these files seem the same

jlpolit: oh crap - you're right...I was converting to a sparse dataframe and then switched back and forgot it looked the same as the other one

meltzg (approved):

custom_model_runner/datarobot_drum/resource/predict_mixin.py
```diff
@@ -53,10 +58,20 @@ def do_predict(self, logger=None):
...
+                out_data = self._predictor.transform(f.name)
+            else:
+                out_data = self._predictor.predict(f.name)
 
         if self._target_type == TargetType.UNSTRUCTURED:
             response = out_data
+        elif self._target_type == TargetType.TRANSFORM:
+            if is_sparse(out_data):
+                mtx_payload = make_mtx_payload(out_data)
+                response = '{{"X.transformed":{mtx_payload}}}'.format(mtx_payload=mtx_payload)
```
yakov-g, line 71: You defined the key for X.transformed, but forgot to use it :)
We need to move it to the common.py, can be done later.

yakov-g (approved):

custom_model_runner/datarobot_drum/resource/predict_mixin.py
jlpolit: :facepalm:

4dahalibut, line 71: How are we going to know in DR if the payload is arrow or mtx?

custom_model_runner/datarobot_drum/resource/predict_mixin.py
```diff
@@ -76,6 +91,19 @@ def do_predict(self, logger=None):
 
         return response, response_status
 
+    def do_predict(self, logger=None):
+        if self._target_type == TargetType.TRANSFORM:
+            wrong_target_type_error_message = (
+                "This project has target type {}, "
+                "use the /transform/ endpoint.".format(self._target_type)
+            )
+            if logger is not None:
```
4dahalibut, line 100: why would logger be none?

custom_model_runner/datarobot_drum/resource/transform_helpers.py
```diff
@@ -0,0 +1,40 @@
...
+from scipy.io import mmwrite, mmread
+from scipy.sparse.csr import csr_matrix
+from scipy.sparse import vstack
+
+
+X_TRANSFORM_KEY = "X.transformed"
+
+
+def is_sparse(df):
+    return hasattr(df, "sparse") or type(df.iloc[0].values[0]) == csr_matrix
```
4dahalibut, line 14: why are there two conditions here?

custom_model_runner/datarobot_drum/resource/transform_helpers.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def make_arrow_payload(df):
+    return pa.ipc.serialize_pandas(df, preserve_index=False).to_pybytes()
```
4dahalibut, line 18: have you tried deserializing this with python2.7?

custom_model_runner/datarobot_drum/resource/transform_helpers.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def read_arrow_payload(response_dict):
+    bytes = response_dict[X_TRANSFORM_KEY]
+    df = pa.ipc.deserialize_pandas(bytes)
```
4dahalibut, line 23: same - have you tried writing this with python 2.7?

custom_model_runner/datarobot_drum/resource/predict_mixin.py
jlpolit: not sure- I was following an existing pattern

custom_model_runner/datarobot_drum/resource/predict_mixin.py
jlpolit: good point - how do you feel about changing the key name to have a sparse/dense indicator? something like X.transformed.sparse, X.transformed.dense?

tests/drum/test_inference.py
```diff
@@ -236,6 +241,100 @@ def test_custom_models_with_drum_nginx_prediction_server(
...
+            response = requests.post(
+                run.url_server_address + "/transform/", files={"X": open(input_dataset)}
+            )
+            print(response.text)
+            assert response.ok
+
+            in_data = pd.read_csv(input_dataset)
+
+            # transformed_mat = (json.loads(response.text))["transformations"]
+            # actual_num_predictions = len(transformed_mat)
```
4dahalibut, line 333: commented out code

custom_model_runner/datarobot_drum/resource/transform_helpers.py
jlpolit: the first one is the case where they're actually using a pandas sparse datatype, the second is the case where they've just dumped a csr matrix into a pandas DF. we could also just enforce one pattern or another

tests/drum/test_inference.py
```diff
@@ -236,6 +241,100 @@ def test_custom_models_with_drum_nginx_prediction_server(
...
+
+            in_data = pd.read_csv(input_dataset)
+
+            if framework == SKLEARN_TRANSFORM_DENSE:
+                transformed_df = read_arrow_payload(eval(response.text))
+                actual_num_predictions = transformed_df.shape[0]
+            else:
+                transformed_mat = read_mtx_payload(eval(response.text))
+                actual_num_predictions = transformed_mat.shape[0]
+            assert in_data.shape[0] == actual_num_predictions
```
4dahalibut, line 290: would you mind fleshing this out a bit?

custom_model_runner/datarobot_drum/resource/predict_mixin.py
4dahalibut: I'm not sure - is the response json? can we put something else in the json?

custom_model_runner/datarobot_drum/resource/transform_helpers.py
4dahalibut: ok but we're deserializing the data inside of drum and this is in drum so its us using a pandas sparse df

jlpolit: but it depends on what the user put in their transform hook, since that's the data structure we're serializing

custom_model_runner/datarobot_drum/resource/predict_mixin.py
jlpolit: yeah we could add another key to indicate sparse vs dense

custom_model_runner/datarobot_drum/resource/transform_helpers.py
jlpolit: good call - will see how that works...presumably we particularly need to worry about the case where we're serializing in python3 and deserializing in python 2?

custom_model_runner/datarobot_drum/resource/predict_mixin.py
4dahalibut: :+1:

custom_model_runner/datarobot_drum/resource/predict_mixin.py
4dahalibut: ok I guess we don't have to do this now but still weird.

custom_model_runner/datarobot_drum/resource/transform_helpers.py
4dahalibut: lets just land on one of them - unless there's a difference between python and R

custom_model_runner/datarobot_drum/resource/transform_helpers.py
rvorobii: @jlpolit for Arrow the version is payload
prior to Arrow v1.0.0 there're no backward/forward compatibility guarantees, so Arrow for instance if you serialized data with Arrow 0.24.1, you have to read it with 0.24.1

In the use cases where I added Arrow I check what Arrow version is installed in DRUM, and use Arrow to communicate between DRUM and DataRobot only if DRUM Arrow version works with DataRobot Arrow version (for Arrow < 1.0.0 that means exactly the same version)

jlpolit: that tracks with what I'm seeing when trying to deserialize between drum and DR right now- what's your fallback behavior if the versions don't match up? just write_csv to bytesIO? sounds like I'll have to add that arrow version check here

rvorobii: yes, the fallback is to use csv
I do this in 2 steps:
1. Check what Arrow version is on DRUM side (check DRUM's `/capabilities/` endpoint)
2. Decide if Arrow can be used. If not (versions are not compatible) - use csv

yes, you have to check that Arrow versions are the same before using Arrow

jlpolit: so the use case you're talking about is whether you're going to pass an arrow object from DR to drum, right? which means if I want to determine whether to pass an arrow object the other direction I will need to either do that same check on the DR side and then pass, say, a `use_arrow` flag to the endpoint to tell drum what to do, or I'll have to pass my DR arrow version in the payload so drum can decide for itself

rvorobii: > so the use case you're talking about is whether you're going to pass an arrow object from DR to drum, right?

yes!
this is prediction requests' data passed from DR to DRUM

> which means if I want to determine whether to pass an arrow object the other direction I will need to either do that same check on the DR side and then pass, say, a use_arrow flag to the endpoint to tell drum what to do, or I'll have to pass my DR arrow version in the payload so drum can decide for itself

yes
regarding your options, I like first one a little bit more
In DR, custom model capabilities (which include Arrow version) are already set to `CustomModelExecutable`. 
Then before doing predictions to CM model I check the capabilities and decide if I can use Arrow.
for your case, it may be easy to also check the capabilities and pass this `use_arrow` flag.
In this case we'll want have logic that checks Arrow versions compatibility duplicated in DR and DRUM.
Also note that the capabilities are not set to `CustomModelExecutable` for all execution paths yet (TBD in future). For instance, the capabilities aren't set to `CustomModelExecutable` for variable importance job.

jlpolit: hey @rvorobii do you know if there's a way to tell pyarrow to serialize using an older version? I'm seeing some stuff in `pyarrow.ipc.IpcWriteOptions` that suggests it may be possible

rvorobii: I'm not aware if this is possible.
Haven't seen anything related in Arrow docs.

jlpolit: ok, I _think_ I have a solution

jlpolit: @4dahalibut  the current implementation will take arrow version from DR, and if DR is using an older arrow than drum it returns in an older format- I've confirmed it works going from version 0.2 to 0.1.4. lmk if that seems like a reasonable solution to you (i.e. that it's not too onerous for DR to send an arrow version to drum when it calls this endpoint)

4dahalibut: I think that's fine

rvorobii: @jlpolit also, we should be able to disable Arrow usage - it's not as battle-tested as csv is, so in case in there're any issues with Arrow we should be able to disable it. Or make it OFF by default, but let users to opt-in.

for instance, for my use case where DR can use Arrow to make predictions to DRUM, it's possible to disable Arrow via EngConfig, so that CSV is always used (this is especially important for On Prem releases).

4dahalibut: Roman - since our feature is only barely in alpha, its much more okay if it breaks.

rvorobii: you guys know better

jlpolit: as it stands right now, if you don't pass an arrow version to the endpoint it will still default to csv

4dahalibut: I think it should probably just fail but we can do it this way for now

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**6.15** USD.

PR #238, opened 2020-12-02
[RAPTOR-3784] compare host and container drum versions - drop 2

yakov-g:

## Summary
Noticed, that comparison always fails because version str is `drum 1.4.5`


meltzg (approved):

yakov-g:

rvorobii (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**1.72** USD.

PR #239, opened 2020-12-03
[WIP] pass X and y to transform

jlpolit:


tests/fixtures/transform_custom.py
```diff
@@ -2,18 +2,20 @@
 from scipy.sparse.csr import csr_matrix
 
 
-def transform(data, transformer):
+def transform(X, y, transformer):
```
jlpolit, line 5: @4dahalibut  what are your thoughts on the best convention for the transform hook? I don't think we want to change the general behavior of transform for non-transform jobs, but I also don't love having the hook having a different set of inputs/outputs in transform mode vs other target types (thought about putting y=None as a third arg instead of this placement, not sure if that's any better).

PR #241, opened 2020-12-03
[RAPTOR-3660] Add prediction validation to drum fit

jlpolit:


custom_model_runner/datarobot_drum/resource/drum_server_helpers.py
```diff
@@ -0,0 +1,203 @@
+import os
```
jlpolit, line 1: obvoiusly redeclaring a bunch of functionality from tests inside of drum is kind of silly- any reason not to just move these into drum and call from that location in tests? alternatively, maybe there's a better way to call drum server from within a drum fit run and I should be taking a different approach altogether?

jlpolit: @yakov-g @4dahalibut  is there a preferred way to call drum server from `run_test_predict`? if these helpers can live in drum I'll just move them over fully, but very open to other approaches

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -353,8 +358,65 @@ def run_test_predict(self):
...
             df.to_csv(__tempfile.name, index=False)
             self.options.input = __tempfile.name
+        if self.target_type != TargetType.TRANSFORM:
+            self._check_prediction_side_effects()
         self._run_fit_and_predictions_pipelines_in_mlpiper()
 
+    def _check_prediction_side_effects(self):
+        rtol = 2e-02
+        atol = 1e-06
+        if self.options.input.endswith(".mtx"):
```
jlpolit, line 368: will it always be clear from file prefix whether the input data is sparse?

custom_model_runner/datarobot_drum/drum/common.py
```diff
@@ -16,6 +16,7 @@
 CLASS_LABELS_ARG_KEYWORD = "class_labels"
 TARGET_TYPE_ARG_KEYWORD = "target_type"
 X_TRANSFORM_KEY = "X.transformed"
+RESPONSE_PREDICTIONS_KEY = "predictions"
```
yakov-g, line 19: We have "REGRESSION_PRED_COLUMN" key defined upper. Could we use it?

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**6.83** USD.

PR #242, opened 2020-12-03
[RAPTOR-4154, RAPTOR-4158] Handle NA values in Arrow consistently to CSV

rvorobii:

## Summary
`None` and `numpy.nan` values after serializing to CSV and deserializing become `numpy.nan`.
`None` and `numpy.nan` values after serializing to Arrow and deserializing become `numpy.nan` for numerical columns and `None` for object columns.

Since we support both CSV and Arrow, NA values handling should be consistent.
(+ some existing models don't expect `None`, because with CSV it's always `numpy.nan`).


yakov-g (approved):

meltzg (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**3.28** USD.

PR #243, opened 2020-12-04
[RAPTOR-4163]Fix for the label data for multiclass

scottp-ml:

## Summary
Fix the input class label data used for validation to be a string during comparison.  The incoming data from the model is cast to a string, however this input to the validation was not.


4dahalibut (approved):

yakov-g (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**4.23** USD.

PR #245, opened 2020-12-05
[RAPTOR-3660] pred consistency check

jlpolit:

## Summary

We've had situations where a CFDS ran a model locally in drum fit and got the all clear to upload to DR, but once running in DR encountered prediction consistency errors due to instability in the preprocessing pipeline. This is due to our prediction checks in drum fit not making predictions in the same way as we do in DR (i.e. not using the prediction server). 

This PR introduces a similar prediction consistency check to the DR app, calling the predict endpoint locally so we can check for any randomness in predictions.


custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/score.R
```diff
@@ -2,6 +2,7 @@
 library(caret)
 library(devtools)
 library(stringi)
+library(Matrix)
```
jlpolit, line 5: this was a bug waiting to happen

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -353,8 +359,68 @@ def run_test_predict(self):
...
+
+        if self.target_type == TargetType.BINARY:
+            labels = [self.options.negative_class_label, self.options.positive_class_label]
+        elif self.target_type == TargetType.MULTICLASS:
+            labels = self.options.class_labels
+        else:
+            labels = None
+
+        with DrumServerRun(
+            self.target_type.value,
```
jlpolit, line 395: for some reason the drum push tests are failing on jarvis here but not locally - the traceback seemingly corresponds to code from a previous commit? I dunno

custom_model_runner/datarobot_drum/resource/drum_server_utils.py
```diff
@@ -7,7 +7,7 @@
```
jlpolit, line 7: someone definitely chime in if I shouldn't have moved this and the utils file directly below into drum resources

custom_model_runner/datarobot_drum/drum/common.py
```diff
@@ -16,6 +16,7 @@
 CLASS_LABELS_ARG_KEYWORD = "class_labels"
 TARGET_TYPE_ARG_KEYWORD = "target_type"
 X_TRANSFORM_KEY = "X.transformed"
+RESPONSE_PREDICTIONS_KEY = "predictions"
```
yakov-g, line 19: We have REGRESSION_PRED_COLUMN. Do we need to have another one, not capitalized?

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -353,8 +359,68 @@ def run_test_predict(self):
                 df = handle_missing_colnames(df)
             df.to_csv(__tempfile.name, index=False)
             self.options.input = __tempfile.name
+        if self.target_type != TargetType.TRANSFORM:
+            self._check_prediction_side_effects()
         self._run_fit_and_predictions_pipelines_in_mlpiper()
 
+    def _check_prediction_side_effects(self):
```
yakov-g, line 366: Do you mind working a bit on validation infrastructure:
* https://github.com/datarobot/datarobot-user-models/blob/master/custom_model_runner/datarobot_drum/drum/perf_testing.py - from here, move `validation_test` to a separate file
* move `_check_prediction_side_effects` to the new file/class created on the prev step.
* make validation check to perform these two validations
* in current place use `prediction side effects` validation as standalone

custom_model_runner/datarobot_drum/resource/drum_server_utils.py
yakov-g: Ok.... Agreed..
Could you just add description to the classes and methods in `drum_server_utils` and `utils` files.
E.g. `DrumServerRun` helper class used in validation and tests to start server as a thread... etc.

yakov-g (changes requested):
Looks good, just add more documentation. Let's discuss if there are any questions

custom_model_runner/datarobot_drum/drum/common.py
jlpolit: it seems the key from the prediction server comes back keyed to a lower-case 'predictions'...so we presumably need both keys (although this could probably be renamed to be clearly about pred server output?) unless we want to standardize to either capital or lowercase in both situations

custom_model_runner/datarobot_drum/drum/drum.py
jlpolit: can do

jlpolit: ...would you move the entire parent class of that method? I'm not super familiar with this corner of the codebase

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -353,8 +359,68 @@ def run_test_predict(self):
                 df = handle_missing_colnames(df)
             df.to_csv(__tempfile.name, index=False)
             self.options.input = __tempfile.name
+        if self.target_type != TargetType.TRANSFORM:
```
4dahalibut, line 362: can we delete this caveat?

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -353,8 +359,68 @@ def run_test_predict(self):
...
+            df = pd.DataFrame(mmread(self.options.input).tocsr())
+        else:
+            df = pd.read_csv(self.options.input)
+        samplesize = min(1000, max(int(len(df) * 0.1), 10))
+        data_subset = df.sample(n=samplesize, random_state=42)
+
+        if is_sparse:
+            __tempfile_sample = NamedTemporaryFile(suffix=".mtx")
+            sparse_mat = vstack(x[0] for x in data_subset.values)
+            mmwrite(__tempfile_sample.name, sparse_mat)
```
4dahalibut, line 382: ????

4dahalibut: why are you writing it out again?

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -353,8 +359,68 @@ def run_test_predict(self):
...
+            self._check_prediction_side_effects()
         self._run_fit_and_predictions_pipelines_in_mlpiper()
 
+    def _check_prediction_side_effects(self):
+        rtol = 2e-02
+        atol = 1e-06
+        input_extension = os.path.splitext(self.options.input)
+        is_sparse = input_extension[1] == ".mtx"
+
+        if is_sparse:
```
4dahalibut, line 372: please refactor so you don't double the if check

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -353,8 +359,68 @@ def run_test_predict(self):
...
+            matches = np.isclose(preds_full_subset, preds_sample, rtol=rtol, atol=atol)
+            if not np.all(matches):
+                message = """
+                            Error: Your predictions were different when we tried to predict twice.
+                            No randomness is allowed.
+                            The last 10 predictions from the main predict run were: {}
+                            However when we reran predictions on the same data, we got: {}""".format(
+                    preds_full_subset[~matches][:10], preds_sample[~matches][:10]
+                )
+                raise ValueError(message)
```
4dahalibut, line 422: can you make sure that the subset used for predictions is accessible if the check fails?

tests/fixtures/custom_pred_consistency/custom.py
```diff
@@ -0,0 +1,127 @@
+import pandas as pd
+import numpy as np
```
4dahalibut, line 2: could you please tidy this? thanks.

meltzg (commented):
context in the PR description would be nice

custom_model_runner/datarobot_drum/drum/drum.py
jlpolit: ...what's a reasonable place to write this to?

4dahalibut: temp, and then print a path

custom_model_runner/datarobot_drum/drum/common.py
yakov-g: got it, it's RESPONSE, lets keep it

custom_model_runner/datarobot_drum/drum/drum.py
yakov-g: Lets start by moving `_check_prediction_side_effects` into existing `CMRunTests` class

tests/fixtures/custom_pred_consistency/custom.py
jlpolit: I added some docstrings to explain what's going on - this turns out to be a pretty good example of how to separately build a preprocessing pipeline and a model and tell drum how to load/run them where needed rather then bundling them into one pkl object

4dahalibut (approved):

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/score.R
yakov-g: Yeah, encountered it as well

yakov-g (approved):

yakov-g:

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**15.96** USD.

PR #246, opened 2020-12-05
[RAPTOR-4106] add monitoring to R and Java predictors

yakov-g:

## Summary
Add monitoring to R and Java predictors.
Tested manually.


yakov-g:

yakov-g:

rvorobii (approved):
LGTM

meltzg (approved):

yakov-g:
@rvorobii about monitoring....
I don't know....
We have `monitoring` and `push` which will work only with the APP. Should we hide these options? Or we could just mention that these are just DR related functionality....
That's like public API.  Client is open for everyone , but worth nothing without DR.

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**8.77** USD.

PR #247, opened 2020-12-07
[RAPTOR-4005] Allow transform to take and output X and y

jlpolit:


custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -467,7 +474,7 @@ def _prepare_fit_pipeline(self, run_language):
             else:
                 raise DrumCommonException(
                     "Target type {} requires class label information. No labels were supplied and "
-                    "labels could not be inferred from the target."
+                    "labels could not be inferred from the target.".format(self.target_type.value)
```
jlpolit, line 477: unrelated, but uncovered while working on this and figured we may as well tell people what target type we think they're using if they hit this error

tests/fixtures/transform_custom.py
```diff
@@ -2,18 +2,19 @@
 from scipy.sparse.csr import csr_matrix
 
 
-def transform(data, transformer):
+def transform(X, y, transformer):
```
jlpolit, line 5: I don't love having transform behave differently for custom transformers vs everything else....open to ideas about what the right way is to do this (add y=None as an optional arg after transformer? I dunno)

custom_model_runner/datarobot_drum/resource/pipelines/prediction_pipeline.json.j2
```diff
@@ -21,7 +21,8 @@
                 "monitor_settings": "{{ monitor_settings }}",
                 "target_type": "{{ target_type }}",
                 "query_params": {{ query_params }},
-                "content_type": {{ content_type }}
+                "content_type": {{ content_type }},
+                "target_csv": "{{ target_csv }}"
```
jlpolit, line 25: is this a bad idea?

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -395,6 +396,12 @@ def _prepare_prediction_server_or_batch_pipeline(self, run_language):
                     "output_filename": '"{}"'.format(options.output) if options.output else "null",
                 }
             )
+            if self.target_type == TargetType.TRANSFORM:
```
4dahalibut, line 399: can you change test_predict to use server instead of score?

tests/fixtures/transform_custom.py
jlpolit: ....update: we're going with (data, transform, y=None) for now (or maybe `(X, transform, y=None)`?)

jlpolit:
closing in favor of https://github.com/datarobot/datarobot-user-models/pull/251

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**5.48** USD.

PR #254, opened 2020-12-11
[RAPTOR-4164] fix binary num class order

scottp-ml:

## Summary
Fix a bug with numeric class labels containing 0 for binary classification not correctly creating a class order list.


meltzg (approved):
great find. we should publish once this goes in

yakov-g (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**5.53** USD.

PR #256, opened 2020-12-14
[RAPTOR-3916] make drum require Python >=3.4,<=3.8

yakov-g:

## Summary
Limit DRUM supported Py versions to >=3.4,<=3.8

Pyarrow fails to be installed in Py3.9.

To move forward with supported Py version we need to add a test that installs drum under different Python's.
Probably we can use  `tox`.
https://datarobot.atlassian.net/browse/RAPTOR-4230


meltzg (approved):

yakov-g:

rvorobii (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**3.35** USD.

PR #258, opened 2020-12-16
[RAPTOR-4203] Make Arrow logic optional, release v1.4.9

rvorobii:

## Summary
DRUM supports Arrow as a prediction request payload format.
The problem is that Arrow is not supported by Python 3.9, and Arrow installation can require compiling native binaries.

To simplify DRUM installation and extend python versions support, Arrow should be made an optional feature, as probably most users are fine without Arrow.
DRUM will not bring Arrow by default. If Arrow support is required - it'll be a responsibility of an end user to install Arrow.
If Arrow is not installed, but Arrow payload is received - error response will be returned. 

## Testing
Probably we need a separate set of jobs to test DRUM w/o Arrow installed?


custom_model_runner/datarobot_drum/drum/utils.py
```diff
@@ -230,8 +230,11 @@ def read_structured_input_data_as_df(binary_data, mimetype):
         try:
             if mimetype == PredictionServerMimetypes.TEXT_MTX:
                 return pd.DataFrame.sparse.from_spmatrix(mmread(io.BytesIO(binary_data)))
-            elif mimetype == PredictionServerMimetypes.APPLICATION_X_APACHE_ARROW_STREAM:
-                df = pyarrow.ipc.deserialize_pandas(binary_data)
+            elif (
+                mimetype == PredictionServerMimetypes.APPLICATION_X_APACHE_ARROW_STREAM
+                and get_pyarrow_module() is not None
```
yakov-g, line 235: I think we should check mush earlier if format is supported. I'll leave general comment.

yakov-g (changes requested):
What we have now:
- python predictor supports (csv/arrow/mtx)
- r predictor supports (csv/mtx) - I forgot to update supported formats with mtx
- java predictor supports csv
That's all related to the in-built data reading.
If user provides the hook `read_input_data` we don't care. (I'm talking about DRUM only)

So my idea is to check supported formats here: https://github.com/datarobot/datarobot-user-models/blob/f630111a1f87cc16066100affaa6dfd01abee396/custom_model_runner/datarobot_drum/resource/predict_mixin.py#L78
-by mime type - that incoming data supported in supported_payload_formats
(The same check should be added for drum score )

But this check is only relevant if  `read_input_data`  hook is not implemented.
If custom models has this hook, user even don't need to provide mimetype.
We can add a method that checks if `read_input_data hook` is implemented; and check incoming data against supported_payload_formats .

yakov-g (approved):

meltzg (approved):

custom_model_runner/datarobot_drum/resource/predict_mixin.py
```diff
@@ -118,6 +140,9 @@ def _transform(self, logger=None):
             feature_binary_data, feature_mimetype, feature_charset = self._fetch_data_from_request(
                 "X", logger=logger
             )
+            mimetype_support_error_response = self._check_mimetype_support(feature_mimetype)
```
rvorobii, line 143: @4dahalibut  this PR affects custom transform code. please have a look

4dahalibut: so for transforms we want pyarrow to be the only option...

rvorobii: right now it can be either Arrow or CSV, correct?
are you going to drop CSV support?

rvorobii: cc @yakov-g

4dahalibut: Yes that is what I would like to do @jlpolit

jlpolit: if everyone's on board with that, I'll make that change

rvorobii: let's hear from @yakov-g - Arrow complicates DRUM installation

yakov-g: So....with current approach, if `pyarrow` won't be installed `transform`; user will see the error `install pyarrow` - looks  ok to me.

We could also add an early check for pyarrow here:
- https://github.com/datarobot/datarobot-user-models/blob/1bff8dff4c0d21e23d2196f727a5eb68cee1259f/custom_model_runner/datarobot_drum/drum/args_parser.py#L832
- Something like this:
```
if options.target_type == "transform":
   try:
       import pyarrow
    except ImportError:
       print("transform requires arrow to be supported")
        exit(1)
```

4dahalibut: why can't we just install pyarrow

rvorobii: if installing pyarrow is not terribly bad - simly installing it would make things simpler for everyone and easier to maintain and test

yakov-g: We can require pyarrow - only issues is that some Py and arrow versions are not cross compatible.

yakov-g: Btw, even if we keep `pyarrow` in our requirements file, this PR should go in as is.

rvorobii: do you mean so that in case `pyarrow` installation fails on some python version or platform - we'll still have fully functional DRUM just without Arrow?
oh, and without custom transforms (if Arrow is the only format there)

4dahalibut: well nonfunctional combinations will probably only be found in developer setups... So I say we proceed with that for now.

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
```diff
@@ -88,8 +88,12 @@ def configure(self, params):
     def supported_payload_formats(self):
         formats = SupportedPayloadFormats()
         formats.add(PayloadFormat.CSV)
+        formats.add(PayloadFormat.MTX)
```
4dahalibut, line 91: Is this tested?

yakov-g: yes

custom_model_runner/datarobot_drum/resource/predict_mixin.py
yakov-g: > do you mean so that in case `pyarrow` installation fails on some python version or platform - we'll still have fully functional DRUM just without Arrow?
> oh, and without custom transforms (if Arrow is the only format there)

Yes. Also people could install with `--no-deps`

requirements_test.txt
```diff
@@ -1,9 +1,11 @@
 dr-usertool
 pytest
 pytest-runner
+mock
```
4dahalibut, line 4: please always pin versions

custom_model_runner/datarobot_drum/drum/language_predictors/r_predictor/r_predictor.py
yakov-g: ....or maybe no, i'll verify soon

custom_model_runner/datarobot_drum/drum/common.py
```diff
@@ -300,3 +318,19 @@ def make_predictor_capabilities(supported_payload_formats):
...
     }
+
+
+try:
+    import pyarrow
+except ImportError:
+    pyarrow = None
+
+
+def get_pyarrow_module():
```
4dahalibut, line 329: is this the best way to do this? I've never seen this pattern

rvorobii: I've never seen the case like ours where we have optional module and part of the logic depending on it

I you've got better ideas - please let us know

4dahalibut: heres what I do https://github.com/datarobot/datarobot-user-models/blob/6100db14a27461c8ed06a48032dbbc04f345206d/custom_model_runner/datarobot_drum/drum/push.py#L43

rvorobii: the idea was to have DRUM working if there's no Arrow, without forcing user to install it

4dahalibut: Although that is different I guess...

yakov-g: I've seen Zohar using this, and learn from him:
https://github.com/datarobot/mlpiper/blob/c9af874c3ece97870887b3ef8b3ddde122281b7c/mlpiper-py/mlpiper/components/restful/metric.py#L8

I mean `try: import` is just standard way.
And how you pass imported module around, should not matter

rvorobii: if user doesn't need Arrow - user won't notice it's missing

yakov-g: > heres what I do
> 
> https://github.com/datarobot/datarobot-user-models/blob/6100db14a27461c8ed06a48032dbbc04f345206d/custom_model_runner/datarobot_drum/drum/push.py#L43

But you fail if module is not imported, here we don't.

yakov-g: Here more opinions:
https://stackoverflow.com/questions/9689355/python-import-or-pass-modules-as-parameters

yakov-g: So our way look fine to me.

requirements_test.txt
rvorobii: sorry this is a leftover from previous iterations

custom_model_runner/requirements.txt
```diff
@@ -13,4 +13,4 @@ scipy>=1.1,<2
 strictyaml==1.0.6
 texttable
 py4j~=0.10.9.0
-pyarrow
\ No newline at end of file
+pyarrow
```
rvorobii, line 16: @yakov-g @4dahalibut I've made `pyarrow` required again (don't know why this diff is even shown)
probably we should not pin `pyarrow` version - we don't care about any particular version, and there're more chances it'll get installed if unpinned

so, is this PR good to go?

yakov-g: 1. I'd go without `pyarrow`, but @4dahalibut and @jlpolit should tell how many people use transform.
2. If we go with `pyarrow`, we should not pin it, as new pip with deps check would scream, if requirements are not met.
3. I still need to add a test case for R+mtx. So @4dahalibut or @jlpolit please approve, and i'll merge once test is in.

4dahalibut: Lots of people will be using transforms. I'm waiting on at least sparse R testing to continue.

yakov-g: So...
mtx case is implicitly tested in `test_fit_sparse`, when model is verified after fitting.

To add such test case into inference test, I'll need to add new R artifact, trained on sparse data. I can not use R artifact trained on csv data with sparse payload as it is missing column names.

Do you want me to add an artifact and separate test casr for R with mtx?

yakov-g: @4dahalibut ^

yakov-g: @4dahalibut added the test case for R + mtx

4dahalibut: I still think that this needs to be pinned at least in some range. having unpinned deps is really not good.

yakov-g: The thing is - I don't know which policy to apply for pinning it.
These are our requirements:
```
argcomplete==1.11.1
datarobot==2.21.3
docker>=4.2.2<5.0.0
flask
jinja2
memory_profiler<1.0.0
mlpiper~=2.4.0
numpy
pandas
progress
requests
scipy>=1.1,<2
strictyaml==1.0.6
texttable
py4j~=0.10.9.0
pyarrow
```
We pin only packages essential for drum. ML packages can change from user to user and we don't pin them, assuming (probably that's bad), that we use just basic functionality, like reading DF.

Maybe we should go over everything and pin versions to  `> than`

In particular, for arrow, we use 0.14.1 for DR, while the latest is 2.0.0. 
They are more than a year apart, so pinning to `pyarrow>0.14.1` is like not pinning at all.

yakov-g:

yakov-g:

yakov-g:

yakov-g:

yakov-g:

yakov-g:

yakov-g:

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**30.72** USD.

PR #259, opened 2020-12-17
[RAPTOR-4014] Don't require transform hook if custom transform uses sklearn object

jlpolit:

## Summary
- If the user defines their pipeline in sklearn, we won't require a transform hook (for custom transforms). We will assume they don't intend to transform y if they don't include a hook.
- By the same token, if a user's hook doesn't return both X and y, we will assume they don't intend to transform y and won't error or complain


model_templates/training/python3_sklearn_transform_no_hook/README.md
```diff
@@ -0,0 +1,45 @@
+## Sklearn Transform Template
```
jlpolit, line 1: I ended up adding a bunch of almost-identical model templates to cover test cases, which seems pretty lame. I'm sure there's a better way

tests/drum/test_fit.py
```diff
@@ -214,6 +211,56 @@ def test_fit(
             cmd, "Failed in {} command line! {}".format(ArgumentsOptions.MAIN_COMMAND, cmd)
         )
 
+    @pytest.mark.parametrize(
```
jlpolit, line 214: it felt like these were starting to get different enough that it made sense to break it out, but I could be wrong

yakov-g (approved):

README.md
```diff
@@ -108,6 +108,7 @@ include any necessary hooks in a file called `custom.py` for Python models or `c
         - `transformer` deserialized transformer loaded by DRUM or `load_model` 
         - `y` defaults to None, and will not be used for unsupervised downstream problems. Contains the target labels. Transform can simply pass the raw values on, may transform the values (i.e perform a log transform), 
         or use the `y` values in the transformation of `X` if needed.
+        - As of version 1.4.9, the transform hook need not return `y` if no changes are made to the target. `y` is still a required argument to this hook for custom transforms.
```
4dahalibut, line 111: why is y required?

4dahalibut: can we just check if the transform wants y?

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -57,6 +58,20 @@ def __init__(self, model_dir, target_type=None):
         self._model_dir = model_dir
         self._target_type = target_type
 
+    @staticmethod
+    def _apply_sklearn_transformer(data, model):
+        try:
+            transformed = model.transform(data)
+            if type(transformed) == csr_matrix:
+                output_data = pd.DataFrame.sparse.from_spmatrix(transformed)
+            else:
```
4dahalibut, line 67: lets make sure its a dataframe or numpy array instead of a csc?

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -122,12 +137,16 @@ def load_model_from_artifact(self):
         ):
             self._find_predictor_to_use()
 
-        # TODO: RAPTOR-4014 we eventually won't require this hook for some languages/frameworks
         if (
             self._target_type == TargetType.TRANSFORM
             and not self._custom_hooks[CustomHooks.TRANSFORM]
+            # don't require transform hook if sklearn transformer
+            and "sklearn" not in str(type(self._model))
```
4dahalibut, line 144: is this how estimators do it??

custom_model_runner/datarobot_drum/resource/predict_mixin.py
```diff
@@ -147,7 +147,9 @@ def _transform(self, logger=None):
         # make output
         if is_sparse(out_data):
             if use_arrow:
-                target_payload = make_arrow_payload(out_target) if out_target is not None else {}
+                target_payload = (
+                    make_arrow_payload(out_target, arrow_version) if out_target is not None else {}
```
4dahalibut, line 151: k

model_templates/training/python3_sklearn_transform_no_hook/README.md
4dahalibut: do we use r2d2 in tests?

custom_model_runner/datarobot_drum/drum/model_adapter.py
jlpolit: estimators check whether the object is an instance of BaseEstimator, but there's a bunch of potential transformer objects for sklearn....I can check whether it has a `transform` method as well (or instead). or we can decide exactly which sklearn classes should be supported and check that it's one of those. or some combination thereof?

custom_model_runner/datarobot_drum/resource/predict_mixin.py
jlpolit: uncaught bug from before

README.md
jlpolit: I guess by "required" I mean "we're going to pass y to this hook, but you don't actually need to do anything with it". Does it make sense to not have that requirement? I can check the number of args expected by the method and just pass in y if it wants three/has a `y` arg

model_templates/training/python3_sklearn_transform_no_hook/README.md
jlpolit: I don't know- who would know?

4dahalibut: @yakov-g

yakov-g: We don't use r2d2 in tests because it's just a tool to test users environment.

@jlpolit 
Can you just do these tests like we do in integration tests: code dir is assembled for every test case: with custom/ without custom, etc.

jlpolit: good call - will update accordingly

README.md
jlpolit: ...changed to check what transform takes

custom_model_runner/datarobot_drum/resource/transform_helpers.py
```diff
@@ -66,3 +67,16 @@ def validate_transformed_output(transformed_output, should_be_sparse=False):
     else:
         assert type(transformed_output) == pd.DataFrame
         assert transformed_output.shape[1] == 10
+
+
+def parse_multi_part_response(response):
```
jlpolit, line 72: @4dahalibut  this is a vaguely hacky way to pick apart a multi-part response and parse into a dict, but it does seem to work with all the data types we want to send and not require eval- line 175 of the predict_mixin above is how the response is now getting encoded.

custom_model_runner/datarobot_drum/drum/runtime.py
```diff
@@ -88,5 +88,9 @@ def health():
     def predict():
         return {"message": "ERROR: {}".format(exc_value)}, HTTP_513_DRUM_PIPELINE_ERROR
 
+    @model_api.route("/transform/", methods=["POST"])
```
jlpolit, line 91: @4dahalibut your original fix had this as `predict()`, let me know if this should be switched back

custom_model_runner/datarobot_drum/resource/transform_helpers.py
```diff
@@ -1,13 +1,30 @@
...
 
+from cgi import FieldStorage
 from io import BytesIO, StringIO
 
 from scipy.io import mmwrite, mmread
 from scipy.sparse.csr import csr_matrix
 from scipy.sparse import vstack
 
 
+def filter_urllib3_logging():
```
jlpolit, line 13: see: https://stackoverflow.com/questions/49338811/does-requests-properly-support-multipart-responses

custom_model_runner/datarobot_drum/drum/model_adapter.py
jlpolit: since this is just applying a "raw" sklearn pipeline or transformer with no `transform` hook, I don't know that we want to constrain the output to dense formats? I guess we can explicitly tell users that they have to set sparse_threshold to 0 for column transformers, but that doesn't necessarily seem like a good idea if their transformation results in a massive sparse matrix....

4dahalibut: My comment refers to the fact that the transformed data could be a CSC or COO matrix instead of a CSR matrix, and the code will fail badly in those cases.

4dahalibut:

4dahalibut (approved):

4dahalibut:

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**25.2** USD.

PR #262, opened 2020-12-17
[RAPTOR-4069] support sparse input to transform

jlpolit:


custom_model_runner/datarobot_drum/drum/perf_testing.py
```diff
@@ -647,11 +647,16 @@ def check_prediction_side_effects(self, target_temp_location=None):
             )
 
             if self.target_type == TargetType.TRANSFORM:
```
jlpolit, line 649: note to self: be sure to check case where both input and output of transform are sparse

jlpolit:
this is out of date now

PR #266, opened 2021-01-04
[RAPTOR-4105] add reading params from env vars

yakov-g:

## Summary
- make drum able to read params from env vars
- remove now redundant transition from env vars to cli args in start_server.sh
- set `/opt/code/start_server.sh` as an entry point in the public env

## Rationale
This is done to make CM PPS docker usage easier when starting container from either CLI or docker compose, e.g

`docker  run -e TARGET_TYPE=regression cm_pps_image`


custom_model_runner/datarobot_drum/drum/common.py
```diff
@@ -175,6 +175,32 @@ class ArgumentsOptions:
...
+    VARS_WITH_VALUES = [
+        TARGET_TYPE,
+        CODE_DIR,
+        NEGATIVE_CLASS_LABEL,
+        POSITIVE_CLASS_LABEL,
+        CLASS_LABELS_FILE,
+        ADDRESS,
+        MAX_WORKERS,
+    ]
+    BOOL_VALUES = [WITH_ERROR_SERVER, SHOW_STACKTRACE, PRODUCTION, MONITOR]
```
rvorobii, line 201: nit
VARS_WITH_VALUES -> VALUE_VARS
BOOL_VALUES           -> BOOL_VARS

rvorobii (approved):
LGTM:+1:

custom_model_runner/datarobot_drum/drum/common.py
yakov-g: ok!

rvorobii: (it's just the first one is called VARS_... and the second one is called ..._VALUES)

rvorobii: the suggested renaming makes the naming a bit more consistent imo

yakov-g: no problem!

meltzg (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**27.11** USD.

PR #269, opened 2021-01-07
[RAPTOR-4388] don't run side effects for transform

jlpolit:

## Summary

Transforms have a much wider array of potential outputs than predict, so until we come up with an equivalent to prediction side effects that apply to any and all possible transform outputs we'll just disable the side effects check for transforms. Drum will still confirm that we can spin up a server and hit the transform endpoint without error, but won't do any further checking of the output.


jlpolit:

jlpolit:

jlpolit:

4dahalibut:
pls add test for nonnumerics

4dahalibut (changes requested):
two issues

4dahalibut (approved):

yakov-g (approved):

datarobot-jarvis:
:dollar: Total cost of executing tests in Jenkins: $**8.17** USD.

PR #270, opened 2021-01-13
[RAPTOR-4423] Return colnames for sparse transform server; return both X and y format in payload

jlpolit:

## Summary

As the name suggests; our y format is different from X format in the case of sparse data, so need to pass that info back. We also don't have colnames for mtx matrices, so need to pass them separately as a csv for sparse output.


4dahalibut (approved):

scottp-ml (approved):
LGTM

PR #272, opened 2021-01-15
[RAPTOR-4156] Sampling with replacement was confusing modelers

4dahalibut:

## Summary
We no longer sample data with replacement. I mistakenly thought that sampling with replacement would only happen if the asked for sample was larger than data length which is not the case. This confused CFDSes greatly when their data had repeated rows in it.


jlpolit (approved):
you got R too! lgtm

yakov-g (approved):

PR #274, opened 2021-01-19
[RAPTOR-4243] Update error messages 

scottp-ml:

## Summary
Add a better error message for the case when multiple custom.py/R files are present in subdirectories.  The current message does not make it clear why the model failed to run correctly.


jlpolit (approved):
LGTM

4dahalibut (approved):

PR #277, opened 2021-01-22
[RAPTOR-4487] Warn but don't error in DRUM for prediction consistency check

jlpolit:

## Summary

We're going to allow users to "fail" the prediction consistency check in both drum and DR, but still raise a warning so they're aware if there's a potential issue.


jlpolit:
jarvis test functional_inference_model_templates

4dahalibut (approved):

yakov-g (approved):

PR #278, opened 2021-01-22
[RAPTOR-4443] handle case where X matrix has colname that's the same as target column

jlpolit:


custom_model_runner/datarobot_drum/drum/common.py
```diff
@@ -361,3 +363,10 @@ def verify_pyarrow_module():
     if pyarrow is None:
         raise ModuleNotFoundError("Please install pyarrow to support Arrow format")
     return pyarrow
+
+
+def generate_temp_target_name(target_name, colnames):
```
jlpolit, line 368: this might be overkill?

4dahalibut: Yeah maybe we should just do something like __target__ here.

scottp-ml (approved):
LGTM

4dahalibut (approved):

PR #279, opened 2021-01-25
[RAPTOR-4438] drum server fails if labels have spaces

jlpolit:

## Summary

Fit was failing during performance testing for binary projects with spaces in the class names because we start the server by passing class labels, and the unescaped spaces make bash angry. 

I don't know if we deal with this problem when we predict in DR (running some models now to check)...and they look fine, so seems to be only an issue on the drum side which is nice.

Two thoughts:
1. Do we really want to be running the performance test during fit in DR?
2. will multiclass have the same problem? (@meltzg )


meltzg:
@jlpolit there are 2 ways of passing class labels for multiclass
1. a file with newline delimited labels. This is what is used by datarobot. Spaces should not be an issue
2. a list of labels in the commandline.  https://github.com/datarobot/datarobot-user-models/blob/af4a4c6b09aa89f5392e935547962ddb5cd935e8/tests/drum/test_args_parser.py#L31-L46 tests this.  Since the list is already space-delimited, you already need quotes for labels with spaces

4dahalibut:
why are we running performance test during fit within DR?

4dahalibut:
We should not be

jlpolit:
> why are we running performance test during fit within DR?

@4dahalibut we shouldn't be, in my opinion...happy to make a ticket to bypass that if folks are on board

4dahalibut (approved):

yakov-g (approved):

PR #281, opened 2021-01-26
[RAPTOR-4506] skip predict if SKIP_PREDICT env var is present

jlpolit:


custom_model_runner/datarobot_drum/drum/main.py
```diff
@@ -115,6 +115,10 @@ def _is_arg_registered_in_subparser(sub_parser, arg):
             if sub_parser is None:
                 return
 
+            skip_predict = os.environ.get("SKIP_PREDICT")
+            if skip_predict is not None:
+                sys.argv.extend([ArgumentsOptions.SKIP_PREDICT])
+
             for env_var_key in ArgumentOptionsEnvVars.VALUE_VARS + ArgumentOptionsEnvVars.BOOL_VARS:
```
yakov-g, line 122: Why don't you use existing boilerplate?
Add SKIP_PREDICT into ArgumentOptionsEnvVars.BOOL_VARS ...

yakov-g (changes requested):

custom_model_runner/datarobot_drum/drum/main.py
jlpolit: sure

4dahalibut:
you don't want to always just skip predict?

jlpolit:
> you don't want to always just skip predict?

I assumed we want to let users run prediction checks locally in drum, while giving an easy way to bypass in DR

yakov-g (approved):

meltzg (approved):

PR #284, opened 2021-02-04
[RAPTOR-4593] refactor YAML schema for model-metadata.yaml

yakov-g:

## Summary
* added `validate_config_fields(config, *fields)` which checks if provided fields present in config
* make validation and environmentId fields optional and use `validate_config_fields` to ensure fields present according to functionality
* added `customPredictor` section. It can contain any data that user plans to add and parse on his own. DRUM doesn't use this field, but it has to be defined to make DRUM not fail when it present.

## Rationale
Current YAML parsing schema for model-metadata.yaml treats `validation`, `environmentId` fields as mandatory which is relevant only for `drum push` feature. In general, required fields should be checked according to functionality.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/common.py
+++ b/custom_model_runner/datarobot_drum/drum/common.py
@@ -321,7 +321,7 @@ def validate_config_fields(model_config, *fields):
         if f not in model_config:
             missing_sections.append(f)
 
-    if len(missing_sections):
+    if missing_sections:
         raise DrumCommonException(
             "The following keys are missing in {} file.\n"
             "Missing keys: {}".format(MODEL_CONFIG_FILENAME, missing_sections)
--- a/custom_model_runner/datarobot_drum/drum/common.py
+++ b/custom_model_runner/datarobot_drum/drum/common.py
@@ -4,8 +4,9 @@ import sys
 
 from contextlib import contextmanager
 from enum import Enum
-from strictyaml import Bool, Int, Map, Optional, Str, load, YAMLError, Seq
+from strictyaml import Bool, Int, Map, Optional, Str, load, YAMLError, Seq, Any
 from pathlib import Path
+from datarobot_drum.drum.exceptions import DrumCommonException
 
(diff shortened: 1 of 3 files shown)


custom_model_runner/datarobot_drum/drum/common.py
```diff
@@ -290,11 +309,25 @@ def config_logging():
...
 )
 
 
+def validate_config_fields(model_config, *fields):
+    missing_sections = []
+    for f in fields:
+        if f not in model_config:
+            missing_sections.append(f)
+
+    if len(missing_sections):
```
rvorobii, line 324: change according to PEP-8 (https://www.python.org/dev/peps/pep-0008/)
image


```suggestion
    if missing_sections:
```

tests/drum/test_units.py
```diff
@@ -185,6 +204,51 @@ def training_metadata_yaml_with_proj():
...
+        with open(os.path.join(temp_dir, MODEL_CONFIG_FILENAME), mode="w") as f:
+            f.write(config_yaml)
+        read_model_metadata_yaml(temp_dir)
+    finally:
+        shutil.rmtree(temp_dir)
+
+
+def test_yaml_metadata_missing_fields(custom_predictor_metadata_yaml):
+    try:
+        temp_dir = tempfile.mkdtemp()
```
rvorobii, line 235: use pytest's `tmp_path` fixture -- pytest will take care of the cleanup
https://docs.pytest.org/en/stable/tmpdir.html

rvorobii (approved):
The changes LGTM

Although, it gets complicated - some fields are required only if other fields have specific values (e.g. if "type" == "training" than "environmentId", "validation" is required for DRUM push etc).

We may need to consider using validation library like "trafaret" for this.
For example
- there can be a common trafaret for both options and model config for `drum push`
- we can define that model type is either "training" and "inference", each having its own trafaret with field requirements

yakov-g:

meltzg (approved):

PR #288, opened 2021-02-11
[RAPTOR-4520,4712] fix dataset temp filepath naming, release v1.4.15

yakov-g:

## Summary
1.
Fix temp dataset filepath generation during validation.
I've been using a column name in the filename. So in this bug, it happen that column's name was `smth/smth` :)

Replace filename with `null_value_imputation_columnXYZ`

2.
Fix a small regression I introduced, for the case of calling `drum` without any params.

3.
Added a fix for Py/R objects casting. I've been changing type of the `_class_labels` in-place, but this parameter is passed into 
`datarobot.mlops` API after predictions are made.


yakov-g:

yakov-g:

custom_model_runner/datarobot_drum/drum/perf_testing.py
```diff
@@ -523,25 +523,21 @@ def validation_test(self):
...
-            ) as temp_f:
-                temp_data_name = temp_f.name
-                df_tmp = df.copy()
-                df_tmp[column_name] = None
-                df_tmp.to_csv(temp_data_name, index=False)
-                CMRunnerUtils.replace_cmd_argument_value(
-                    cmd_list, ArgumentsOptions.INPUT, temp_data_name
-                )
+        for i, column_name in enumerate(column_names):
+            tmp_dataset_file_path = os.path.join(
```
rvorobii, line 527: doesn't this need to be removed after it's used?

custom_model_runner/datarobot_drum/drum/perf_testing.py
```diff
@@ -523,25 +523,21 @@ def validation_test(self):
...
+                null_datasets_dir, "null_value_imputation_column{}".format(i)
```
rvorobii, line 528: how does this fixes the issue?

the file name still can be "null_value_imputation_columnsmth/smth" if column name is "smth/smth"

yakov-g: I'm not adding a column name into path anymore

custom_model_runner/datarobot_drum/drum/perf_testing.py
yakov-g: Whole folder is removed if test passes and kept if test fails.

custom_model_runner/datarobot_drum/drum/perf_testing.py
rvorobii: oh, you're right, sorry

rvorobii (approved):

meltzg (changes requested):
need a test testing that perf testing works with in `/` column names

yakov-g:
Added a test case which fails without the path fix.

yakov-g:
@rvorobii 
added more testing and one more fix for https://datarobot.atlassian.net/browse/RAPTOR-4712
I want them to go in ASAP.

meltzg (approved):

PR #289, opened 2021-02-12
[RAPTOR-4564] add validation utility for expected Dockerfile entries

scottp-ml:

## Summary
Adds a command to check for expected entries in the dockerfile that users might not otherwise know are required to run within DataRobot.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/args_parser.py
+++ b/custom_model_runner/datarobot_drum/drum/args_parser.py
@@ -694,6 +694,20 @@ class CMRunnerArgsRegistry(object):
         return parser
 
     @staticmethod
+    def _register_subcommand_check_docker(subparsers):
+        description = """
+        Verify that expected entries are present in the Docker file for your custom environment.  
+        """
+        parser = subparsers.add_parser(
+            ArgumentsOptions.CHECK_DOCKER,
+            help="Check for required entries in a docker file for a custom environment.",
+            description=description,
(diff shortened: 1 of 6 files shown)


yakov-g:
Hey @scottp-ml ,

I'm not sure about these validations.
When someone creates his own environment they only need to follow the rules: https://github.com/datarobot/datarobot-user-models/#custom_environment_templates.

Even if they take our env as an example they can organize Dockerfile in many ways, e.g:
* _base_image - customer's env may not use datarobot base env
* _dr_requirements - that's just the way we organize our envs
* _code_dir_copy - they may be copying files one by one instead of `./`
* _code_dir_env - may be, but can be provided as a param to drum command
*  _address - may be, but can be provided as a param to drum command
* _entrypoint - not required at all (I probably should remove these)

As a general rule, we recommend testing model and env before uploading them to DR.
When running DRUM with `--docker` param:
* if image is provided to `--docker` this image will be started
* if docker context is provided to `--docker` image will be built and started.
* 
So if user has an image or a context he already can check locally and see if it works. This could be improved:
* we could generate `Dockerfile` (or just required commands) and `start_server.sh`
* we could validate already built image; + improve error messages

Ping me to discuss

4dahalibut:
I think this validation should (in addition? ) live within DR

yakov-g:
> I think this validation should (in addition? ) live within DR

I think we have startup tests for that, when assemble the model

PR #292, opened 2021-02-18
[RAPTOR-3911] show progress spinner when building a docker image, release v1.4.16

yakov-g:

### Added
* show progress spinner while building docker image (when provide context folder to --docker)
### Fixed
* fix:  always unset `entrypoint` when we run DRUM in a container

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/drum.py
+++ b/custom_model_runner/datarobot_drum/drum/drum.py
@@ -629,7 +629,7 @@ class CMRunner:
         in_docker_fit_target_filename = "/opt/fit_target.csv"
         in_docker_fit_row_weights_filename = "/opt/fit_row_weights.csv"
 
-        docker_cmd = "docker run --rm --interactive --user $(id -u):$(id -g) "
+        docker_cmd = "docker run --rm --entrypoint '' --interactive --user $(id -u):$(id -g)"
         docker_cmd_args = " -v {}:{}".format(options.code_dir, in_docker_model)
 
         in_docker_cmd_list = raw_arguments
@@ -731,7 +731,17 @@ class CMRunner:
 
         self._print_verbose("Checking DRUM version in container...")
         result = subprocess.run(
-            ["docker", "run", "-it", options.docker, "sh", "-c", "drum --version"],
+            [
+                "docker",
+                "run",
+                "-it",
+                "--entrypoint",
+                "",
+                options.docker,
+                "sh",
+                "-c",
+                "drum --version",
+            ],
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
         )
@@ -748,6 +758,10 @@ class CMRunner:
             if len(err):
                 print(err)
             time.sleep(0.5)
+        else:
+            self._print_verbose(
+                "Host DRUM version matches container DRUM version: {}".format(host_drum_version)
+            )
         self._print_verbose("-" * 20)
         p = subprocess.Popen(docker_cmd, shell=True)
         try:
@@ -777,8 +791,10 @@ class CMRunner:
                 # If image with the tag `my_env` exists, it will be untagged.
                 tag = os.path.basename(docker_image_or_directory)
                 client_docker_low_level = docker.APIClient()
-                spinner = Spinner('Building docker image: ')
-                for _ in client_docker_low_level.build(path=docker_image_or_directory, rm=True, tag=tag):
+                spinner = Spinner("Building docker image: ")
+                for _ in client_docker_low_level.build(
+                    path=docker_image_or_directory, rm=True, tag=tag
+                ):
                     spinner.next()
                 print("\nImage built, tag: {}\n".format(tag))
 
--- a/custom_model_runner/datarobot_drum/drum/drum.py
+++ b/custom_model_runner/datarobot_drum/drum/drum.py
@@ -3,6 +3,7 @@ import glob
 import json
 import logging
 import os
+import re
 import shutil
 import subprocess
 import sys
@@ -777,9 +778,10 @@ class CMRunner:
 
         if os.path.isdir(docker_image_or_directory):
             docker_image_or_directory = os.path.abspath(docker_image_or_directory)
-            self.logger.info(
-                "Building a docker image from directory {}...".format(docker_image_or_directory)
+            docker_build_msg = "Building a docker image from directory: {}...".format(
+                docker_image_or_directory
             )
+            self.logger.info(docker_build_msg)
             self.logger.info("This may take some time")
             try:
                 # Set image tag to the dirname of the docker context.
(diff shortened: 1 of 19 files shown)


meltzg (commented):
real important shit right here

meltzg (approved):

yakov-g:
@meltzg 
I realized that low level APIClient doesn't throw BuildError exception.
So I had to parse status messages. ( The same way it is done in the docker APIs itself: https://github.com/docker/docker-py/blob/e6689e0bb9af849bd9d1509fd3b2db52e0d6d776/docker/models/images.py#L285-L297)

Added tests

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -730,7 +732,17 @@ def _run_inside_docker(self, options, run_mode, raw_arguments):
 
         self._print_verbose("Checking DRUM version in container...")
         result = subprocess.run(
-            ["docker", "run", "-it", options.docker, "sh", "-c", "drum --version"],
+            [
+                "docker",
+                "run",
+                "-it",
+                "--entrypoint",
+                "",
```
meltzg, line 740: what's this empty arg for?

meltzg (approved):
one question, but otherwise g2g

custom_model_runner/datarobot_drum/drum/drum.py
4dahalibut: +1

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -759,26 +775,65 @@ def _run_inside_docker(self, options, run_mode, raw_arguments):
...
-            except docker.errors.BuildError as e:
-                self.logger.error("Hey something went wrong with image build!")
-                for line in e.build_log:
+                # Set image tag to the dirname of the docker context.
+                # This may be confusing if try to build different images from different contexts,
+                # with the same folder name:
+                # /path1/my_env
+                # /path2/my_env
+                #
+                # If image with the tag `my_env` exists, it will be untagged.
```
4dahalibut, line 793: why this

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -759,26 +775,65 @@ def _run_inside_docker(self, options, run_mode, raw_arguments):
...
+                tag = os.path.basename(docker_image_or_directory)
+                client_docker_low_level = docker.APIClient()
+                spinner = Spinner(docker_build_msg + "  ")
+                json_lines = []
+                # Build docker, rotate spinner according to build progress
+                # and save status messages from docker build.
+                for line in client_docker_low_level.build(
+                    path=docker_image_or_directory, rm=True, tag=tag
```
4dahalibut, line 801: if you really wanted progress you could do something like find the total dockerfile lines and print progress per line that was built or someth

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -759,26 +775,65 @@ def _run_inside_docker(self, options, run_mode, raw_arguments):
...
+                        build_error = True
+                        break
                     if "stream" in line:
-                        self.logger.error(line["stream"].strip())
+                        match = re.search(
+                            r"(^Successfully built |sha256:)([0-9a-f]+)$", line["stream"]
+                        )
+                        if match:
+                            image_id = match.group(2)
+                if image_id is None or build_error:
```
4dahalibut, line 820: probably best to have different errors here.

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -759,26 +775,65 @@ def _run_inside_docker(self, options, run_mode, raw_arguments):
...
+                    raise DrumCommonException(
+                        "Failed to build a docker image:\n{}".format(all_lines)
+                    )
+
+                print("\nImage successfully built; tag: {}; image id: {}\n".format(tag, image_id))
+
+                ret_docker_image = image_id
+            except docker.errors.APIError as e:
+                self.logger.error("Hey something went wrong with image build!")
+                self.logger.error(str(e))
```
4dahalibut, line 831: you can do logger.exception, and maybe less jovial wording?

tests/drum/test_drum_docker.py
```diff
@@ -0,0 +1,77 @@
...
+        [True, False],
+    )
+    def test_docker_image_creation(
+        self,
+        docker_build_fails,
+        tmp_path,
+    ):
+        # py-slim image is used in another test container, so it is already expected to be in the registry
+        dockerfile = dedent(
+            """
```
4dahalibut, line 27: test for if build is canceled out of?

custom_model_runner/datarobot_drum/drum/drum.py
yakov-g: Passing empty entrypoint
`--entrypoint ""`

yakov-g: Unsetting if one is set within the image

meltzg: might be worth leaving a comment

custom_model_runner/datarobot_drum/drum/drum.py
yakov-g: It's nice to tag the image, and the easiest way is to take dir name.

I could add a param, `--tag` or `--docker-tag` which will be used if dir path is provided with `--docker`, but i'd hold on until more feedback.

custom_model_runner/datarobot_drum/drum/drum.py
yakov-g: Good idea....

custom_model_runner/datarobot_drum/drum/drum.py
yakov-g: I took this code snippet from:
https://github.com/docker/docker-py/blob/e6689e0bb9af849bd9d1509fd3b2db52e0d6d776/docker/models/images.py#L285-L297

As you can see they only rely on `Successfully built`, so in our case - if there is no image id just show all the messages and throw,

custom_model_runner/datarobot_drum/drum/drum.py
yakov-g: done

custom_model_runner/datarobot_drum/drum/drum.py
yakov-g: fixed

custom_model_runner/datarobot_drum/drum/drum.py
yakov-g: I tried this.
But drawback is that for a step `RUN pip install datarobot-drum` there is a lot of output: several lines for each pulled dependency.
For progressbar, to increment it properly, I need to estimate number of output lines, while for spinner - i just spin it on every output.

E.g:
3 steps in Dockerfile, each step outputs 100 lines: in case of progress bar I can only set it''s length to 3 and promote on every output, that contains `Step`.

4dahalibut (approved):

PR #293, opened 2021-02-24
[RAPTOR-4012] try to install model deps when docker context provided to --docker

yakov-g:
# This repository is public. Do not put here any private DataRobot or customer's data: code, datasets, model artifacts, .etc.

## Summary
When Python/R code dir contains `requirements.txt` and user provides a folder with Dockerfile to `--docker` command,
DRUM will try install dependencies while building an image.

Note:
- DRUM doesn't try to install deps into user's env, when running without `--docker`, as I suppose user knows what is going on in his env.
- DRUM doesn't try to install deps into existing image.... (maybe we should??), but that was not in the current ticket.

## Rationale
User's model may use additional dependencies listed in requirments.txt. These requirements a handled in DR app when environment is built.
But it is not handeld by drum. So if user wants to test his model locally with the env, deps won't be installed automatically,
This feature solves this.

PR #294, opened 2021-02-26
[RAPTOR-4790,4791] bug fixes, release v1.5.0

yakov-g:

## Summary
* return proper response if user tries to requiest /predict/ endpoint in unstructured mode; or /predictUnstructured endpoint in structured mode
* fix perf-test table formatting
* 
release v1.5.0
This PR will go after https://github.com/datarobot/datarobot-user-models/pull/293

### DIFF
--- a/custom_model_runner/datarobot_drum/resource/predict_mixin.py
+++ b/custom_model_runner/datarobot_drum/resource/predict_mixin.py
@@ -250,7 +250,7 @@ class PredictMixin:
 
     def do_predict(self, logger=None):
         wrong_target_type_error_message = (
-            "This project has target type '{}', use the {{}} endpoint.".format(
+            "This model has target type '{}', use the {{}} endpoint.".format(
                 self._target_type.value
             )
         )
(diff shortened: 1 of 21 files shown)


meltzg (approved):

custom_model_runner/datarobot_drum/resource/predict_mixin.py
```diff
@@ -249,19 +249,42 @@ def _transform(self, logger=None):
         return response, response_status
 
     def do_predict(self, logger=None):
+        wrong_target_type_error_message = (
+            "This project has target type '{}', use the {{}} endpoint.".format(
```
klichukb, line 253: Should we refer to it as to model instead of project?

klichukb (approved):

custom_model_runner/datarobot_drum/resource/predict_mixin.py
yakov-g: yeah....

PR #295, opened 2021-03-09
[RAPTOR-4830] refactor data transfer from py to java via py4j

yakov-g:

## Summary
Performance degradation investigation happened because DRUM changed the way how data is passed from Py to Java via py4j.
Before the change, incoming data was saved as a CSV file, file name was passed into predictor, and predictor read the file.
I changed this to directly pass the incoming data array into the predictor.

As we saw in the validation results, Py and R improved some performance, while Java gained some improved performance on the small data, but lost on the big chunks. (I confirmed that by switching back to files).

Implement hybrid approach where chunks of up to 33K will be sent directly as array, while bigger chunks will be saved into temp file.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/language_predictors/java_predictor/java_predictor.py
+++ b/custom_model_runner/datarobot_drum/drum/language_predictors/java_predictor/java_predictor.py
@@ -9,6 +9,7 @@ import atexit
 import pandas as pd
 import re
 from itertools import chain
+import tempfile
 
 from io import StringIO
 from contextlib import closing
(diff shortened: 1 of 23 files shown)


custom_model_runner/datarobot_drum/drum/language_predictors/java_predictor/java_predictor.py
```diff
@@ -150,10 +150,20 @@ def has_read_input_data_hook(self):
...
-        input_text_data = kwargs.get(StructuredDtoKeys.BINARY_DATA).decode("utf-8")
+        input_text_bytes = kwargs.get(StructuredDtoKeys.BINARY_DATA)
 
         start_predict = time.time()
-        out_csv = self._predictor_via_py4j.predict(input_text_data)
+        # If data size is more than 33K, pass it as a file to Java,
+        # as passing big chunks to py4j as an array is 10% slower
+        DATA_BUFFER_LIMIT_33K = 33792
+        if len(input_text_bytes) > DATA_BUFFER_LIMIT_33K:
+            with tempfile.NamedTemporaryFile(mode="wb") as tf:
```
zohar-mizrahi, line 160: Have you considered using a memory mapped file? This may increase performance in a great deal.

timsetsfire (approved):

custom_model_runner/datarobot_drum/drum/language_predictors/java_predictor/java_predictor.py
yakov-g: created a ticket

PR #296, opened 2021-03-15
[RAPTOR-4880] Sparse columns support

4dahalibut:

## Summary

We added sparse data support with the mtx extension, but need to pass columns in a separate file as a result.


custom_model_runner/datarobot_drum/resource/transform_helpers.py
```diff
@@ -67,17 +67,11 @@ def read_csv_payload(response_dict, transform_key):
 
 
 def make_mtx_payload(df):
```
jlpolit, line 69: :+1: :+1:

custom_model_runner/datarobot_drum/drum/args_parser.py
```diff
@@ -472,6 +472,18 @@ def _reg_arg_num_rows(*parsers):
                 "Set to ALL to use all rows. Default is 100",
             )
 
+    @staticmethod
+    def _reg_arg_sparse_colfile(*parsers):
+        for parser in parsers:
+            parser.add_argument(
+                ArgumentsOptions.SPARSE_COLFILE,
+                default=None,
```
yakov-g, line 480: Worth adding `type=CMRunnerArgsRegistry._is_valid_file`?

4dahalibut: yeah sure

custom_model_runner/datarobot_drum/resource/predict_mixin.py
```diff
@@ -47,7 +47,7 @@ def _validate_content_type_header(header):
         return ret_mimetype, ret_charset
 
     @staticmethod
-    def _fetch_data_from_request(file_key, logger=None):
+    def _fetch_data_from_request(file_key, logger=None, optional=False):
```
yakov-g, line 50: I think `resolve_mimetype_by_filename` is going to fail for sparse.colnames file

yakov-g, line 50: I don't really like this construction with `optional` and  `elif len(request.data) and file_key == "X"; looks confusing.
Maybe add another method: `_fetch_additional_files_from_request`?

jlpolit (approved):

custom_model_runner/datarobot_drum/resource/predict_mixin.py
yakov-g: No, it won't . My mistake.

custom_model_runner/datarobot_drum/resource/predict_mixin.py
```diff
@@ -95,7 +97,11 @@ def _check_mimetype_support(self, mimetype):
     def _predict(self, logger=None):
         response_status = HTTP_200_OK
         try:
+
             binary_data, mimetype, charset = self._fetch_data_from_request("X", logger=logger)
+            sparse_data, _, _ = self._fetch_data_from_request(
+                "X.colnames", logger=logger, optional=True
```
yakov-g, line 103: I suggest to move `X.colnames` into a constant.... 
Would be nice if you could also move `X` into a constant

public_dropin_environments/python3_keras/fit.sh
```diff
@@ -17,6 +18,10 @@ if [ -f "${weights}" ]; then
     CMD="${CMD} --row-weights-csv ${weights}"
 fi
 
+if [ -f "${sparse_colnames}" ]; then
+    CMD="${CMD} --sparse-column-file ${sparse_colnames}"
```
yakov-g, line 22: I suggest to create env var for `--sparse-column-file`, once this env var is set, DRUM will automatically pick it up.
Boilerplate is there:
* add into proper record into ArgumentOptionsEnvVars and ArgumentOptionsEnvVars.VALUE_VARS....

What actually happens: in the beginning, if env var is set, parameter will be added into sys.argv.

yakov-g: So you won't need this condition in every environment.

yakov-g: @jlpolit did it for SKIP_PREDICT

yakov-g (approved):
I approve, but consider addressing my comments

public_dropin_environments/python3_keras/fit.sh
4dahalibut: I'd like for people to be able to use this on the cli though.

custom_model_runner/datarobot_drum/resource/predict_mixin.py
4dahalibut: It seems to work for me.

public_dropin_environments/python3_keras/fit.sh
yakov-g: it doesn't cancel providing argument explicitly as `--sparse-column-file`

PR #298, opened 2021-03-18
[RAPTOR-4967] Add parameters to DRUM metadata schema

jmbannon:

## Summary
Custom training models will be getting hyperparameter support in 7.1 via supplying it in the model-metadata YAML.
This PR allows the field. We will make it more specific once we add param validation in DRUM.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/common.py
+++ b/custom_model_runner/datarobot_drum/drum/common.py
@@ -306,6 +306,7 @@ class ModelMetadataKeys(object):
     # customPredictor section is not used by DRUM,
     # it is a place holder if user wants to add some fields and read them on his own
     CUSTOM_PREDICTOR = "customPredictor"
+    PARAMETERS = 'parameters'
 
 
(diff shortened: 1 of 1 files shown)


custom_model_runner/datarobot_drum/drum/common.py
```diff
@@ -306,6 +306,7 @@ class ModelMetadataKeys(object):
     # customPredictor section is not used by DRUM,
     # it is a place holder if user wants to add some fields and read them on his own
     CUSTOM_PREDICTOR = "customPredictor"
+    PARAMETERS = "parameters"
```
4dahalibut, line 309: not hyperparameters?

jmbannon: changed to hyperparameters (its a single word so no camelcase on the 'p')

4dahalibut (approved):

scottp-ml (approved):
LGTM

PR #301, opened 2021-03-19
[RAPTOR-4975] add validation parameter to mode-metadata.yaml

scottp-ml:

## Summary
Custom training models will be getting support for input type validation in 7.1.  The specification will be provided through the model-metatdata YAML file.  This adds that field for future use.


4dahalibut (approved):

yakov-g:
Don't we want to put it under Training Models section?

scottp-ml:
I think it makes sense to be able to define the expected input and output types for any model.

jmbannon (approved):

PR #304, opened 2021-03-24
[RAPTOR-4897] early detection of missing labels for binary/multiclass

yakov-g:

## Summary
Binary and multiclass labels can be provided as CLI parameters or in metadata yaml. Eventually labels from yaml has never been plumbed through.

In this patch I resolve parameters if they are provided in CLI, yaml, or both, add testcases and do some clean up, like:
- require yaml  for PUSH mode
- don't resolve labels for FIT
- remove class labels CLI arguments from push and fit parsers, etc.

Sorry if PR is a bit long and cumbersome

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/common.py
+++ b/custom_model_runner/datarobot_drum/drum/common.py
@@ -293,6 +293,18 @@ def config_logging():
     logging.basicConfig(format="%(asctime)-15s %(levelname)s %(name)s:  %(message)s")
 
 
+def get_metadata(options):
+    code_dir = Path(options.code_dir)
+    if options.model_config is None:
+        raise DrumCommonException(
+            "You must have a file with the name {} in the directory {}. \n"
+            "You don't. \nWhat you do have is these files: \n{} ".format(
+                MODEL_CONFIG_FILENAME, code_dir, os.listdir(code_dir)
+            )
+        )
+    return options.model_config
+
+
 class ModelMetadataKeys(object):
     NAME = "name"
     TYPE = "type"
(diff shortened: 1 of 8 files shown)


custom_model_runner/datarobot_drum/drum/common.py
```diff
@@ -362,6 +374,51 @@ def read_model_metadata_yaml(code_dir):
             except YAMLError as e:
                 print(e)
                 raise SystemExit(1)
+
+        if model_config[ModelMetadataKeys.TARGET_TYPE] == TargetType.BINARY.value:
+            if model_config[ModelMetadataKeys.TYPE] == "inference":
```
yakov-g, line 379: Validate class labels only for inference models

custom_model_runner/datarobot_drum/drum/common.py
```diff
@@ -293,6 +293,18 @@ def config_logging():
     logging.basicConfig(format="%(asctime)-15s %(levelname)s %(name)s:  %(message)s")
 
 
+def get_metadata(options):
```
yakov-g, line 296: Just moving this method from another place

yakov-g:

rvorobii (approved):

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -64,56 +66,150 @@ def __init__(self, runtime):
...
+        ):
+            self.options.positive_class_label = None
+            self.options.negative_class_label = None
+            self.options.class_labels = None
+            self.options.class_labels_file = None
+            return
+
+        if self.target_type == TargetType.BINARY:
+            pos_options = getattr(self.options, "positive_class_label", None)
+            neg_options = getattr(self.options, "negative_class_label", None)
```
4dahalibut, line 139: why do you need getattr here?

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -64,56 +66,150 @@ def __init__(self, runtime):
...
+
+            try:
+                pos_model_config = self.options.model_config.get(
```
4dahalibut, line 142: you don't need .get here.

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -64,56 +66,150 @@ def __init__(self, runtime):
...
+                )
+        else:
+            if self.options.content_type is None:
+                self.options.content_type = "text/plain; charset=utf8"
+
+    def _resolve_class_labels(self):
+        if (
+            self.run_mode in [RunMode.NEW]
+            or self.run_mode == RunMode.PUSH
+            and self.options.model_config[ModelMetadataKeys.TYPE] == "training"
```
4dahalibut, line 129: can you add parentheses here?

custom_model_runner/datarobot_drum/drum/drum.py
yakov-g: It's more stable. If i start to play around with parsers: register/unregister params from parsers it won't fail

custom_model_runner/datarobot_drum/drum/drum.py
yakov-g: This is the way to avoid using a check `if key in dict`:
If key is absent get returns None. Trying to call get on None causes AttributeError.

custom_model_runner/datarobot_drum/drum/drum.py
yakov-g: added

custom_model_runner/datarobot_drum/drum/drum.py
yakov-g: Consider the case: model type is training, but has Inference section.... what should we do?
Because inference section is the same as providing labels in CLI.

4dahalibut (approved):

PR #307, opened 2021-03-29
[RAPTOR-4478] cleanup BasePredictor: add abstract configure() method

yakov-g:

## Summary
Make configure() method abstract to make developer implement it in their code.
Fix signature in the Scala code: Java Object type has to be mapped to AnyRef type in Scala.


timsetsfire (approved):

rvorobii (approved):
add `@Override` to existing children of `BasePredictor`: 
- `ScoringCode`
- `TestCustomPredictor`

other than that - looks good to me

PR #308, opened 2021-03-30
[RAPTOR-5017,5019] fix documentation on the API endpoints, release v1.5.3

yakov-g:

## Summary
Add documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,11 +304,13 @@ If you'd like to use a tool/language/framework that is not supported by our temp
 1) Your environment must include a Dockerfile that installs any requirements you may want.
 2) Custom models require a simple webserver in order to make predictions. We recommend putting this in
 your environment so that you can reuse it with multiple models. The webserver must be listening on port 8080 and implement the following routes:
-    1) `GET /{URL_PREFIX}/` This route is used to check if your model's server is running.
-    2) `POST /URL_PREFIX/predict/` This route is used to make predictions.
+   > **Note: `URL_PREFIX` is an environment variable that will be available at runtime. It has to be read and pasted into the routes.**
+    1) `GET /URL_PREFIX/` and `GET /URL_PREFIX/ping/` These routes are used to check if your model's server is running.
+    2) `GET /URL_PREFIX/health/` This route is used to check if model is loaded and functioning properly.
+    3) `POST /URL_PREFIX/predict/` This route is used to make predictions.
 3) An executable `start_server.sh` file is required to start the model server.
 4) Any code and `start_server.sh` should be copied to `/opt/code/` by your Dockerfile
-> Note: `URL_PREFIX` is an environment variable that will be available at runtime.
+
 
 ## Custom Model Runner <a name="custom_model_runner"></a>
 Custom model runner (DRUM) is a  tool that helps to assemble, test, and run custom models.
--- a/README.md
+++ b/README.md
@@ -204,7 +204,7 @@ Type checking methods can be used to verify types:
 - in R `is.character(data)` or `is.raw(data)`
 
 DRUM uses `Content-Type` header to determine a type to cast `data` to. Content-Type header can be provided in request or in `--content-type` CLI argument.  
-`Content-Type` header format is `type/subtype;parameter`, e.g. `text/plain;charset=utf8`. Only minetype part `text/plain`, and `charset=utf8` parameter matter for DRUM.  
+`Content-Type` header format is `type/subtype;parameter`, e.g. `text/plain;charset=utf8`. Only mimetype part `text/plain`, and `charset=utf8` parameter matter for DRUM.  
 Following rules apply:
 - if charset is not defined, default `utf8` charset is used, otherwise provided charset is used to decode data;
 - if content type is not defined, then incoming `kwargs={"mimetype": "text/plain", "charset":"utf8"}`, so data is treated as text, decoded using `utf8` charset and passed as `str`;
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -304,11 +304,13 @@ If you'd like to use a tool/language/framework that is not supported by our temp
 1) Your environment must include a Dockerfile that installs any requirements you may want.
 2) Custom models require a simple webserver in order to make predictions. We recommend putting this in
 your environment so that you can reuse it with multiple models. The webserver must be listening on port 8080 and implement the following routes:
-    1) `GET /{URL_PREFIX}/` This route is used to check if your model's server is running.
-    2) `POST /URL_PREFIX/predict/` This route is used to make predictions.
+   > **Note: `URL_PREFIX` is an environment variable that will be available at runtime. It has to be read and pasted into the routes.**
+    1) `GET /URL_PREFIX/` and `GET /URL_PREFIX/ping/` These routes are used to check if your model's server is running.
+    2) `GET /URL_PREFIX/health/` This route is used to check if model is loaded and functioning properly.
+    3) `POST /URL_PREFIX/predict/` This route is used to make predictions.
```
yakov-g, line 310: @rvorobii  @klichukb  Should /stats/ endpoint be here?

README.md
```diff
@@ -304,11 +304,13 @@ If you'd like to use a tool/language/framework that is not supported by our temp
 1) Your environment must include a Dockerfile that installs any requirements you may want.
 2) Custom models require a simple webserver in order to make predictions. We recommend putting this in
 your environment so that you can reuse it with multiple models. The webserver must be listening on port 8080 and implement the following routes:
-    1) `GET /{URL_PREFIX}/` This route is used to check if your model's server is running.
-    2) `POST /URL_PREFIX/predict/` This route is used to make predictions.
+   > **Note: `URL_PREFIX` is an environment variable that will be available at runtime. It has to be read and pasted into the routes.**
+    1) `GET /URL_PREFIX/` and `GET /URL_PREFIX/ping/` These routes are used to check if your model's server is running.
+    2) `GET /URL_PREFIX/health/` This route is used to check if model is loaded and functioning properly.
```
rvorobii, line 309: /health route is not required for execution environment

IMO we should say that the required minimum is 
1. `/`
2. `/predict`

then, users can have "extensions":
- `/stats` to get memory usage measurements and charts in CM Testing Performance Check and Stability Check.
- `/health` to be able to properly propagate server initialisation errors to CM Testing Startup Check. Custom models are run in k8s, so if a container fails k8s will attempt to restart it. If model loading fails -- there's no point in failing the model container, because k8s will restart it, then model loading will fail again, and as a result container will go into crash/restart loop (for several minutes). To avoid such a case, if model loading loading fails - model container should not fail, instead the server should be up with `/` route, and the error should be returned via `/health` route with 513 response code.

custom_model_runner/README.md
```diff
@@ -149,11 +149,27 @@ DRUM can also run as a prediction server. To do so, provide a server address arg
...
+* Info route:  
+A GET **URL_PREFIX/info/** route, shows information about running model (metadata, paths, predictor type, etc.).  
+Example: GET http://localhost:6789/
+
+* Statistics route:  
+A GET **URL_PREFIX/stats/** route, shows running model statistics (memory).  
+Example: GET http://localhost:6789/
+
+* Capabilities route:  
+A GET **URL_PREFIX/capabilities/** route, shows payload formats supported by running model.
```
rvorobii, line 168: if we want users to be able to create their own custom environments - the description here should be very detailed.
e.g. request schema, response schema

yakov-g: But these are only DRUM endpoints. 
Maybe I just should mention they are not accessible on DR prediction server

README.md
yakov-g: /health is ok
But for stats, we need schema?

yakov-g: you can review again

rvorobii (commented):
the docs are _better_ now, but if user decided to create a custom execution environment - the user will be frustrated.
to power users for creating their own environments, IMO we need request/response schema + detailed description for each route we have.
at least for the minimal possible environment, where
- `/`
- `/predict`

are required.


I understand creating custom execution environments are an advanced topic, especially now when users on Cloud cannot upload their own environments, but if we want to cover this in docs - we should do it covering all the details, so that users **really** can implement custom environments.
as for example with such level of detail in docs, I won't be able to do this without looking into DRUM code

custom_model_runner/README.md
yakov-g: So what are you asking for regarding this paragraph?

yakov-g:
So could you point to some examples for such detailed description + schema?

custom_model_runner/README.md
meltzg: > description here should be very detailed. e.g. request schema, response schema

yakov-g: So not here. This is DRUM only documentation, that goes here: https://pypi.org/project/datarobot-drum/

And you are talking about creating your own custom env.

meltzg: no, i mean here. These routes are in drum server. they should be fully documented.

yakov-g: ok

README.md
```diff
@@ -304,11 +304,17 @@ If you'd like to use a tool/language/framework that is not supported by our temp
...
 2) Custom models require a simple webserver in order to make predictions. We recommend putting this in
 your environment so that you can reuse it with multiple models. The webserver must be listening on port 8080 and implement the following routes:
-    1) `GET /{URL_PREFIX}/` This route is used to check if your model's server is running.
-    2) `POST /URL_PREFIX/predict/` This route is used to make predictions.
+   > **Note: `URL_PREFIX` is an environment variable that will be available at runtime. It has to be read and pasted into the routes.**
+    1) Mandatory endpoints:
+        1) `GET /URL_PREFIX/` This route is used to check if your model's server is running.
+        2) `POST /URL_PREFIX/predict/` This route is used to make predictions.
+    2) Nice-to-have extensions endpoints:  
+        1) `GET /URL_PREFIX/stats/` This route is used to check if model is loaded and functioning properly.
```
rvorobii, line 312:
```suggestion
        1) `GET /URL_PREFIX/stats/` This route is used to fetch memory usage data for DataRobot Custom Model Testing.
```

rvorobii (approved):

meltzg (approved):

yakov-g:

yakov-g:

yakov-g:

yakov-g:

PR #313, opened 2021-04-08
[RAPTOR-4944] Training model and transform parameter support

jmbannon:

## Summary
This PR adds parameter support for Python training models and transforms via JSON file.

The next addition that will follow this PR is validation of parameter definitions in the model-metadata.yaml file. This validation is done already in DataRobot, but we should add local validation as well to speed up error propagation to the user.


custom_model_runner/datarobot_drum/drum/args_parser.py
```diff
@@ -485,6 +485,16 @@ def _reg_arg_sparse_colfile(*parsers):
...
             )
 
+    @staticmethod
+    def _reg_arg_parameter_file(*parsers):
+        for parser in parsers:
+            parser.add_argument(
+                ArgumentsOptions.PARAMETER_FILE,
+                default=None,
+                type=CMRunnerArgsRegistry._is_valid_file,
+                help="Model parameters stored in a JSON file",
```
4dahalibut, line 495: why not call em hyperparams?

model_templates/training/python3_sklearn_binary_hyperparameters/custom.py
```diff
@@ -0,0 +1,171 @@
...
+
+    if not parameters:
+        raise ValueError("Did not receive parameters")
+
+    # Parameters are provided during fit as a dict with names set according to your model-metadata.yaml file.
+    # In this example, we set the 'model' parameters in our sklearn pipeline.
+    estimator.set_params(
+        **{
+            "model__penalty": parameters["penalty"],
+            "model__dual": parameters["dual"],
```
4dahalibut, line 65: what would you think about just replacing the normal sklearn binary one with this?

tests/drum/test_fit.py
```diff
@@ -199,6 +203,59 @@ def test_fit(
...
+            (SKLEARN_BINARY_HYPERPARAMETERS, BINARY_TEXT, None, SKLEARN_BINARY_PARAMETERS),
+            (SKLEARN_BINARY_HYPERPARAMETERS, BINARY_SPACES, None, SKLEARN_BINARY_PARAMETERS),
+            (SKLEARN_TRANSFORM_HYPERPARAMETERS, REGRESSION, None, SKLEARN_TRANSFORM_PARAMETERS),
+        ],
+    )
+    @pytest.mark.parametrize("weights", [WEIGHTS_CSV, WEIGHTS_ARGS, None])
+    def test_fit_hyperparameters(
+        self, resources, framework, problem, docker, parameters, weights, tmp_path,
+    ):
+        if framework == RDS:
```
4dahalibut, line 218: Are you going to make an R template? 

Also, can you clean up this function if you're not going to use the various logic branches?

4dahalibut:
image

4dahalibut:
Also, do you think we should add a functional test?

jmbannon:
> Also, do you think we should add a functional test?

We should, once we add the code in the DR repo to support hyperparams. So not in this PR

custom_model_runner/datarobot_drum/drum/args_parser.py
jmbannon: These are the parameter values, not the hyperparam definitions

model_templates/training/python3_sklearn_binary_hyperparameters/custom.py
jmbannon: I'd be okay with that, but first lets get it working e2e in DR

tests/drum/test_fit.py
jmbannon: Eventually, and sure

custom_model_runner/datarobot_drum/drum/common.py
```diff
@@ -181,6 +182,7 @@ class ArgumentsOptions:
     CLASS_LABELS_FILE = "--class-labels-file"
     SKIP_DEPS_INSTALL = "--skip-deps-install"
     SPARSE_COLFILE = "--sparse-column-file"
+    PARAMETER_FILE = "--parameter-file"
```
yakov-g, line 185: You may want to add PARAMETER_FILE into `ArgumentOptionsEnvVars`, so it will be automatically taken by DRUM if you provide `PARAMETER_FILE` env var. 
It is useful in the env's `test_fit.sh` so you won't need to read env var and pass it into --parameter-file

yakov-g (approved):

custom_model_runner/datarobot_drum/drum/args_parser.py
4dahalibut: they're hyperparameter values though.

4dahalibut (approved):

custom_model_runner/datarobot_drum/drum/args_parser.py
jmbannon: Screen Shot 2021-04-09 at 10 42 27 AM
We call them parameters in DR so I think we should stick to that wording

4dahalibut: well uhh, we call them hyperparameters some places don't we? It's weird that its inconsistent.

4dahalibut: it shouldn't matter types vs values.

jmbannon: we call the definitions hyperparameters, which define min, max, values, etc

jmbannon:
Will address the comments in a follow-up

PR #326, opened 2021-05-03
[RAPTOR-5277,5301] validate classification predictions add up to one for all the predictors

yakov-g:
## Summary
Enable validation that  classification probabilities sum up to 1 during predict, into the common code in DRUM, so that validation works for all the predictors, on every predict.

### DIFF
--- a/custom_model_runner/CHANGELOG.md
+++ b/custom_model_runner/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+#### [1.5.6] - in progress
+##### Fixed
+- validate classification probabilities add up to one for all the predictors
+
 #### [1.5.5] - 2021-04-28
 ##### Updated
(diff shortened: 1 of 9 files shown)


yakov-g:
jarvis test this

yakov-g:

rvorobii (approved):
I think having the validation on every predict is better -- this will allow to always find the issue of not summing up to 1 (user may just not run or forget `drum validation`).
This will guarantee easy transition of models from DRUM to DataRobot.

Although the other option is also fine.

yakov-g:
From
@zachmayer 
(Note that we will have multilabel classification soon, where it is ok for predictions not to sum to 1)
For multiclass, I think it makes sense to raise an exception for predictions that do not sum to 1, with an option the user can set to disable that validation

meltzg (changes requested):
I need more convincing than `it's the responsibility of the user to validate` for allowing invalid predictions to be passed along as valid.

rvorobii:
@yakov-g as I said I'm for validating the sum on every predict.

also I said that the `drum validation` option is also fine -- but I don't like it at all now. most likely it'll lead to many (invalid) models that are incompatible with DR with this option.

yakov-g:
Yes, I'm enabling validation on every prediction

meltzg (approved):

PR #338, opened 2021-05-24
[RAPTOR-5369] Install DRUM 1.5.6.post1 into java env

yakov-g:

## Summary
I thought version `1.5.6.post1` will be automatically pulled when version is pinned to `datarobot-drum==1.5.6`, but not.

### DIFF
--- a/public_dropin_environments/java_codegen/dr_requirements.txt
+++ b/public_dropin_environments/java_codegen/dr_requirements.txt
@@ -1,2 +1,2 @@
 pyarrow==0.14.1
-datarobot-drum==1.5.6
+datarobot-drum==1.5.6.post1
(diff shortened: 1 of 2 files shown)


public_dropin_environments/java_codegen/dr_requirements.txt
```diff
@@ -1,2 +1,2 @@
 pyarrow==0.14.1
-datarobot-drum==1.5.6
+datarobot-drum==1.5.6.post1
```
rvorobii, line 2: why `1.5.6.post1` and not `1.5.7`?

yakov-g: Small change since the last release, which I only want to test on the Java env.
No need to update all the envs first on staging then on prod.

rvorobii (approved):

meltzg (approved):

PR #340, opened 2021-05-26
[YOLO] output total run time number per test case

yakov-g:

## Summary
Add total run time per test case, It can be useful , when comparing performance. It is easier to see difference in total time than in time per iteration.

size   samples   iters    min     avg     max    total   used (MB)   total physical (MB)
========================================================================================
3 KB        10     150   0.018   0.020   0.024   3.006     417.801             31442.730


zohar-mizrahi (approved):
LGTM

rvorobii (approved):

PR #341, opened 2021-05-26
[Yolo] bug fix in score.jl

timsetsfire:
a small bug was present in the score.jl language predictor.  If a MLJ artificat was provided with custom.jl file which defined load_hook and no score hook, then necessary libraries to complete scoring would not be loaded. 

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/language_predictors/julia_predictor/score.jl
+++ b/custom_model_runner/datarobot_drum/drum/language_predictors/julia_predictor/score.jl
@@ -74,9 +74,7 @@ end
 function load_serialized_model(model_dir)
 
     model = nothing
-    if defined_hooks["load_model"]
-        return Custom.load_model(model_dir)
-    end
+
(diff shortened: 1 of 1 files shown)


yakov-g (approved):

timsetsfire:
jenkins cmrunner_integration

timsetsfire:

4dahalibut:
this could have test coverage

4dahalibut:
did you manually test?

timsetsfire:
Yes.  sorry i didn't post more detail.  The main issue was that if a custom `load_model` was defined then a custom `score` must because the logic in the score.jl resulted in skipping the loading libraries necessary to use the default scoring method if a custom `load_model` was used.  I only caught this when i went with a slightly more complex julia model that required a custom load_model function but should not have required a custom score function.  

Test coverage might be a good idea -> maybe a test for custom.py with single hooks defined.  Not sure.

rvorobii (approved):

PR #347, opened 2021-05-27
[RAPTOR-5435] add dependbot configuration

scottp-ml:

## Summary
Enabling dependbot is one of the security requirements for custom models.

### DIFF
--- a/dependbot.yml
+++ b/dependbot.yml
@@ -3,13 +3,13 @@ updates:
   - package-ecosystem: "pip"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "monthly"
     reviewers:
       - "datarobot/custom-models"
   - package-ecosystem: "pip"
     directory: "/custom_model_runner/"
     schedule:
-      interval: "daily"
+      interval: "monthly"
     reviewers:
       - "datarobot/custom-models"
   # check the dropin environments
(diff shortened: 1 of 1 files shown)


yakov-g:
I'm afraid we'll be overflown with deps update PRs.
Does it take care of any requirements files? We have `requirements.txt`, `dr_requirements.txt`, `requirements_dev.txt` etc.
There is also an image we build for tests https://github.com/datarobot/datarobot-user-models/tree/master/tests/fixtures/cmrun_docker_env, should be replaces with the python sklearn env though.

4dahalibut:
Can you leave comments about how you tested this?

scottp-ml:
@yakov-g yes, it will pick up all of the different requirements files.  For pip it just looks at everything that is .txt  And it will probably be an annoying number of PRs, we can drop the frequency of updating if that seems appropriate.

@4dahalibut I used the provided validation tool, but there doesn't seem to be a good way to check otherwise since it is turning on automation within github.

4dahalibut (approved):

yakov-g:
So the thing is, that our implicit policy is not to bump up versions.
Our public envs establish some baseline and if customer needs different versions, he can configure it using `requirements.txt` in a model.
We don't know if bumping versions may cause any issues for customers.

On the other hand - we have default dependabot enabled, which notifies us that update is required when there is a vulnerability.

So i'm not sure we want to get periodical notifications on all the packages.
Ping: @klichukb @meltzg, what do you think.

scottp-ml:
This is required for public beta of custom tasks, according to infosec assessment.  "Verify that all components are up to date, preferably using a dependency checker during build or compile time. ([C2](https://www.owasp.org/index.php/OWASP_Proactive_Controls#tab=Formal_Numbering))"  They indicated that for beta we just need dependabot, later on we will need more checks.  @yakov-g @meltzg @klichukb

yakov-g:
So, i tried to bump tensorflow version, which required bumping some other packages which led to moderate time consumption testing process.
I don't mind on the deps checker, just remind that this process may be time consuming and may require example models changes. Also CFDSs should be updated on the env packages changes.

Thus, I'd change bot to check monthly.

@meltzg @klichukb

scottp-ml:
@namrop  ^^^

meltzg (approved):

PR #360, opened 2021-06-16
[RAPTOR-5590] Load custom.r if custom.R fails

jmbannon:

## Summary
We allow `.r` and `.R` files everywhere except, we never try loading `custom.r` hooks anywhere.
This PR tries to load `Custom.R`, and if that fails, load `custom.r`


yakov-g (approved):

yakov-g:
I think integration tests run almost for 90 minutes, which is limit. I think it is required to be split

4dahalibut (approved):

PR #362, opened 2021-06-16
[RAPTOR-5492]deal with ugly list representation in f-string

scottp-ml:

## Summary
Get the correct string representation from the elements of a list of enums.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/typeschema_validation.py
+++ b/custom_model_runner/datarobot_drum/drum/typeschema_validation.py
@@ -260,7 +260,7 @@ class DataTypes(BaseValidator):
         )
         types[Values.DATE] = dataframe.select_dtypes("datetime").shape[1] > 0
 
-        types_present = [str(k) for k, v in types.items() if v]
+        types_present = [k for k, v in types.items() if v]
 
         base_error = f"Datatypes incorrect. Data has types: {DataTypes.list_str(types_present)}"
 
--- a/custom_model_runner/datarobot_drum/drum/typeschema_validation.py
+++ b/custom_model_runner/datarobot_drum/drum/typeschema_validation.py
@@ -20,13 +20,6 @@ logger = logging.getLogger("drum." + __name__)
 T = TypeVar("T")
 
 
-def list_str(l):
-    """f-strings do not do a great job dealing with lists of objects.  The __str__ method isn't called on the
-    contained objects, and the result is in [].  This provides the nicely formatted representation we want
-    in the error message"""
-    return ", ".join([str(x) for x in l])
-
-
 class BaseEnum(PythonNativeEnum):
     def __str__(self) -> str:
         return self.name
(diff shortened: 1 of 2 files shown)


custom_model_runner/datarobot_drum/drum/typeschema_validation.py
```diff
@@ -20,6 +20,13 @@
 T = TypeVar("T")
 
 
+def list_str(l):
```
eric-s-s, line 23: as this is only used in the one function, why not just move it there and get rid of the docs.

also, just return `[str(el) for el in input_list]`

custom_model_runner/datarobot_drum/drum/typeschema_validation.py
```diff
@@ -253,15 +260,15 @@ def validate(self, dataframe):
         )
         types[Values.DATE] = dataframe.select_dtypes("datetime").shape[1] > 0
 
-        types_present = [k for k, v in types.items() if v]
+        types_present = [str(k) for k, v in types.items() if v]
```
eric-s-s, line 263: this definitely breaks things.  you'll have use your function on L265

eric-s-s (changes requested):
yikes!  see https://engprod.drdev.io/api/pr_redirect/datarobot-user-models/363 for how things got broken.  we no longer PR test the unit tests, because they need to be in `tests/drum/unit` (and i don't know if that will work) because if we did, this would fail.  that PR should have a solution

custom_model_runner/datarobot_drum/drum/typeschema_validation.py
scottp-ml: that would still have the [] in the output, which we need to get rid of.

eric-s-s:
@scottp-ml see: https://github.com/datarobot/datarobot-user-models/pull/364  you will want to wait for that and then rebase.

eric-s-s:
also, since you have a very specific want for this error message, please include a test showing that it does what you want.  probably in  `tests.drum.unit.test_typeschema_validation.TestSchemaValidator`  (once the other PR merges.  it's the wrong location here)

tests/drum/unit/test_typeschema_validation.py
```diff
@@ -692,3 +693,16 @@ def test_num_col_values(condition, value, fails):
             NumColumns(condition, value)
     else:
         NumColumns(condition, value)
+
+
+def test_datatypes_error_formatting():
```
eric-s-s, line 698: this should really test at a higher level and in the `TestSchemaValidator`

here's a copypaste test i took from: `tests.drum.unit.test_typeschema_validation.TestSchemaValidator.test_data_types`

```
    def test_data_types_error_message(
        self, ten_k_diabetes
    ):
        condition = Conditions.IN
        values = [Values.CAT, Values.NUM]
        target = "readmitted"

        yaml_str = input_requirements_yaml(Fields.DATA_TYPES, condition, values)
        schema_dict = self.yaml_str_to_schema_dict(yaml_str)
        validator = SchemaValidator(schema_dict)
        ten_k_diabetes.drop(target, inplace=True, axis=1)

        match_str = r'has types:( \w+,?){2,3}.* exactly match: CAT, NUM'
        with pytest.raises(DrumSchemaValidationException, match=match_str):
            validator.validate_inputs(ten_k_diabetes)
```

eric-s-s (approved):
i have nits.  they are opinions and will not hold up this PR one way or the other. otherwise good.

jmbannon (approved):

PR #363, opened 2021-06-16
[YOLO] fix enum labeling

eric-s-s:
### Summary

this is a test of @scottp-ml 's PR to see if we can get it failing.  it fails locally

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/typeschema_validation.py
+++ b/custom_model_runner/datarobot_drum/drum/typeschema_validation.py
@@ -1,7 +1,7 @@
 from abc import ABC, abstractmethod
 import base64
 import logging
-from enum import auto, Enum
+from enum import auto
 from enum import Enum as PythonNativeEnum
 from io import BytesIO
 import operator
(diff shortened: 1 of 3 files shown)


custom_model_runner/datarobot_drum/drum/typeschema_validation.py
```diff
@@ -20,6 +20,13 @@
 T = TypeVar("T")
 
 
+def list_str(l):
```
eric-s-s, line 23: remove this

custom_model_runner/datarobot_drum/drum/typeschema_validation.py
```diff
@@ -253,15 +260,18 @@ def validate(self, dataframe):
         )
         types[Values.DATE] = dataframe.select_dtypes("datetime").shape[1] > 0
 
-        types_present = [k for k, v in types.items() if v]
+        types_present = [str(k) for k, v in types.items() if v]
+
+        def list_of_strings(input_list):
+            return [str(el) for el in input_list]
```
scottp-ml, line 266: this still should be ', '.join(...) to get printed the way we want for the message using f-string

scottp-ml (approved):
LGTM

PR #365, opened 2021-06-16
[YOLO] Try xdist out for test performance

carsongee:

## Summary
 Attempt to speed up integration tests using xdist

## Rationale
🏃

### DIFF
--- a/custom_model_runner/datarobot_drum/resource/drum_server_utils.py
+++ b/custom_model_runner/datarobot_drum/resource/drum_server_utils.py
@@ -25,9 +25,12 @@ def _wait_for_server(url, timeout, process_holder):
         if timeout <= 0:
             if process_holder is not None:
                 print("Killing subprocess: {}".format(process_holder.process.pid))
-                os.killpg(os.getpgid(process_holder.process.pid), signal.SIGTERM)
-                time.sleep(0.25)
-                os.killpg(os.getpgid(process_holder.process.pid), signal.SIGKILL)
+                try:
+                    os.killpg(os.getpgid(process_holder.process.pid), signal.SIGTERM)
+                    time.sleep(0.25)
+                    os.killpg(os.getpgid(process_holder.process.pid), signal.SIGKILL)
+                except psutil.ProcessLookupError:
+                    assert False, "Server failed to start: url: {}".format(url)
 
             assert timeout, "Server failed to start: url: {}".format(url)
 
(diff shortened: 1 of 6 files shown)


yakov-g:

yakov-g:
@eric-s-s 
I'm trying to start running pytest in parallel to decrease test time.
I had to comment out some of the test cases, because of the way parametrization is done.
Could you tell me what would be the best approach to change it? Just define lists statically?

You can try to uncomment a test case and run it locally using: `pytest ./tests/drum/ -n auto`

tests/conftest.py
```diff
@@ -578,3 +584,57 @@ def variety_resources(
     resource.target = get_variety_target
     resource.class_labels = get_variety_classes_labels
     return resource
+
+
+@pytest.fixture
```
eric-s-s, line 589: is there a need to move these?  if not, it would save time for the reader to leave them closer to where the tests are.  maybe in `tests/drum/unit/conftest.py` if you want to move them out of the test itself?

eric-s-s:
> @eric-s-s
> I'm trying to start running pytest in parallel to decrease test time.
> I had to comment out some of the test cases, because of the way parametrization is done.
> Could you tell me what would be the best approach to change it? Just define lists statically?
> 
> You can try to uncomment a test case and run it locally using: `pytest ./tests/drum/ -n auto`, you'll see it fails to collect tests.
> 
> just install pytest-xdist

i have no idea.  my first go-to solution would be to make them global constants right above the test they're used in:  have you tried this?

```python


NUMERIC_CONDITIONS = list(set(Conditions) - set(Conditions.non_numeric()))
# if it needs to be more deterministic
NUMERIC_CONDITIONS = sorted(set(Conditions) - set(Conditions.non_numeric()), key=lambda el: el.name)


class TestSomething
    @pytest.mark.parametrize("condition", NUMERIC_CONDITIONS)
    def test_datatypes_unallowed_conditions(self, condition):
        values = [Values.NUM, Values.TXT]
        input_data_type_str = input_requirements_yaml(self.field, condition, values)


```

also, another option is to run unit tests in series.  they take less than 1 second to run

yakov-g:

tests/conftest.py
yakov-g: moved fixtures back

yakov-g:

custom_model_runner/datarobot_drum/drum/typeschema_validation.py
```diff
@@ -72,6 +72,27 @@ def single_value_conditions(cls) -> List["Conditions"]:
             cls.NOT_LESS_THAN,
         ]
 
+    @classmethod
```
eric-s-s, line 75: these added class methods aren't really useful in any way, they are _only_ for _a_ uniti test suite.  can we just make them constants within the unit tests, please.  having them here, will just confuse any future users. 

also, this isn't numeric conditions, it's `ONLY_FOR_NUMERIC`

custom_model_runner/datarobot_drum/drum/typeschema_validation.py
```diff
@@ -103,6 +124,20 @@ def input_values(cls) -> List["Values"]:
     def output_values(cls) -> List["Values"]:
         return [cls.NEVER, cls.DYNAMIC, cls.ALWAYS, cls.IDENTITY]
 
+    @classmethod
```
eric-s-s, line 127: same comment as above, please move to the tests which are the only place they're used

custom_model_runner/datarobot_drum/drum/typeschema_validation.py
eric-s-s: if they need to be imported, maybe make a constants file in the unit test directory?

yakov-g: DONE

eric-s-s (approved):
LGTM

yakov-g (approved):

PR #366, opened 2021-06-17
[RAPTOR-5607 | HOTFIX] Bandaid issue when using -cd and --docker

jmbannon:

## Summary
When using `-cd` instead of `--code-dir` with `--docker`, `drum fit` can fail within docker.
Using `-cd` works locally. I've spent a few hours trying to figure out what is wrong with no avail. This hotfix will at least make `-cd` work from an outsider perspective, but the underlying issue still remains.

When using -cd, custom.py is not found within docker:
Screenshot from 2021-06-17 11-57-32

## Changes
Replace `-cd` with `--code-dir` when constructing the docker command


eric-s-s:
i _think_ this is bad directory shenanigans:

custom_model_runner/datarobot_drum/drum/drum.pyL798

```
        CMRunnerUtils.replace_cmd_argument_value(in_docker_cmd_list, "-cd", in_docker_model)
```

and what that does is:

replace `"-cd"` in the command list with: `"/opt/model"`  _but it doesn't do this for --code-dir_

eric-s-s:
nope! i was wrong.  the exact same thing is right above it for code-dir:
```
        CMRunnerUtils.replace_cmd_argument_value(
            in_docker_cmd_list, ArgumentsOptions.CODE_DIR, in_docker_model
        )
```

eric-s-s:
i suspect that passing two arguments like this is an incorrect usage:
```
            parser.add_argument(
                "-cd",
                ArgumentsOptions.CODE_DIR,
                default=None,
                required=True,
                type=type_callback,
                help=help_message,
            )
```

and that replacing the command in docker is failing silently here:
datarobot_drum.drum.utils.CMRunnerUtils.replace_cmd_argument_value
```
    def replace_cmd_argument_value(cls, cmd_list, arg_name, arg_value):
        try:
            ind = cmd_list.index(arg_name)
        except ValueError:
            return
        cmd_list[ind + 1] = arg_value
```
but who knows.

eric-s-s (approved):
:shrug: if it works it works.

yakov-g:
maybe let's just remove support for `-cd` ?

jmbannon:
> maybe let's just remove support for `-cd` ?

I'd be okay with that, no other args have short versions

yakov-g:
> > maybe let's just remove support for `-cd` ?
> 
> I'd be okay with that, no other args have short versions

Cool,
Are you going to do this?

jmbannon:
@yakov-g New PR is here: https://github.com/datarobot/datarobot-user-models/pull/368

jmbannon:

custom_model_runner/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Show full R traceback for fit/predict errors
 ##### Fixed
 - Loading hooks from both custom.R and custom.r
+- Using -cd instead of --code-dir failed sometimes when running with docker
```
4dahalibut, line 12: maybe better to be more official here. sometimes -> occasionally

jmbannon: Will fix in the release PR. Should remove 'sometimes' because it happens every time

scottp-ml (approved):

PR #367, opened 2021-06-17
[RAPTOR-5588] Fix R fit bug when handling single-col data

jmbannon:

## Summary
If R was passed a dataframe that contained only a target and single column of data, it would coerce X to be a 1-d array which caused problems (the code assumes we're handling a 2-d matrix). This PR ensures the matrix structure isn't dropped by adding DROP=False to all manipulations of X


4dahalibut (approved):

tests/conftest.py
```diff
@@ -179,6 +182,7 @@
     BINARY_TEXT: "binary",
     BINARY_SPACES: "binary",
     REGRESSION: "regression",
+    REGRESSION_SINGLE_COL: "regression",
```
4dahalibut, line 185: why not just use the regression key here...

jmbannon:

jmbannon:

jmbannon:
jarvis test cmrunniner_integration

yakov-g (approved):

yakov-g:

jmbannon:

jmbannon:

PR #377, opened 2021-07-13
[RAPTOR-5792] Check R colname validation as both literals and doubles

jmbannon:

## Summary
Class labels here refers to the output prediction dataframe's column names, not the actual data. 

DR will sometimes send class labels in R as `c("0.0", "1.0")` or `c("0", "1")`. We compared class labels' literal form, so if DR expected `"1.0"` and the custom code provided `1.0`, it would fail.

This PR adds functionality to compare class labels as both literals and casted as doubles. It only casts to doubles if all class labels are casted successfully. If it does match as doubles, then set the column names to the expected version.


scottp-ml (approved):
LGTM

jmbannon:

meltzg (approved):

PR #380, opened 2021-07-19
[RAPTOR-5781] update generic_predictor.py

timsetsfire:
Tied to bug report https://datarobot.atlassian.net/browse/RAPTOR-5781

Transform hook can return either a tuple or a dataframe -> updated generic_predictor.py accordingly.  

### DIFF
--- a/custom_model_runner/datarobot_drum/resource/components/Python/generic_predictor/generic_predictor.py
+++ b/custom_model_runner/datarobot_drum/resource/components/Python/generic_predictor/generic_predictor.py
@@ -106,11 +106,13 @@ class GenericPredictorComponent(ConnectableComponent):
             binary_data, mimetype = StructuredInputReadUtils.read_structured_input_file_as_binary(
                 input_filename
             )
-            transformed_output = self._predictor.transform(binary_data=binary_data, mimetype=mimetype)
+            transformed_output = self._predictor.transform(
+                binary_data=binary_data, mimetype=mimetype
+            )
             if type(transformed_output) == tuple:
                 transformed_df = transformed_output[0]
             elif type(transformed_output) == DataFrame:
-                transfromed_df = transformed_output   
+                transfromed_df = transformed_output
             transformed_df.to_csv(output_filename, index=False)
         else:
             binary_data, mimetype = StructuredInputReadUtils.read_structured_input_file_as_binary(
--- a/custom_model_runner/datarobot_drum/resource/components/Python/generic_predictor/generic_predictor.py
+++ b/custom_model_runner/datarobot_drum/resource/components/Python/generic_predictor/generic_predictor.py
@@ -109,9 +109,9 @@ class GenericPredictorComponent(ConnectableComponent):
             transformed_output = self._predictor.transform(
                 binary_data=binary_data, mimetype=mimetype
             )
-            if type(transformed_output) == tuple:
+            if isinstance(transformed_output, tuple):
                 transformed_df = transformed_output[0]
-            elif type(transformed_output) == DataFrame:
+            elif isinstance(transformed_output, DataFrame):
                 transfromed_df = transformed_output
             transformed_df.to_csv(output_filename, index=False)
         else:
--- a/custom_model_runner/datarobot_drum/resource/components/Python/generic_predictor/generic_predictor.py
+++ b/custom_model_runner/datarobot_drum/resource/components/Python/generic_predictor/generic_predictor.py
@@ -109,10 +109,7 @@ class GenericPredictorComponent(ConnectableComponent):
             transformed_output = self._predictor.transform(
                 binary_data=binary_data, mimetype=mimetype
             )
-            if isinstance(transformed_output, tuple):
-                transformed_df = transformed_output[0]
-            elif isinstance(transformed_output, DataFrame):
-                transfromed_df = transformed_output
+            transformed_df = transformed_output[0]
             transformed_df.to_csv(output_filename, index=False)
         else:
             binary_data, mimetype = StructuredInputReadUtils.read_structured_input_file_as_binary(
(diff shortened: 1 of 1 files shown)


custom_model_runner/datarobot_drum/resource/components/Python/generic_predictor/generic_predictor.py
```diff
@@ -105,7 +106,13 @@ def _materialize(self, parent_data_objs, user_data):
             binary_data, mimetype = StructuredInputReadUtils.read_structured_input_file_as_binary(
                 input_filename
             )
-            transformed_df = self._predictor.transform(binary_data=binary_data, mimetype=mimetype)
+            transformed_output = self._predictor.transform(
```
rvorobii, line 109: can we make `transform` always return either dataframe or tuple?
what is the reason it's sometimes dataframe and sometimes tuple

jmbannon: its dependent on sending `y` to the transform. I agree though, we should just always make it a tuple

jmbannon (approved):

custom_model_runner/datarobot_drum/resource/components/Python/generic_predictor/generic_predictor.py
timsetsfire: so i'm curious on how to move forward with this.  To make it always return a tuple would require changes to model_adapter.py and we could essentially undo everything that was done to generic_predictor.  

part of model_adapter to change 👇 
```
        if self._custom_hooks.get(CustomHooks.TRANSFORM):
            try:
                transform_hook = self._custom_hooks[CustomHooks.TRANSFORM]
                transform_params = signature(transform_hook).parameters
                if len(transform_params) == 3:
                    transform_out = transform_hook(data, model, target_data)
                elif len(transform_params) == 2:
                    transform_out = transform_hook(data, model)
                else:
                    raise ValueError(
                        "Transform hook must take 2 or 3 arguments; "
                        "hook provided takes {}".format(len(transform_params))
                    )
                if type(transform_out) == tuple:
                    output_data, output_target = transform_out
                else:
                    output_data = transform_out
                    output_target = target_data

```
@yakov-g thoughts??

yakov-g: So, as I understand, this solution won't work as `transform` which is defined here,
https://github.com/datarobot/datarobot-user-models/blob/a91996cc2674162424924cffe6e4de97e75a03e6/custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py#L123-L128
will always return tuple, returned all the way from model_adapter (even if the hook returns a single value). So we already always return tuple. 
Thus, looks like, all that has to be done is adding `[0]`: 
`transformed_df = self._predictor.transform(binary_data=binary_data, mimetype=mimetype)[0]`

rvorobii (commented):
I'm confused -- still don't understand if always returning a tuple is doable (which to me seems as a proper solution).
@jmbannon @yakov-g you're more familiar with this code, leaving this to you guys to find a proper solution.

custom_model_runner/datarobot_drum/resource/components/Python/generic_predictor/generic_predictor.py
```diff
@@ -105,7 +106,13 @@ def _materialize(self, parent_data_objs, user_data):
...
+                binary_data=binary_data, mimetype=mimetype
+            )
+            if type(transformed_output) == tuple:
+                transformed_df = transformed_output[0]
+            elif type(transformed_output) == DataFrame:
+                transfromed_df = transformed_output
```
meltzg, line 115:
```suggestion
            if isinstance(transformed_output, tuple):
                transformed_df = transformed_output[0]
            elif isinstance(transformed_output, DataFrame):
                transfromed_df = transformed_output
```

meltzg (changes requested):
could you add a test?

custom_model_runner/datarobot_drum/resource/components/Python/generic_predictor/generic_predictor.py
timsetsfire: changed.  

need to figure out what to test here.  @yakov-g any thoughts?

yakov-g: The test could be to check cases when transformed_output is either tuple or DataFrame, but, as I said before, looks like PythonPredictor always returns tuple. And transform is not implemented for other languages. 

So I cant think of any new test cases.

custom_model_runner/datarobot_drum/resource/components/Python/generic_predictor/generic_predictor.py
```diff
@@ -105,7 +106,13 @@ def _materialize(self, parent_data_objs, user_data):
             binary_data, mimetype = StructuredInputReadUtils.read_structured_input_file_as_binary(
                 input_filename
             )
-            transformed_df = self._predictor.transform(binary_data=binary_data, mimetype=mimetype)
+            transformed_output = self._predictor.transform(
+                binary_data=binary_data, mimetype=mimetype
+            )
+            if isinstance(transformed_output, tuple):
```
yakov-g, line 112: So just once again: do you agree that `transformed_output` is always a tuple? Why then we check for DataFrame type?

timsetsfire: @yakov-g finally address this and agree output is always a tuple and eliminated hanlding of dataframe.

yakov-g (approved):

custom_model_runner/datarobot_drum/resource/components/Python/generic_predictor/generic_predictor.py
timsetsfire: resolve -> remove checks.  I agree that the PythonPredictor always returns a tuple.

meltzg (approved):

PR #385, opened 2021-07-27
[RAPTOR-5868][RAPTOR-5911] Transforms: Use default type schema when not provided && Fix strict validation flag

jmbannon:

## Summary
- Enables type schema for transforms when it is not included in the model metadata, and when strict validation is enabled
- Fix the strict validation argument (operated before as `--enable-strict-validation`)
  - This will enable strict validation in all future drop-in environments


custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -80,13 +80,27 @@ def __init__(self, runtime):
...
+            # always populate the validator, even if info isn't provided. Use the default type schema if no
+            # schema is provided, strict validation is enabled, and the target type is transform
+            type_schema = {}
+            use_default_type_schema = False
+            strict_validation = not self.options.disable_strict_validation
+            if self.options.model_config:
+                type_schema = self.options.model_config.get("typeSchema", {})
+
+            if not type_schema and strict_validation and self.target_type == TargetType.TRANSFORM:
+                print(
```
jmbannon, line 92: decided to not put this behind the verbose flag since it's seems important to convey to the user

4dahalibut: maybe print to stderr? Also, can you add something which explains what the default is, or where to find this info?

4dahalibut: should we be using logger.warning instead? @yakov-g

custom_model_runner/datarobot_drum/drum/typeschema_validation.py
```diff
@@ -449,13 +450,45 @@ class SchemaValidator:
     actual validation on the respective dataframes.
     """
 
-    def __init__(self, type_schema: dict, strict=True, verbose=False):
+    _DEFAULT_TYPE_SCHEMA_CODEDIR_PATH = os.path.abspath(
+        os.path.join(os.path.dirname(__file__), "..", "resource", "default_typeschema")
+    )
```
4dahalibut, line 455: There may be weirdness building this into a wheelfile.

custom_model_runner/datarobot_drum/resource/default_typeschema/model-metadata.yaml
```diff
@@ -0,0 +1,30 @@
+name: default-transform-model-metadata
```
4dahalibut, line 1: Delete the above init.py probably.

4dahalibut: I see what you're saying about this being a weird thing to put in the drum repo though.

custom_model_runner/datarobot_drum/drum/drum.py
jmbannon: logger does not go to stdout. It has to be print/stdout because that's what gets printed to the user

jmbannon: In this file, all user-messages are prints

custom_model_runner/datarobot_drum/drum/typeschema_validation.py
jmbannon: There isn't. This is how other resource files are referenced

custom_model_runner/datarobot_drum/resource/default_typeschema/model-metadata.yaml
jmbannon: Nope, need it to include it in the wheel

4dahalibut (approved):
Approved given you manually tested locally and in DR, and you'll write functional tests for this.

custom_model_runner/datarobot_drum/drum/drum.py
jmbannon: And for this info, we should put it in the README for transforms

4dahalibut: are you sure logger doesn't go to stdout? where does it go?

4dahalibut: and we should be printing to stderr from inside drum, not stdout.

yakov-g: > are you sure logger doesn't go to stdout? where does it go?
I think Jesse means logger goes to whatever it is configured to go, so may not go to stdout.
Currently it is not configured, so by default looks like it is stderr.

4dahalibut: ok lets use logger then here pls.

jmbannon: it won't be shown to the user if it's in logger. Locally, `info`s do not go to stdout

jmbannon: If we plan to change logger to go to stdout then we need to evaluate everything we write in there and make sure it looks user friendly and not double log (in a few places we log + print). So until then, the print needs to stay

4dahalibut: kk...

yakov-g (approved):

PR #391, opened 2021-08-05
[RAPTOR-5739] Applying Jesse's technique to python

4dahalibut:

## Summary
In  https://github.com/datarobot/datarobot-user-models/pull/377 we started checking if values could be converted into floats, and sorting them by their float-equivalent values if they could. Here I use the same technique in python.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/artifact_predictors/sklearn_predictor.py
+++ b/custom_model_runner/datarobot_drum/drum/artifact_predictors/sklearn_predictor.py
@@ -9,6 +9,7 @@ from datarobot_drum.drum.common import (
     extra_deps,
     SupportedFrameworks,
 )
+from datarobot_drum.drum.utils import marshal_labels
 from datarobot_drum.drum.exceptions import DrumCommonException
 from datarobot_drum.drum.artifact_predictors.artifact_predictor import ArtifactPredictor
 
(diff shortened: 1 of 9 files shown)


tests/drum/test_units.py
```diff
@@ -700,3 +700,26 @@ def test_run_java_server_entry_point_succeed(self):
         # required to properly shutdown py4j Gateway
         pred._setup_py4j_client_connection()
         pred._stop_py4j()
+
+
+@pytest.mark.parametrize("data_dtype,label_dtype", [(int, int), (float, int)])
```
meltzg, line 705: how about `(int, float)`?

custom_model_runner/datarobot_drum/drum/utils.py
```diff
@@ -304,3 +304,11 @@ def capture_R_traceback_if_errors(r_handler, logger):
                 "t_traceback_err": traceback_exc,
             }
         raise
+
+
+def stringify_and_intify_label(label):
+    if isinstance(label, float):
+        if hash(label) != hash(int(label)):
+            raise ValueError("Class labels must be integers or strings")
```
meltzg, line 312: why _must_ labels be ints or strings? I know it's likely not common, but what's the issue with supporting labels like `1.1/2.2`?

rvorobii (approved):
other than the points Greg has -- looks good

tests/drum/test_units.py
4dahalibut: so we're going to just not allow support float labels.

custom_model_runner/datarobot_drum/drum/utils.py
4dahalibut: because I don't like them and they're bad

jmbannon: I don't think we should restrict this. I could see some 'regression' datasets with a few unique target values that someone might want to try as multiclass. We support it in DR so we should support it in training models

jmbannon (changes requested):
We should allow floats

custom_model_runner/datarobot_drum/drum/artifact_predictors/sklearn_predictor.py
```diff
@@ -59,14 +60,9 @@ def predict(self, data, model, **kwargs):
...
-                    str(label) for label in self.class_labels
-                ):
-                    error_message = "Wrong class labels {}. Use class labels detected by sklearn model: {}".format(
-                        self.class_labels, model.classes_
-                    )
-                    raise DrumCommonException(error_message)
-                labels_to_use = model.classes_
+                labels_to_use = marshal_labels(
+                    expected_labels=list(self.class_labels), actual_labels=list(model.classes_)
+                )
```
jmbannon, line 65: I don't think we need to change labels_to_use here. This only covers sklearn models. The check/fix should be downstream from this

jmbannon (approved):

PR #400, opened 2021-08-13
[RAPTOR-5822] add schema for remaining templates

scottp-ml:

## Summary
Part 2.  Add typeschema to tasks and estimator templates.


task_templates/estimators/python_anomaly/model-metadata.yaml
```diff
@@ -0,0 +1,25 @@
+name: python_anomaly
+type: training
+targetType: anomaly
+# modelID: optional
+environmentID: 5e8c889607389fe0f466c72d
+#trainingModel:
+#  trainOnProject: optional project on which to train
+validation:
+  # Path is relative to this file
+  input: ../../../tests/testdata/boston_housing.csv
```
jmbannon, line 10: doesn't this have text? We should make sure these validation ones work

task_templates/estimators/r_binary_classification/model-metadata.yaml
```diff
@@ -0,0 +1,25 @@
+name: r_binary_classification
+type: training
+targetType: binary
+# modelID: optional
+environmentID: 5e8c889607389fe0f466c72d
```
jmbannon, line 5: this needs to match the recent R public env ID

task_templates/estimators/r_regression/model-metadata.yaml
```diff
@@ -0,0 +1,25 @@
+name: r_regression
+type: training
+targetType: regression
+# modelID: optional
+environmentID: 5e8c889607389fe0f466c72d
```
jmbannon, line 5: R env id

jmbannon (changes requested):
Lets set the validation input to whatever the functional test uses for each template. And make sure R uses the R env ID

scottp-ml:
@jmbannon The issue with setting them to accept the input of the functional tests is that these were written to ignore columns that aren't useful.  So if you add text to them, that means they would be expected to work if they only received text, which isn't true.

task_templates/transforms/custom_class_python/model-metadata.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - field: number_of_columns
+      condition: NOT_LESS_THAN
+      value: 1
+    - field: sparse
+      condition: EQUALS
+      value: FORBIDDEN
+    - field: contains_missing
+      condition: EQUALS
+      value: SUPPORTED
+  output_requirements:
```
MikhailYakubovskiy, line 26: Scott, we should also add that this task never outputs missing values

task_templates/estimators/r_binary_classification/model-metadata.yaml
```diff
@@ -0,0 +1,25 @@
+name: r_binary_classification
+type: training
+targetType: binary
+# modelID: optional
+environmentID: 610305345da4263681a176b5
```
jmbannon, line 5: On staging the R environmentID is `5ea850ca1d41c8173c2feef6`

jmbannon: https://staging.datarobot.com/model-registry/custom-environments/5ea850ca1d41c8173c2feef6

PR #403, opened 2021-08-17
[YOLO] remove unnecessary ls in dockerfile

scottp-ml:

## Summary
Remove an ls from the dockerfile used to run tests.  The ls runs into permission issues in the jenkins environment, however access to files in /tmp is ok.  This allows the build to proceed correctly so that tests can be run.

### DIFF
--- a/tests/fixtures/cmrun_docker_env/Dockerfile
+++ b/tests/fixtures/cmrun_docker_env/Dockerfile
@@ -26,9 +26,9 @@ RUN pip3 install -r drum_requirements.txt --no-cache-dir && \
     rm -rf drum_requirements.txt
 
 # Copying the fresh wheel file
+#RUN mkdir tmp
 COPY datarobot_drum-*.whl /tmp/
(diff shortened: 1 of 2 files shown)


4dahalibut (approved):

tests/fixtures/cmrun_docker_env/Dockerfile
```diff
@@ -26,9 +26,9 @@ RUN pip3 install -r drum_requirements.txt --no-cache-dir && \
     rm -rf drum_requirements.txt
 
 # Copying the fresh wheel file
+#RUN mkdir tmp
```
meltzg, line 29: nuke me?

meltzg (approved):
other than removing commented line, looks fine. Mysterious, but fine.

tests/fixtures/cmrun_docker_env/Dockerfile
scottp-ml: fixed

PR #405, opened 2021-08-18
[RAPTOR-6092]Image transform example

scottp-ml:

## Summary
This adds an example transform with utilities to handle the input and output of image data.  DataRobot requires image data to be base 64 encoded string values, which might otherwise not be obvious to users.  This example has the utilities necessary to easily handle loading and saving the images in the expected format


custom_model_runner/datarobot_drum/drum/typeschema_validation.py
```diff
@@ -242,7 +242,11 @@ def is_text(x):
         -------
         boolean: True for is text, False for not text
         """
-        if pd.api.types.is_string_dtype(x) and pd.api.types.infer_dtype(x) != "boolean":
+        if (
+            pd.api.types.is_string_dtype(x)
+            and pd.api.types.infer_dtype(x) != "boolean"
+            and pd.api.types.infer_dtype(x) != "bytes"
```
4dahalibut, line 248: looks good!

4dahalibut (changes requested):
Please uncomment the tests. Otherwise lgtm.

task_templates/transforms/python3_image_transform/custom.py
```diff
@@ -0,0 +1,69 @@
...
+        so that the trained object can be used during scoring inside transform()
+    """
+
+    # dump the transform function
+    # into an artifact [in this example - artifact.pkl]
+    # and save it into output_dir so that it can be used to impute on new data
+    output_dir_path = Path(output_dir)
+    if output_dir_path.exists() and output_dir_path.is_dir():
+        with open("{}/artifact.pkl".format(output_dir), "wb") as fp:
+            pickle.dump(_process_image, fp)
```
4dahalibut, line 46: Actually, I think for this I'd rather see nothing get pickled, and we can then only have stuff in the transform. I think that would be easier for people to understand.

task_templates/transforms/python3_image_transform/custom.py
```diff
@@ -0,0 +1,69 @@
...
+from pathlib import Path
+
+from img_utils import (
+    b64_to_img,
+    img_to_b64,
+    img_to_grayscale,
+)
+
+
+def _process_image(raw_data):
```
4dahalibut, line 12: can you add type annotations here, and in img_utils? Might help with readability.

task_templates/transforms/python3_image_transform/model-metadata.yaml
```diff
@@ -0,0 +1,35 @@
...
+  output_requirements:
+    - field: data_types
+      condition: EQUALS
+      value: IMG
+    - field: sparse
+      condition: EQUALS
+      value: NEVER
+    - field: contains_missing
+      condition: EQUALS
+      value: NEVER
```
4dahalibut, line 35: Are these the defaults? Should we have them in the metadata?

task_templates/transforms/python3_image_transform/custom.py
scottp-ml: except the model adapter expects an artifact and fails if there isn't one.  This is also the pattern used in all the other examples

task_templates/transforms/python3_image_transform/model-metadata.yaml
scottp-ml: The default is supported.

4dahalibut (approved):

task_templates/transforms/python3_image_transform/custom.py
```diff
@@ -0,0 +1,77 @@
...
+from img_utils import (
+    b64_to_img,
+    img_to_b64,
+    img_to_grayscale,
+)
+
+
+def _process_image(raw_data: Union[str, bytes]) -> bytes:
+    img = b64_to_img(raw_data)
+    img = img_to_grayscale(img)
```
eric-s-s, line 13: please use different names, ex:
```suggestion
    initial_img = b64_to_img(raw_data)
    grayscale_img = img_to_grayscale(initial_img)
```

custom_model_runner/datarobot_drum/drum/typeschema_validation.py
eric-s-s: can you add a unit test please

eric-s-s (approved):
i want that unit test.  otherwise looks good

task_templates/transforms/python3_image_transform/model-metadata.yaml
```diff
@@ -0,0 +1,35 @@
+name: python3_anomaly_detection
+type: training
+targetType: transform
+# modelID: optional
+environmentID: 5e8c889607389fe0f466c72d
```
eric-s-s, line 5: silly question.  what is this environmentID ?  is it a specific execution env in a specific instance of DR?

scottp-ml: environment IDs are defined in the dropen environment env_info file and is the same between environments

PR #417, opened 2021-09-01
[RAPTOR-6035]predict default schema

scottp-ml:

## Summary
Default schema was not being used when just predict is run.  This caused a task that should have an error caught by the schema to instead error within datarobot by returning the wrong type.


tests/drum/test_fit.py
```diff
@@ -369,6 +367,33 @@ def test_transform_fit_fails_default(
             )
             assert "WARNING: No type schema provided. For transforms, we" not in stdout
 
+    @pytest.mark.parametrize(
+        "framework, language", [(SKLEARN_TRANSFORM_WITH_Y, PYTHON), (R_TRANSFORM_NO_Y, R_FIT),]
+    )
+    @pytest.mark.parametrize("problem", [REGRESSION, BINARY, ANOMALY])
+    def test_transform_fit_disallow_y_output(
```
jmbannon, line 374: was  this meant for a different PR?

scottp-ml: yes

custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
```diff
@@ -57,8 +58,11 @@ def configure(self, params):
             )
 
         model_metadata = read_model_metadata_yaml(self._code_dir)
-        self._schema_validator = (
-            SchemaValidator(model_metadata.get("typeSchema", {})) if model_metadata else None
+
+        # Set the environment variable STRICT_VALIDATION to use validation
+        self._schema_validator = SchemaValidator.create_validator(
+            model_metadata=model_metadata,
+            strict_validation=os.getenv("STRICT_VALIDATION") is not None,
```
jmbannon, line 65: should use `os.getenv("STRICT_VALIDATION", False)` because this will be set to true even if `STRICT_VALIDATION=0`

custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
```diff
@@ -57,8 +58,13 @@ def configure(self, params):
             )
 
         model_metadata = read_model_metadata_yaml(self._code_dir)
-        self._schema_validator = (
-            SchemaValidator(model_metadata.get("typeSchema", {})) if model_metadata else None
+
+        # Set the environment variable STRICT_VALIDATION to use validation
+        print(os.getenv("STRICT_VALIDATION", False) in [1, "True", "true", "TRUE"])
```
4dahalibut, line 63: I don't think this is good, but I don't know what it is.

scottp-ml: debug, removed

PR #421, opened 2021-09-09
[RAPTOR-6306] R Sparse Transforms: fix off-by-one error

jmbannon:

## Summary
Testing the recent R sparse changes on staging, I found there was an off-by-one error.
This made me realize and fix two things:
1. If the last row in the R sparse matrix is all zeroes, it will be dropped when creating it in python
2. R indices are 1-based, python is 0-based, so we need to translate

This PR fixes both those issues. No new tests are added because the fix for (1), passing the number of rows and cols back, makes tests fail unless we implement (2) because without it, the last row would be out of range

In addition, checks are added to make sure number of input rows matches output rows

### DIFF
--- a/tests/fixtures/r_transform_fit_custom_non_numeric.R
+++ b/tests/fixtures/r_transform_fit_custom_non_numeric.R
@@ -2,11 +2,8 @@ library(stringi)
 
 fit <- function(X, y, output_dir, class_order=NULL, row_weights=NULL, ...){}
 
-makestr <- function(n){
-    stri_rand_strings(1, n)
-}
-
 transform <- function(X, transformer, y=NULL, ...){
-    # ignore the model and output string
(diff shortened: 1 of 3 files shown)


tests/fixtures/r_transform_fit_custom_non_numeric.R
```diff
@@ -2,11 +2,8 @@ library(stringi)
...
 
-makestr <- function(n){
-    stri_rand_strings(1, n)
-}
-
 transform <- function(X, transformer, y=NULL, ...){
-    # ignore the model and output string
-    as.data.frame(sapply(X, makestr), stringAsFactors=FALSE)
+    # ignore the model and output strings for each dataframe cell
+    X[] <- lapply(X, function(z) { stri_rand_strings(1, 5) })
```
jmbannon, line 7: the original code produced a different number of rows from the original X

meltzg (approved):

PR #424, opened 2021-09-16
[RAPTOR-5872]update messages for users

scottp-ml:

## Summary
Update a couple of the messages for users when schema validation fails.  This should make it more obvious what the issue is.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/typeschema_validation.py
+++ b/custom_model_runner/datarobot_drum/drum/typeschema_validation.py
@@ -394,7 +394,7 @@ class NumColumns(BaseValidator):
         passes = conditions_map[self.condition](n_columns, test_value)
         if not passes:
             return [
-                f"Num columns error, {n_columns} did not satisfy: {self.condition} {test_value}"
+                f"Num columns error, {n_columns} received but the schema says: {self.condition} {test_value}"
             ]
 
         return []
@@ -421,7 +421,7 @@ class ContainsMissing(BaseValidator):
             io_type = "Output"
 
         if any_missing and value in [missing_output_disallowed, missing_input_disallowed]:
-            return [f"{io_type} contains missing values, the task does not support missing."]
+            return [f"{io_type} contains missing values, but the schema says that they are not allowed."]
         return []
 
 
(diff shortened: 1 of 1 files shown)


drenfr01 (approved):
LGTM

custom_model_runner/datarobot_drum/drum/typeschema_validation.py
```diff
@@ -394,7 +394,7 @@ def validate(self, dataframe):
         passes = conditions_map[self.condition](n_columns, test_value)
         if not passes:
             return [
-                f"Num columns error, {n_columns} did not satisfy: {self.condition} {test_value}"
+                f"Num columns error, {n_columns} received but the schema says: {self.condition} {test_value}"
```
4dahalibut, line 397:
```suggestion
                f"Incorrect column length. {n_columns} received. However, the schema dictates that columns should be {self.condition} {test_value}"
```

custom_model_runner/datarobot_drum/drum/typeschema_validation.py
```diff
@@ -421,7 +421,9 @@ def validate(self, dataframe):
             io_type = "Output"
 
         if any_missing and value in [missing_output_disallowed, missing_input_disallowed]:
-            return [f"{io_type} contains missing values, the task does not support missing."]
+            return [
+                f"{io_type} contains missing values, but the schema says that they are not allowed."
```
4dahalibut, line 425:
```suggestion
                f"{io_type} contains missing values, which the supplied schema does not allow."
```

4dahalibut (changes requested):
slightly different wording maybe

custom_model_runner/datarobot_drum/drum/typeschema_validation.py
MikhailYakubovskiy: How about 
f"Incorrect number of columns. {n_columns} received. However, the schema dictates that number of columns should be {self.condition} {test_value}"

custom_model_runner/datarobot_drum/drum/typeschema_validation.py
MikhailYakubovskiy: How about
f"{io_type} contains missing values, which the supplied task schema does not allow."

4dahalibut (approved):

PR #426, opened 2021-09-17
[YOLO] output string instead of bytes in img template

scottp-ml:

## Summary
Update the image utilities in the image transform template to output strings instead of bytes.  Bytes caused an issue now that validation of transform output is running where images were failing to load and being counted as categorical.


4dahalibut (approved):

drenfr01 (approved):
LGTM!

PR #433, opened 2021-09-22
[MODEL-6972] Write AD Estimator in R

drenfr01:

## Summary
Adding template for AD in R and also modifying score.R to allow anomaly target type

## Rationale
Presumably our R users will also want examples for anomaly detection


drenfr01:
jarvis test failure

drenfr01:
jarvis test failed

drenfr01:

scottp-ml (approved):
LGTM

task_templates/pipelines/r_lang_anomaly_detection/README.md
```diff
@@ -0,0 +1,14 @@
...
+The expected arguments to the fit method should remain the same, although the internal functionality can be tweaked to 
+use different modeling or preprocessing techniques.
+
+Inside you will find several commented out methods related to prediction behavior. 
+Uncomment and implement provided methods to modify this behavior from the default.
+
+### To run locally using 'drum'
+Paths are relative to `datarobot-user-models` root:
+`drum fit --code-dir task_templates/pipelines/r_lang --input tests/testdata/boston_housing.csv --target-type regression --target MEDV`
+If the command succeeds, your code is ready to be uploaded.
```
jmbannon, line 13: Update regression/classification to anomaly detection

task_templates/pipelines/r_lang_anomaly_detection/create_pipeline.R
```diff
@@ -0,0 +1,11 @@
+create_pipeline<-function(X) {
```
jmbannon, line 1: We should do an estimator instead of a pipeline, I thought we wanted to steer away from pipelines?

jmbannon (commented):
We should also add this as a functional test to `tests.functional.test_custom_task_templates.TestCustomTaskTemplates.test_custom_task_templates`

task_templates/pipelines/r_lang_anomaly_detection/create_pipeline.R
drenfr01: @jmbannon I thought we were just going to put the pipelines folder under estimators? I'm fine with either, do you have a strong opinion?

jmbannon: right but we want to encourage users to use DR preprocessing and not have to write single-task BPs w/everything in a pipeline. I think we should make this an estimator. Thoughts @MikhailYakubovskiy ?

MikhailYakubovskiy: right, we want to have a few pipelines as they are useful in some use cases (e.g. once we have PPS, they'll likely require pipelines?). But generally we want users to rely on blueprints. We also want to have fewer templates to maintain. Let's have a simple R anomaly estimator, but not create a pipeline for it.

drenfr01: Sold! I'll redo this as an estimator then

tests/functional/test_custom_task_templates.py
```diff
@@ -84,13 +92,6 @@ def project_multiclass_skyserver(self):
                 "pytorch_drop_in_env",
                 "binary",
             ),
-            (
```
drenfr01, line 87: Note: removing tests below that no longer are applicable. Before we had model-metadata.yaml working completely you could take the same code folder and apply target types. Usually this worked because there was logic in the fit hook to switch between classification and regression & there was no score hook (so DataRobot dynamically applied the right scoring functions). Now with model-metadata.yaml we have to specify the type, so this approach won't work.

After discussing with Jesse we decided not to create a bunch of new pipeline templates (because they're being discouraged in the future and test time considerations). Instead I just removed all the tests that won't work

drenfr01:
@yakov-g first of all, hysterical Github profile pic. Second, thanks for the heads up! Mind letting me know when it's in?

yakov-g:

task_templates/estimators/r_anomaly_detection/requirements.txt
```diff
@@ -0,0 +1,2 @@
+ranger
```
jmbannon, line 1: In a follow-up maybe mention how requirements.txt work with R?

jmbannon (approved):
beautiful

PR #436, opened 2021-09-26
[RAPTOR-6477] make uwsgi an extra dependency for DRUM

yakov-g:

## Summary
I changed `uwsgi` to be an extra dependency for `mlpiper` instead of mandatory. Thus also making it an extra dependency for DRUM.

+ Fix docs


yakov-g:

docker/java_dropin_env_base/Dockerfile
```diff
@@ -38,5 +38,6 @@ RUN pip3 install --no-cache-dir setuptools wheel
 
 
 # do the trick to install drum deps
-RUN pip3 install --no-cache-dir datarobot-drum
+# uwsgi is an extra dependency and it is not installed together with DRUM, so install it explicitly
+RUN pip3 install --no-cache-dir datarobot-drum uwsgi
```
meltzg, line 42: we don't actually use it though in our environments. can we skip it?

meltzg (approved):
would like it removed from the env base images unless we really need it. Otherwise g2g

yakov-g:
> would like it removed from the env base images unless we really need it. Otherwise g2g
Currently it is in the base images anyway, as previously mlpiper installed uwsgi.

drenfr01 (approved):
LGTM - but I assume we don't need to update the environments because this is only a local change?

PR #438, opened 2021-09-27
[RAPTOR-6439] handle defining bool args param through env var

yakov-g:

## Summary
When such env var was defined as False, e.g.:    PRODUCTION=False, it has been treated as "defined",  
thus argument "--production" has been passed into DRUM

### DIFF
--- a/tests/drum/test_args_parser.py
+++ b/tests/drum/test_args_parser.py
@@ -1,4 +1,5 @@
 import argparse
+import os
 import sys
 from tempfile import NamedTemporaryFile
 from unittest.mock import patch
@@ -6,8 +7,9 @@ from unittest.mock import patch
 import pytest
 
 from datarobot_drum.drum.args_parser import CMRunnerArgsRegistry
-from datarobot_drum.drum.common import ArgumentsOptions
+from datarobot_drum.drum.common import ArgumentsOptions, ArgumentOptionsEnvVars
 from datarobot_drum.resource.utils import _exec_shell_cmd
+from datarobot_drum.drum.utils import unset_drum_supported_env_vars
 
 
 class TestDrumHelp:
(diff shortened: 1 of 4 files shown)


yakov-g:

yakov-g:

yakov-g:

yakov-g:

meltzg (approved):

tests/drum/test_args_parser.py
```diff
@@ -149,3 +151,23 @@ def test_enable_strict_validation(self, parser):
...
+
+class TestBooleanArgumentOptions:
+    def test_boolean_argument_options_passed_through_env_var(self):
+        args = f"{ArgumentsOptions.MAIN_COMMAND} server  --with-error-server --skip-predict".split()
+        with patch.object(sys, "argv", args):
+            unset_drum_supported_env_vars()
+            CMRunnerArgsRegistry.get_arg_parser()
+            os.environ[ArgumentOptionsEnvVars.PRODUCTION] = "True"
+            os.environ[ArgumentOptionsEnvVars.MONITOR] = "False"
+            os.environ[ArgumentOptionsEnvVars.WITH_ERROR_SERVER] = "1"
```
jmbannon, line 164: how about a test for "0" as well

jmbannon: (okay with adding it in a follow-up)

jmbannon (approved):

tests/drum/test_args_parser.py
yakov-g: Here I test that if  `--with-error-server` is in CMD and in the env var, it ends up with a single  `--with-error-server` in the sys.argv.

The case you referring to, probably is the case with `--skip-predict`

Testing `False` vs `0` doesn't really make sense as its going to be checking that trafaret api works correctly

PR #440, opened 2021-09-30
[YOLO] release 1.5.16

scottp-ml:


jmbannon (approved):

yakov-g (changes requested):
missing keras and pmml drum version bump

we have 8 envs so there should be 8 pairs of files changed: requirements and info

yakov-g (approved):

PR #446, opened 2021-10-08
[RAPTOR-6382] fix solver issue

scottp-ml:

## Summary
There is a known bug with the default solver used in logistic regression for multiclass.  This PR switches the solver to a known working one.


yakov-g:
jarvis please test black

PR #450, opened 2021-10-13
[RAPTOR-6579] Fix bool being in num not cat edge

scottp-ml:

## Summary
In DataRobot boolean columns, which are True/False, are passed along the NUM edge and not the Categorical edge.  These were incorrectly getting called CAT and failing validation when only NUM was expected.  A warning is given when boolean values are encountered to help users debug  their task if they were not anticipating the presence of boolean values.  To help with this there is now also an example transform that can be used to preprocess the bools to integer values.


yakov-g:

yakov-g:
jarvis test functional_inference_model_templates

yakov-g:

4dahalibut:

yakov-g:

PR #452, opened 2021-10-14
[YOLO] use yaml.safe_load()

yakov-g:

## Summary
Tests fail because new pyyaml==6.0 requires additional argument in the `load()` method, use `safe_load()` instead


yakov-g:

yakov-g:

yakov-g:

PR #454, opened 2021-10-15
[RAPTOR-6805] list files in case multiple artifacts are found

yakov-g:


yakov-g:

PR #457, opened 2021-10-18
Add new cases to functional tests [RAPTOR-6641]

gkoundry:

## Summary
Add functional test cases for:
- A blueprint with a DR pre-processing task (PNI2) and a custom task estimator (pytorch)
- A 3-class multiclass dataset with a user partition column that ensures the training data only has 2 classes. 
- A project with weights and a custom task that checks that the weights are correct


gkoundry:

PR #461, opened 2021-10-20
[RAPTOR-6182] Add validation on model metadata output_requirements

lkanggithub:
## Summary
The model metadata (in yaml) defines the essential info (e.g., target type, input output specs) of a custom task (schema is defined at https://github.com/datarobot/DataRobot/blob/d5e77b79f90ea7dfde0683cbd474becd457791ce/raptor_lake/custom_model/model_metadata_utils.py#L541-L557). The `output_requirements` (i.e., ModelMetadataKeys.OUTPUT_REQUIREMENTS) specifies the output spec of one custom task. The `output_requirements` only applies to a custom task of which target type is Transformer.
This PR adds the validation on model metadata before other checks are triggered in CMRunner.run(). When `output_requirements` is found and the custom task is not a Transform task, error will be returned to user.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/drum.py
+++ b/custom_model_runner/datarobot_drum/drum/drum.py
@@ -463,9 +463,6 @@ class CMRunner:
 
         self._print_welcome_header()
 
-        if hasattr(self, "schema_validator"):
-            self.schema_validator.validate_type_schema(self.target_type)
-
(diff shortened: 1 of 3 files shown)


lkanggithub:
jarvis please test functional_drop_in functional_training_model_templates

PR #463, opened 2021-10-20
[RAPTOR-6106] Refactor prediction validation and marshalling

4dahalibut:

## Summary
I moved predictions validation and marshalling from being different in different artifact/language predictors to being centralized in the base language predictor, and then I tested it better

## Rationale
While trying to straighten out class label marshalling, I decided that all the artifact predictors share logic that would better be done in one place. Now we pass raw labels downstream and do all marshalling in one place


tests/drum/test_data_marshalling.py
```diff
@@ -27,6 +32,70 @@ def test__order_by_float():
...
+    ):
+        marshal_predictions(
+            request_labels=labels, predictions=preds, target_type=TargetType.MULTICLASS
+        )
+
+
+def test_marshal_predictions_binary_reshape_happy():
+    preds = np.array([1, 1, 1])
+    labels = [1, 2]
+    marshal_predictions(request_labels=labels, predictions=preds, target_type=TargetType.BINARY)
```
meltzg, line 62: please test that the resulting predictions are correct

tests/drum/test_data_marshalling.py
```diff
@@ -27,6 +32,70 @@ def test__order_by_float():
...
+
+
+def test_marshal_predictions_reshape_regression_happy():
+    preds = np.array([1, 1, 1])
+    labels = [1]
+    marshal_predictions(request_labels=labels, predictions=preds, target_type=TargetType.REGRESSION)
```
meltzg, line 68: please test that the resulting predictions are correct

tests/drum/test_data_marshalling.py
meltzg: please add test for binary with both values supplied in each row

tests/drum/test_data_marshalling.py
```diff
@@ -27,6 +32,70 @@ def test__order_by_float():
...
+    expected = pd.DataFrame(data=preds, columns=labels)
+    pd.testing.assert_frame_equal(
+        marshal_predictions(
+            request_labels=labels, predictions=preds, target_type=TargetType.MULTICLASS
+        ),
+        expected,
+    )
+
+
+def test_marshal_predictions_multiclass_wrong_label_length():
```
meltzg, line 47: please add a test for num columns > num labels

tests/drum/test_data_marshalling.py
```diff
@@ -27,6 +32,70 @@ def test__order_by_float():
...
+def test_marshal_predictions_bad_shape_regression():
+    labels = [1]
+    preds = np.zeros((1, 2))
+    with pytest.raises(DrumCommonException, match="must contain only 1 column"):
+        marshal_predictions(
+            request_labels=labels, predictions=preds, target_type=TargetType.REGRESSION
+        )
+
+
+def test_marshal_predictions_dont_add_to_one():
```
meltzg, line 95: please add a test for invalid other ways of adding to 1 (ex. 2, -1)

meltzg (changes requested):
needs more test cases

tests/drum/test_data_marshalling.py
4dahalibut: how come this is bad?

meltzg: because 2, -1 isn't a valid probability distribution but it still sums to 1

4dahalibut: what is invalid about it? the fact there is a negative? I can assert no negatives if you'd like

meltzg: Yes, as insinuated in my previous comment, -1 is not a valid probability. Probabilities should be in [0.0, 1.0]

custom_model_runner/datarobot_drum/drum/data_marshalling.py
```diff
@@ -0,0 +1,164 @@
...
+        pass
+
+    # Okay lets just do a str.lower
+    assert isinstance(label, str)
+    return label.lower()
+
+
+def _classification_marshal_preds(predictions, request_labels, model_labels):
+    request_labels = _marshal_labels(request_labels, model_labels)
+    predictions = _duplicate_1d_probabilities(predictions, request_labels)
```
meltzg, line 94: nit: _infer_negative_class_probabilities or something like that is probably a better name, but not enough to warrant rerunning tests

meltzg (approved):
minor naming nit, but otherwise, great job. Glad to see code made common

custom_model_runner/datarobot_drum/drum/data_marshalling.py
```diff
@@ -0,0 +1,164 @@
...
+        raise DrumCommonException(
+            f"Regression and anomaly predictions must contain only 1 column. Your predictions have {predictions.shape[1]} columns"
+        )
+
+
+def _validate_probabilities_make_sense(predictions):
+    if (predictions < 0).any():
+        raise DrumCommonException("Your predictions have negative values")
+    if (predictions > 1).any():
+        raise DrumCommonException("Your predictions have values greater than 1")
```
jmbannon, line 130: nit: Your prediction probabilities

custom_model_runner/datarobot_drum/drum/data_marshalling.py
```diff
@@ -0,0 +1,164 @@
...
+    if predictions.shape[1] != len(labels_to_use):
+        raise DrumCommonException(
+            "Your predictions must return the "
+            "probability distribution for the correct number of class labels. "
+            "Expected {} columns, but received {}".format(len(labels_to_use), predictions.shape[1])
+        )
+
+
+def _duplicate_1d_probabilities(predictions, labels):
+    if predictions.shape[1] == 1:
```
jmbannon, line 150: we should make sure target type is binary here

custom_model_runner/datarobot_drum/drum/data_marshalling.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def _validate_predictions_are_one_dimensional(predictions):
+    if predictions.shape[1] != 1:
+        raise DrumCommonException(
+            f"Regression and anomaly predictions must contain only 1 column. Your predictions have {predictions.shape[1]} columns"
+        )
+
+
+def _validate_probabilities_make_sense(predictions):
```
jmbannon, line 126: nit: _validate_probabilities_sum_to_one

jmbannon (approved):

custom_model_runner/datarobot_drum/drum/data_marshalling.py
4dahalibut: we actually can duplicate for the multiclass case too where there are only 2 labels.

PR #471, opened 2021-10-28
[RAPTOR-6642] Validate transform colnames before serialization

jmbannon:

## Summary
For transforms, we should ensure colnames are sane and validated before we serialize

## Changes
- Add validation check and conversion for both sparse and dense transform payloads
- Added unit test


jmbannon:

4dahalibut (approved):

PR #473, opened 2021-11-01
Fix class labels [RAPTOR-6493]

gkoundry:
## Summary
For multiclass projects, use class labels from environment or command line option or else get them from the target without sampling the target.


custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -1109,7 +1115,7 @@ def possibly_intuit_order(
             )
             print(e, file=sys.stderr)
             raise DrumCommonException(e)
-        uniq = df[target_col_name].sample(min(1000, len(df)), random_state=1).unique()
+        uniq = df[target_col_name].unique()
```
4dahalibut, line 1118: did we want to sort these?

gkoundry: They weren't sorted before this PR.  Is there a reason to?

4dahalibut: so the order doesn't change around between runs...

jmbannon (approved):

custom_model_runner/datarobot_drum/drum/drum.py
gkoundry: Sort has been added.  Can you approve?

4dahalibut (approved):

PR #476, opened 2021-11-02
[RAPTOR-6794] add some more messages for users

scottp-ml:

## Summary
Adds messages for the start and stop of training and prediction steps when running drum fit.  Also runs drum fit pipeline in verbose mode so that any exceptions in the pipeline are shown without the user having to add in the --verbose flag.  Only part of it is verbose, so if the user wants timing data they will still need to use the verbose flag.

## Rationale
Make drum easier to use by always providing important failure information so that the user doesn't have to run a second time with --verbose.


custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -508,15 +508,18 @@ def run_fit(self):
...
             interval=1,
             max_usage=True,
             max_iterations=1,
         )
+        print("Fit successful")
         if self.options.verbose:
             print("Maximum fit memory usage: {}MB".format(int(mem_usage)))
-        if self.options.output or not self.options.skip_predict:
+        if self.options.output and not self.options.skip_predict:
+            print("Prediction started")
```
jmbannon, line 522: nit: make it `Starting predict` to match fit's message

jmbannon (approved):

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -508,15 +508,18 @@ def run_fit(self):
...
             interval=1,
             max_usage=True,
             max_iterations=1,
         )
+        print("Fit successful")
         if self.options.verbose:
             print("Maximum fit memory usage: {}MB".format(int(mem_usage)))
-        if self.options.output or not self.options.skip_predict:
+        if self.options.output and not self.options.skip_predict:
+            print("Starting Prediction")
```
4dahalibut, line 522: can you move this under the following if statement?

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -758,7 +762,7 @@ def _run_fit_or_predictions_pipelines_in_mlpiper(self):
         sc.register_report("Full time", "end", StatsOperation.SUB, "start")
         sc.register_report("Init time (incl model loading)", "init", StatsOperation.SUB, "start")
         sc.register_report("Run time (incl reading CSV)", "run", StatsOperation.SUB, "init")
-        with verbose_stdout(self.options.verbose):
+        with verbose_stdout(True):
```
4dahalibut, line 765: why this change?

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -525,6 +528,7 @@ def run_fit(self):
             if self.options.verbose:
                 print("Maximum server memory usage: {}MB".format(int(mem_usage)))
             pred_str = " and predictions can be made on the fit model! \n "
+            print("Prediction successful")
```
4dahalibut, line 531: can you add that this is during drum fit?

4dahalibut (approved):

PR #486, opened 2021-11-08
Fix check for binary target labels [RAPTOR-6424]

gkoundry:

## Summary
Fix bug in code that checks if binary class labels are in the  environment vars.


custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -648,7 +648,10 @@ def _prepare_prediction_server_or_batch_pipeline(self, run_language):
 
     def _needs_class_labels(self):
         if self.target_type == TargetType.BINARY:
-            return self.options.negative_class_label is None or self.options.class_labels is None
+            return (
+                self.options.negative_class_label is None
```
4dahalibut, line 652: do you think we should allow passing in a classLabels.txt for binary target type?

gkoundry: This bit of code affect whether the labels are taken from EITHER the class labels file or the unique target values if that is what you are asking, so it can't be removed.  Or are you thinking about changing the logic of `possibly_intuit_order`?

gkoundry:
jarvis test this

gkoundry:
jarvis test functional_drop_in functional_inference_model_templates functional_training_model_templates

4dahalibut (approved):

scottp-ml (approved):

PR #491, opened 2021-11-12
RAPTOR-7061 Do not block python thread while trying to read stdout from DRUM server

alexromanyuk:
## Summary
Drum perf-test hangs in the Google Colab environment:
image

but it works fine after stop/start cell again.

## Changes
- Set the timeout for `request.get` to 1 second. Otherwise it blocks thread indefinitely, so server process could never start.
- Increased timeout for `wait_for_server` from 20 to 60 seconds. Drum needs ~30 seconds to start in Google Colab.
- Changed blocking operation `process.communicate()` to non-blocking read from the pipe. So we actually could see errors. See example logs of failed process in Google Colab:
non-blocking-io


custom_model_runner/datarobot_drum/drum/perf_testing.py
```diff
@@ -319,13 +321,20 @@ def _wait_for_server_to_start(self):
             self._timeout = self._timeout - 1
             if self._timeout == 0:
                 error_message = "Error: server failed to start while running performance testing."
-                (stdout, stderr) = self._server_process.communicate()
```
alexromanyuk, line 322: `self._server_process.communicate()` blocks until process is terminated, which will never occur.

zohar-mizrahi (approved):

custom_model_runner/datarobot_drum/drum/perf_testing.py
```diff
@@ -309,7 +311,7 @@ def _get_size_and_units(size):
     def _wait_for_server_to_start(self):
         while True:
             try:
-                response = requests.get(self._url_server_address)
+                response = requests.get(self._url_server_address, timeout=1)
```
rvorobii, line 314: what if 1s is not enough, and will give false positives for already running server?

rvorobii: I think we should use `_timeout` instead of 1s -- in does't matter if the request timed out during a single request or after several requests
```suggestion
                response = requests.get(self._url_server_address, timeout=_timeout)
```

rvorobii (changes requested):
please check my comment -- I think this is a proper way of handling timeout

custom_model_runner/datarobot_drum/drum/perf_testing.py
rvorobii: > we need to remove the while: True loop

we remove the loop?

rvorobii (approved):

PR #495, opened 2021-11-17
Call transform hook before fit [RAPTOR-6103]

gkoundry:

## Summary
Make sure transform hook is called before fit for prediction tasks

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/model_adapter.py
+++ b/custom_model_runner/datarobot_drum/drum/model_adapter.py
@@ -602,7 +602,7 @@ class PythonModelAdapter:
                     row_weights=row_weights,
                     parameters=parameters,
                 )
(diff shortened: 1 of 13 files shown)


gkoundry:

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -323,7 +323,7 @@ def load_data(self, binary_data, mimetype, try_hook=True, sparse_colnames=None):
 
         return data
 
-    def preprocess(self, model=None, **kwargs):
+    def preprocess(self, data, model=None):
```
4dahalibut, line 326: I think we should keep kwargs here even if we don't use it.

4dahalibut (approved):

rvorobii (approved):

PR #496, opened 2021-11-19
[RAPTOR-5478]update task routes

scottp-ml:

## Summary
Updates drum push to use the custom tasks version in the DR client to align with new routes.


alexromanyuk (approved):

eric-s-s (approved):

scottp-ml:
2,27 should be available dec 13, I will update at that time.

PR #497, opened 2021-11-23
[RAPTOR-7213] Read form data for unstructured models.

alexromanyuk:
## Summary
DRUM don't see data sent to `predictUnstructured` if user haven't specified request header explicitly (curl sends it with the `application/x-www-urlencoded`):
```
curl -i -X POST http://127.0.0.1:6788/predictUnstructured/ --data-binary @unstructured_data.txt 
```
DRUM --verbose:
```
Incoming data type:  <class 'bytes'>
Incoming data:  b''
Incoming query params:  ImmutableMultiDict([])
Model:  dummy
Incoming content type params:  {'mimetype': 'application/x-www-form-urlencoded'}
```

## Changes
Use `request.get_data()` which does not tries to parse form data (see [Flask docs](https://flask.palletsprojects.com/en/2.0.x/api/#flask.Request.get_data)).

### DIFF
--- a/tests/drum/test_unstructured_mode.py
+++ b/tests/drum/test_unstructured_mode.py
@@ -107,7 +107,8 @@ class TestUnstructuredMode:
                     url = run.url_server_address + endpoint
                     data = open(input_dataset, "rb").read()
                     params = {"ret_mode": ret_mode}
-                    response = requests.post(url=url, data=data, params=params)
+                    headers = {"Content-Type": "application/x-www-urlencoded"}
+                    response = requests.post(url=url, headers=headers, data=data, params=params)
 
                     assert response.ok
                     if ret_mode == "text":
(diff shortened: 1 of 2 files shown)


yakov-g (approved):

tests/drum/test_unstructured_mode.py
```diff
@@ -264,14 +265,17 @@ def test_response_one_var_return(
...
-                assert response.ok
-                content_type_header = response.headers["Content-Type"]
-                mimetype, content_type_params_dict = werkzeug.http.parse_options_header(
-                    content_type_header
-                )
-                assert "application/octet-stream" == mimetype
-                # check params dict is empty
-                assert not any(content_type_params_dict)
-                assert response.content == data_bytes
+                for ct in ["application/octet-stream;", "application/x-www-urlencoded;"]:
```
rvorobii, line 268: with this we have 2 tests in 1 -- let's move it to `pytest.mark.parametrize`

rvorobii (approved):

tests/drum/test_unstructured_mode.py
yakov-g: But parametrize will run test two times, which will start/stop server two times so overall runtime will be longer.
Of course it's negligible for a single test, but we have a lot of such cases, which I implemented without parameterized to save a lot of time.

PR #499, opened 2021-11-24
[RAPTOR-6350] Handle multiline text containing \r

scottp-ml:

## Summary
Handle text columns that contain \r as well as \n.  This allows drum to work correctly locally without modifying the input file.


scottp-ml:

zohar-mizrahi (approved):
LGTM

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -550,12 +550,12 @@ def run_test_predict(self):
             if self.target_type == TargetType.TRANSFORM:
                 target_df = df[self.options.target]
                 __target_temp = NamedTemporaryFile()
-                target_df.to_csv(__target_temp.name, index=False)
+                target_df.to_csv(__target_temp.name, index=False, line_terminator="\r\n")
```
drenfr01, line 553: @scottp-ml Just to confirm, the idea here is that datasets prepared on Mac OS X or Linux systems will ignore the \r here, right?

drenfr01 (commented):
@scottp-ml can you explain more how writing "/r" allows drum to handle datasets with carriage returns? It doesn't look like you modify anywhere how we read datasets...

scottp-ml:
The issue is that the dataset is initially read in correctly, but then we write it back out and read it a second time.  When we write it out the \r inside the string was causing new rows with empty columns because pandas only was dealing with \n in text as something to escape.

drenfr01 (approved):
LGTM!

PR #508, opened 2021-12-07
[RAPTOR-6049] Initialize default task parameter values with hyperparameters in model metadata config

lkanggithub:
## Summary
The `parameterFile` argument of drum is used to indicate the file of task hyper-parameter values. When the `parameterFile` is not specified, the task hyper-parameter values won't be set. With the changes of this PR, the task hyper-parameter values will be initialized with the default values from model-metadata.yaml (if 'hyperparameters' section is available) if `parameterFile` drum argument is not specified.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/drum.py
+++ b/custom_model_runner/datarobot_drum/drum/drum.py
@@ -77,8 +77,10 @@ class CMRunner:
         self.runtime = runtime
         self.options = runtime.options
         self.options.model_config = read_model_metadata_yaml(self.options.code_dir)
-        self.options.default_parameter_values = get_default_parameter_values(
-            self.options.model_config
+        self.options.default_parameter_values = (
+            get_default_parameter_values(self.options.model_config)
+            if self.options.model_config
+            else {}
         )
         self.logger = CMRunner._config_logger(runtime.options)
         self.verbose = runtime.options.verbose
--- a/custom_model_runner/datarobot_drum/drum/drum.py
+++ b/custom_model_runner/datarobot_drum/drum/drum.py
@@ -1175,9 +1175,68 @@ def create_custom_inference_model_folder(code_dir, output_dir):
 
 
 def get_default_parameter_values(model_metadata: Dict) -> Dict[str, Union[int, float, str]]:
-    hyper_param_config = model_metadata.get(ModelMetadataKeys.HYPERPARAMETERS, {})
+    """Retrieve default parameter values from the hyperparameter section of model-metadata.yaml.
+    When `default` is provided, return the default value.
+    When `default` is not provided:
+        - Return `min` value if it is a numeric parameter.
+        - Return an emtpy string if it is a string parameter.
+        - Return the first allowed value if it is the select parameter.
+        - Return the default value of the first component parameter (sorted by parameter type) if it is a multi
+          parameter.
+    """
+
+    def _get_default_numeric_param_value(param_config: Dict) -> Union[int, float]:
+        """Get default value of numeric parameter."""
+        param_default_value = param_config.get("default")
+        if param_default_value is not None:
+            return param_default_value
+        else:
+            return param_config["min"]
+
+    def _get_default_string_param_value(param_config: Dict) -> str:
+        """Get default value of string parameter."""
+        param_default_value = param_config.get("default")
+        if param_default_value is not None:
+            return param_default_value
+        else:
+            return ""
+
+    def _get_default_select_param_value(param_config: Dict) -> str:
+        """Get default value of select parameter."""
+        param_default_value = param_config.get("default")
+        if param_default_value is not None:
+            return param_default_value
+        else:
+            return param_config["values"][0]
+
+    def _get_default_multi_param_value(param_config: Dict) -> str:
+        """Get default value of multi parameter."""
+        param_default_value = param_config.get("default")
+        if param_default_value is not None:
+            return param_default_value
+        else:
+            param_values = param_config["values"]
+            if param_values:
+                first_component_param_type = sorted(param_values.keys())[0]
+                first_component_param = param_values[first_component_param_type]
+                if first_component_param_type in {"int", "float"}:
+                    return _get_default_numeric_param_value(first_component_param)
+                elif first_component_param_type == "select":
+                    return _get_default_select_param_value(first_component_param)
+
+    hyper_param_config = model_metadata.get(ModelMetadataKeys.HYPERPARAMETERS, [])
     default_params = {}
     for param_config in hyper_param_config:
-        if param_config.get("default"):
-            default_params[param_config["name"]] = param_config["default"]
+        param_name = param_config["name"]
+        param_type = param_config["type"]
+        if param_type in {"int", "float"}:
+            default_params[param_name] = _get_default_numeric_param_value(param_config)
+        elif param_type == "string":
+            default_params[param_name] = _get_default_string_param_value(param_config)
+        elif param_type == "select":
+            default_params[param_name] = _get_default_select_param_value(param_config)
+        elif param_type == "multi":
+            default_param_value = _get_default_multi_param_value(param_config)
+            if default_param_value is not None:
+                default_params[param_name] = default_param_value
     return default_params
--- a/custom_model_runner/datarobot_drum/drum/drum.py
+++ b/custom_model_runner/datarobot_drum/drum/drum.py
@@ -15,13 +15,13 @@ import shutil
 import subprocess
 import sys
 import tempfile
+from tempfile import mkdtemp, NamedTemporaryFile
 import time
 from typing import Dict
 from typing import Union
 from distutils.dir_util import copy_tree
 from pathlib import Path
 from progress.spinner import Spinner
-from tempfile import mkdtemp, NamedTemporaryFile
 
 import numpy as np
 import pandas as pd
@@ -51,6 +51,7 @@ from datarobot_drum.drum.enum import (
     TargetType,
     TemplateType,
     ModelMetadataKeys,
+    ModelMetadataHyperParamTypes,
 )
 from datarobot_drum.drum.description import version as drum_version
 from datarobot_drum.drum.exceptions import DrumCommonException, DrumPredException
@@ -76,10 +77,8 @@ class CMRunner:
         self.runtime = runtime
         self.options = runtime.options
         self.options.model_config = read_model_metadata_yaml(self.options.code_dir)
-        self.options.default_parameter_values = (
-            get_default_parameter_values(self.options.model_config)
-            if self.options.model_config
-            else {}
+        self.options.default_parameter_values = get_default_parameter_values(
+            self.options.model_config
         )
         self.logger = CMRunner._config_logger(runtime.options)
         self.verbose = runtime.options.verbose
@@ -945,7 +944,7 @@ class CMRunner:
                 "run",
                 "-it",
                 "--entrypoint",
-                # provide emtpy entrypoint value to unset the one that could be set within the image
+                # provide empty entrypoint value to unset the one that could be set within the image
                 "",
                 options.docker,
                 "sh",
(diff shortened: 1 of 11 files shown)


lkanggithub:
jarvis please test this

lkanggithub:
jarvis please test cmrunner_integration

lkanggithub:
jarvis please test functional_training_model_templates functional_drop_in functional_inference_model_templates

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -1164,3 +1172,12 @@ def create_custom_inference_model_folder(code_dir, output_dir):
...
     if files_in_output & copied_files:
         print("Files were overwritten: {}".format(files_in_output & copied_files))
+
+
+def get_default_parameter_values(model_metadata: Dict) -> Dict[str, Union[int, float, str]]:
+    hyper_param_config = model_metadata.get(ModelMetadataKeys.HYPERPARAMETERS, {})
+    default_params = {}
+    for param_config in hyper_param_config:
+        if param_config.get("default"):
+            default_params[param_config["name"]] = param_config["default"]
```
jmbannon, line 1182: We need to get default values even if they do not supply a default value. It is defined in DR as:
- int/float: the min value
- select: the first value in their array
- string: empty string
- multi: the first multi type's default value

jmbannon (changes requested):
Please retrieve the defaults when a user doesn't supply a default

custom_model_runner/datarobot_drum/drum/drum.py
lkanggithub: > int/float: the min value

Not sure if this is the common practice to use e.g., `Integer.MIN_VALUE` as the default value if int/float is not provided...

> select: the first value in their array
If there is no `hyperparameters` section in model-metadata.yaml or no model-metadata.yaml is provided, we probably don't know what is the allowed values in list?

jmbannon: sorry for not being more clear, min value being the min value the user provides (since it's a required filed).

If there is no hyperparameters at all in model-metadata, then we don't need to worry about providing default values at all at runtime.

jmbannon: Here's a more explicit example. User does not provide any defaults to their parameters:
```
hyperparameters:
  - name: penalty  # default will be l1
    type: select
    values:
      - l1
      - l2
      - elasticnet
  - name: dual  # default will be 0
    type: int
    min: 0
    max: 1
  - name: tol  # default will be 0
    type: float
    min: 0
    max: 1
  - name: solver  # default will be empty string
    type: string
```

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -1164,3 +1172,71 @@ def create_custom_inference_model_folder(code_dir, output_dir):
...
+            return param_config["values"][0]
+
+    def _get_default_multi_param_value(param_config: Dict) -> str:
+        """Get default value of multi parameter."""
+        param_default_value = param_config.get("default")
+        if param_default_value is not None:
+            return param_default_value
+        else:
+            param_values = param_config["values"]
+            if param_values:
```
lkanggithub, line 1219: With the existing schema design, it is possible that `values` is an empty dictionary. The highlight logic is adapted to the following schema design.
```
PARAM_MULTI_T = t.Dict(
    {
        t.Key('name'): t.String(max_length=PARAM_NAME_MAX_LENGTH),
        t.Key('type'): t.Enum('multi'),
        t.Key('values'): t.Dict(
            {
                t.Key('int', optional=True): t.Dict({t.Key('min'): t.Int, t.Key('max'): t.Int,}),
                t.Key('float', optional=True): t.Dict(
                    {t.Key('min'): t.Float, t.Key('max'): t.Float,}
                ),
                t.Key('select', optional=True): t.Dict(
                    {
                        t.Key('values'): t.List(
                            t.String(max_length=PARAM_SELECT_VALUE_MAX_LENGTH, allow_blank=False),
                            min_length=1,
                            max_length=PARAM_SELECT_NUM_VALUES_MAX_LENGTH,
                        ),
                    }
                ),
            }
        ),
        t.Key('default', optional=True): t.Or(
            t.Int, t.Float, t.String(max_length=PARAM_SELECT_VALUE_MAX_LENGTH)
        ),
    }
)
```

custom_model_runner/datarobot_drum/drum/drum.py
lkanggithub: The logic was updated. 
https://github.com/datarobot/datarobot-user-models/blob/c40e34a2ece79f88bc649aa8235fc6935c0d8788/custom_model_runner/datarobot_drum/drum/drum.py#L1177-L1186

lkanggithub:
jarvis please test functional_training_model_templates

lkanggithub:
jarvis please test functional_training_model_templates

tests/drum/test_units.py
```diff
@@ -975,3 +976,180 @@ def test_binary_class_labels_from_target():
...
+                ],
+            },
+            0,
+        ),
+        (
+            {
+                ModelMetadataKeys.HYPERPARAMETERS: [
+                    {"name": "param_name", "type": "multi", "values": {},}
+                ],
+            },
```
jmbannon, line 1082: I think we should throw an exception if this is present in the params file. Pretty sure we error in DR when this happens as well

jmbannon (approved):
Looks good! Please fix the empty multi param bug in a follow-up

lkanggithub:
jarvis please test functional_training_model_templates

lkanggithub:
jarvis please test functional_training_model_templates

lkanggithub:
@4dahalibut Could you help review this pr? Thx!

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -16,6 +16,8 @@
 import sys
 import tempfile
 import time
+from typing import Dict
```
4dahalibut, line 19: sort imports

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -1164,3 +1172,71 @@ def create_custom_inference_model_folder(code_dir, output_dir):
...
     if files_in_output & copied_files:
         print("Files were overwritten: {}".format(files_in_output & copied_files))
+
+
+def get_default_parameter_values(model_metadata: Dict) -> Dict[str, Union[int, float, str]]:
+    """Retrieve default parameter values from the hyperparameter section of model-metadata.yaml.
+    When `default` is provided, return the default value.
+    When `default` is not provided:
+        - Return `min` value if it is a numeric parameter.
+        - Return an emtpy string if it is a string parameter.
```
4dahalibut, line 1182: typo

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -1164,3 +1172,71 @@ def create_custom_inference_model_folder(code_dir, output_dir):
...
+        - Return the first allowed value if it is the select parameter.
+        - Return the default value of the first component parameter (sorted by parameter type) if it is a multi
+          parameter.
+    """
+
+    def _get_default_numeric_param_value(param_config: Dict) -> Union[int, float]:
```
4dahalibut, line 1188: I don't see any reason to have these be nested functions.

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -74,6 +76,11 @@ def __init__(self, runtime):
         self.runtime = runtime
         self.options = runtime.options
         self.options.model_config = read_model_metadata_yaml(self.options.code_dir)
+        self.options.default_parameter_values = (
+            get_default_parameter_values(self.options.model_config)
```
4dahalibut, line 80: you might just want to handle the if else block in the function, also, you don't really need to set default_parameter_values here, you can just call the function when you need it below.

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -1164,3 +1172,71 @@ def create_custom_inference_model_folder(code_dir, output_dir):
...
+        - Return the default value of the first component parameter (sorted by parameter type) if it is a multi
+          parameter.
+    """
+
+    def _get_default_numeric_param_value(param_config: Dict) -> Union[int, float]:
+        """Get default value of numeric parameter."""
+        param_default_value = param_config.get("default")
+        if param_default_value is not None:
+            return param_default_value
+        else:
```
4dahalibut, line 1193: style nit, I don't think there is a need for an else statement.

custom_model_runner/datarobot_drum/drum/drum.py
4dahalibut: we should just insist that values is not an empty dictionary.

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -1164,3 +1172,71 @@ def create_custom_inference_model_folder(code_dir, output_dir):
...
+                    return _get_default_numeric_param_value(first_component_param)
+                elif first_component_param_type == "select":
+                    return _get_default_select_param_value(first_component_param)
+
+    hyper_param_config = model_metadata.get(ModelMetadataKeys.HYPERPARAMETERS, [])
+    default_params = {}
+    for param_config in hyper_param_config:
+        param_name = param_config["name"]
+        param_type = param_config["type"]
+        if param_type in {"int", "float"}:
```
4dahalibut, line 1232: could you define global vars for "int" "float" "select" "string" and "multi" and refactor to use them? Probably would prevent typos.

4dahalibut:
Will defaults be sent into drum from DR?

4dahalibut (commented):
Mostly style nits and questions.

custom_model_runner/datarobot_drum/drum/drum.py
lkanggithub: Could you be more specific on this?

jmbannon: since the `if` block returns, you can remove the else and it'll functionally be the same

custom_model_runner/datarobot_drum/drum/drum.py
lkanggithub: updated
https://github.com/datarobot/datarobot-user-models/blob/3f5e21b85c9a075401eb3cf13e13975c15648b97/custom_model_runner/datarobot_drum/drum/drum.py#L1212-L1217

custom_model_runner/datarobot_drum/drum/drum.py
lkanggithub: updated
https://github.com/datarobot/datarobot-user-models/blob/3f5e21b85c9a075401eb3cf13e13975c15648b97/custom_model_runner/datarobot_drum/drum/drum.py#L1178-L1199

custom_model_runner/datarobot_drum/drum/drum.py
lkanggithub: I would prefer saving the default param values as the attributes of self.options and make it reusable throughout the life cycle of runner. We may handle the if/else within the function. But this will allow `model_metadata` as the only argument of the function `get_default_parameter_values` be None (So... Not sure if this is a good design of function)

custom_model_runner/datarobot_drum/drum/drum.py
lkanggithub: The function was moved out the function. The logic is still encapsulated as the function. Cause it is used in multiple places and better to be encapsulated as func.

custom_model_runner/datarobot_drum/drum/drum.py
lkanggithub: fixed

lkanggithub:
@4dahalibut Could you help review this pr? thx

4dahalibut (approved):

PR #510, opened 2021-12-10
[RAPTOR-7352] Fix keras vizai pipeline

drenfr01:

## Summary

This template was broken because it:
 1) imported torch for no reason (which isn't in the dependencies list)
 2) tried to download weights at fit time (which won't work because it's our docker container has no internet access) 
 3) used an old version of MobileNet which is out of sync with core DataRobot (v2 vs. v3)

## Rationale

We need working templates, this came up because a CFDS tried to use this template and couldn't

### DIFF
--- a/task_templates/3_pipelines/15_python3_keras_vizai_joblib/model-metadata.yaml
+++ b/task_templates/3_pipelines/15_python3_keras_vizai_joblib/model-metadata.yaml
@@ -19,6 +19,7 @@ typeSchema:
       condition: IN # can be one of IN or NOT_IN
       value: # can be one or multiple of NUM, TXT, IMG, DATE, CAT, DATE_DURATION, COUNT_DICT, GEO
         - IMG
+        - NUM
 
     # specify how many columns the task requires
     - field: number_of_columns
--- a/task_templates/3_pipelines/15_python3_keras_vizai_joblib/model-metadata.yaml
+++ b/task_templates/3_pipelines/15_python3_keras_vizai_joblib/model-metadata.yaml
@@ -19,7 +19,6 @@ typeSchema:
       condition: IN # can be one of IN or NOT_IN
       value: # can be one or multiple of NUM, TXT, IMG, DATE, CAT, DATE_DURATION, COUNT_DICT, GEO
         - IMG
-        - NUM
 
     # specify how many columns the task requires
     - field: number_of_columns
--- a/task_templates/3_pipelines/15_python3_keras_vizai_joblib/model_utils.py
+++ b/task_templates/3_pipelines/15_python3_keras_vizai_joblib/model_utils.py
@@ -9,12 +9,19 @@ from typing import Tuple
 # keras imports
 import tensorflow
 from tensorflow.keras.models import Model, load_model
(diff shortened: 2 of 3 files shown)


task_templates/3_pipelines/15_python3_keras_vizai_joblib/model_utils.py
```diff
@@ -158,7 +151,10 @@ def get_image_augmentation_gen(X_data, y_data, bs, seed) -> Iterator[tuple]:
 
 def get_pretrained_base_model() -> Model:
     """ A base pretrained model to build on top of """
-    pretrained_model = MobileNetV2(include_top=False, input_shape=IMG_SHAPE, classes=NUM_CLASSES)
+    weights_file = "mobilenetv3_large.h5"
+    pretrained_model = MobileNetV3Large(
```
jasonjmcghee, line 155: Can we use `Small` instead? Should _dramatically_ improve runtime

task_templates/3_pipelines/15_python3_keras_vizai_joblib/model-metadata.yaml
```diff
@@ -19,6 +19,7 @@ typeSchema:
       condition: IN # can be one of IN or NOT_IN
       value: # can be one or multiple of NUM, TXT, IMG, DATE, CAT, DATE_DURATION, COUNT_DICT, GEO
         - IMG
+        - NUM
```
scottp-ml, line 22: We shouldn't make this change.  @drenfr01 lets talk tomorrow

scottp-ml (changes requested):

task_templates/3_pipelines/15_python3_keras_vizai_joblib/model-metadata.yaml
drenfr01: kk!

jasonjmcghee: FWIW I told @drenfr01 about this change - I could not get it to succeed until I made this change.

scottp-ml: From what I had seen the model was after another task that converted from img -> num.  I did a bunch of testing on the same dataset and the type schema was working correctly for me.

jasonjmcghee: If it's working it's working. Looks like all the models on that project got deleted, so I can't find the one that had that error that prompted us to make the change, but AFAIK there were no other tasks.

jasonjmcghee: Yeah I can't reproduce the problem anymore. I will take another crack at it another time, but if it works by uploading to DataRobot and running it (not just running with drum locally - does this ensure no network?) and it succeeds, I think I'm good with it

scottp-ml: An improvement we need to make is how we handle projects like that one in drum locally, but it is way beyond the scope of this ticket.

scottp-ml:

scottp-ml (approved):

jasonjmcghee (approved):

PR #514, opened 2021-12-13
Release v1.6.2-post1

yakov-g:
- bump log4j to 2.15.0

# This repository is public. Do not put here any private DataRobot or customer's data: code, datasets, model artifacts, .etc.

PR #516, opened 2021-12-14
Update envs with DRUM v1.6.2.post1

yakov-g:

## Summary
Backporting:
- update envs with DRUM v1.6.2.post1

### DIFF
--- a/public_dropin_environments/java_codegen/dr_requirements.txt
+++ b/public_dropin_environments/java_codegen/dr_requirements.txt
@@ -1,3 +1,3 @@
 pandas==1.0.5
 pyarrow==0.14.1
-datarobot-drum==1.6.2
+datarobot-drum==1.6.2.post1
--- a/public_dropin_environments/java_codegen/env_info.json
+++ b/public_dropin_environments/java_codegen/env_info.json
@@ -3,6 +3,6 @@
   "name": "[DataRobot] Java Drop-In (DR Codegen, H2O)",
   "description": "This template can be used as an environment for DataRobot generated scoring code or models that implement the either the IClassificationPredictor or IRegressionPredictor interface from the datarobot-prediction package and for H2O models exported as POJO or MOJO.",
   "programmingLanguage": "java",
-  "environmentVersionId": "61842ff66f7fd74deb73e820",
+  "environmentVersionId": "61b8b81e9f5641724b5bb0ae",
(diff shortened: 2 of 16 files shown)


public_dropin_environments/java_codegen/dr_requirements.txt
```diff
@@ -1,3 +1,3 @@
 pandas==1.0.5
 pyarrow==0.14.1
-datarobot-drum==1.6.2
+datarobot-drum==1.6.2.post1
```
klichukb, line 3: didnt notice its a dot. I think we thought of "1.6.2-post1"

@meltzg will this break anything?

yakov-g: no, it won't
When I use `-` python normalizes it to dot in the filename

yakov-g:
@klichukb @meltzg 
There is a drawback with tests.
When tests run, DRUM artifact is built in the branch and is used in the envs. So in these tests not `1.6.2.post1` will be used but `1.6.2`


'1.6.2.post1` has been used in here: https://github.com/datarobot/datarobot-user-models/pull/514

public_dropin_environments/java_codegen/dr_requirements.txt
klichukb: what do you normalized?

klichukb:
@yakov-g What is this behavior caused by?

yakov-g:
> 

This is normal behavior.
Think of this. 
Developer makes changes in DRUM, creates PR. New artifact is built and used in the env to test the changes(instead of pinned released DRUM) .

I this case, current PR is a bit weird.

yakov-g:
As it is not master, but kind of dead end branch, I can change code to run on 1.6.2.post1.

public_dropin_environments/java_codegen/dr_requirements.txt
yakov-g: https://www.python.org/dev/peps/pep-0440/#post-release-separators

it can be installed with either:
pip install datarobot-drum==1.6.2.post1
pip install datarobot-drum==1.6.2-post1

yakov-g:

klichukb (approved):

burkestar:
🪄 Mischief managed.  I signed off the failing PR check.

PR #533, opened 2022-01-04
[MODEL-7633][MODEL-7634] Drum functionality for new custom task template interface

jmbannon:

## Summary
This PR introduces a new way for users to define custom tasks, using the `CustomTask` class that inherits from one of the target type interfaces. 

Transform Example: https://github.com/datarobot/datarobot-user-models/pull/533/files#diff-cfd144cb0bfcec49f6ca83e453a6b5c91fa0e3ed3d54a08b7c1091470f273b16R12

Regression Example: https://github.com/datarobot/datarobot-user-models/pull/533/files#diff-1cdf3e1fdd46fe1591eb266c7b92cf0bc37925db2c3532865ccb93e35d99a3bfR15

For serializing to an artifact, we simply pickle the entire class instance. This will greatly simplify how users store state between fit and predict. If a class instance variable of theirs is not serializable, they can override the serializable methods to handle it.

## Changes
- Tests using the two new drum custom task interfaces (more to come in future PRs)
- Logic to handle new and legacy custom.py interfaces in model_adapter


custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -67,6 +69,19 @@ def __init__(self, model_dir, target_type=None):
...
+        self._custom_task_class_instance = None
+
+    @property
+    def is_custom_task_class(self):
+        """
+        Returns
+        -------
+        True if the code is using the CustomTask class interface. False if it's using the legacy drum hook interface.
+        """
+        return self._custom_task_class
```
meltzg, line 83: doesn't seem very bool to me

jmbannon: need to update to `is not None`

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -81,6 +96,40 @@ def _apply_sklearn_transformer(data, model):
             ) from None
         return output_data
 
+    def _load_custom_hooks_for_new_drum(self, custom_module):
```
4dahalibut, line 99: load class based and module based hooks.

jmbannon: i like that better, but mind if we do it in a follow-up so we don't add merge conflicts to @drenfr01 and I's subsequent PRs

custom_model_runner/datarobot_drum/drum/model_adapter.py
jmbannon: Fixed. Mind making another pass on this PR

drenfr01 (approved):
LGTM

meltzg (approved):

PR #534, opened 2022-01-05
[MODEL-7700] code dir cannot be the same as output dir for fit

scottp-ml:

## Summary
If a user sets the output dir the same as code-dir there is an error during fit that is not obvious to the user.  This change catches the issue earlier and provides a more actionable and obvious error message.  It does not make sense to allow users to use the code-dir as the output since the contents of code-dir will be copied to the output.  It is intended to create a custom model like package with the serialized model.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/drum.py
+++ b/custom_model_runner/datarobot_drum/drum/drum.py
@@ -490,11 +490,25 @@ class CMRunner:
             print(error_message)
             raise DrumCommonException(error_message)
 
+    def _validate_output_dir(self):
+        """Validate that the output directory is not the same as the input.
+                Raises
+        ------
+        DrumCommonException
+            Raised when the code directory is also used as the output directory.
+        """
+        if self.options.output == self.options.code_dir:
+            raise DrumCommonException(
+                "The code directory may not be used as the output directory."
+            )
+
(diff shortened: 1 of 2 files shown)


custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -509,6 +509,11 @@ def run_fit(self):
         if not self.options.output:
             self.options.output = mkdtemp()
             remove_temp_output = self.options.output
+        else:
+            if self.options.output == self.options.code_dir:
+                raise DrumCommonException(
+                    "The code directory may not be used as the output directory as well."
```
eric-s-s, line 515: can delete "as well".  the sentence reads better without it

eric-s-s: add to the "Raises" in the doc string, please

eric-s-s: 1. please extract to a private method with a descriptive name
2. this doesn't and shouldn't be an `else`.  always check

meltzg (approved):
I agree with @eric-s-s 's suggestion. Otherwise, g2g

eric-s-s (approved):
LGTM

PR #536, opened 2022-01-06
[RAPTOR-7531] update BRANCHING.yaml

yakov-g:


4dahalibut (approved):

meltzg (changes requested):
why? branching.yml is intended to create release artifacts for the public environments so we can have envs tied to a release of DR. This is not intended for drum releases.

PR #561, opened 2022-01-26
DRUM bump to 1.6.2.post6

klichukb:


custom_model_runner/datarobot_drum/drum/language_predictors/java_predictor/pom.xml
```diff
@@ -34,7 +34,7 @@
         <dependency>
             <groupId>org.apache.logging.log4j</groupId>
             <artifactId>log4j-core</artifactId>
-            <version>2.14.1</version>
+            <version>2.17.1</version>
```
klichukb, line 37: This is not what it looks like. Previous versions (2.16) were introduced for some reasons in separate branches in december, but only drum version update was merged back into release/7.3, fixing this to reflect arifacts that we shipped.

klichukb:
jarvis please test cmrunner_integration

4dahalibut (approved):

PR #565, opened 2022-02-01
[RAPTOR-7654] move R environment to use Python 3.7

yakov-g:

## Summary
- Make R evn to use Python 3.7 instead of 3.6
- unpin pyarrow dep in DRUM, so env defined version of pyarrow will be used

### DIFF
--- a/custom_model_runner/requirements.txt
+++ b/custom_model_runner/requirements.txt
@@ -15,7 +15,7 @@ scipy>=1.1,<2
 strictyaml==1.4.2
 texttable
 py4j~=0.10.9.0
-pyarrow==2.0.0
+pyarrow
 Pillow==8.3.2
 julia==0.5.6
 termcolor
--- a/custom_model_runner/requirements.txt
+++ b/custom_model_runner/requirements.txt
@@ -15,7 +15,7 @@ scipy>=1.1,<2
 strictyaml==1.4.2
 texttable
 py4j~=0.10.9.0
-pyarrow
+pyarrow==2.0.0
 Pillow==8.3.2
 julia==0.5.6
 termcolor
--- a/public_dropin_environments/r_lang/Dockerfile
+++ b/public_dropin_environments/r_lang/Dockerfile
@@ -2,13 +2,22 @@
 # It contains a variety of common useful data-science packages and tools.
 FROM datarobot/r-dropin-env-base
 
+RUN apt update -y && apt install -y software-properties-common && \
+    add-apt-repository -y ppa:deadsnakes/ppa && \
+    apt update -y && apt install -y python3.7 python3.7-dev
+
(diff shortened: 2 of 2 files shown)


public_dropin_environments/r_lang/Dockerfile
```diff
@@ -2,13 +2,22 @@
 # It contains a variety of common useful data-science packages and tools.
 FROM datarobot/r-dropin-env-base
 
+RUN apt update -y && apt install -y software-properties-common && \
+    add-apt-repository -y ppa:deadsnakes/ppa && \
+    apt update -y && apt install -y python3.7 python3.7-dev
+
+RUN python3.7 -m pip install setuptools wheel virtualenv && \
```
zohar-mizrahi, line 9: Can't you just use `python3 -m ...`? (instead of `python3.7 -m ...`)

yakov-g: At this stage default Py3.6 is still default...
So I first need to install virtualenv,  and then work in virtual env.

I consider this a temp solution, as we'll need to bump distro, python and R  versions for R env.

zohar-mizrahi (approved):

rvorobii (approved):

custom_model_runner/requirements.txt
```diff
@@ -15,7 +15,7 @@ scipy>=1.1,<2
 strictyaml==1.4.2
 texttable
 py4j~=0.10.9.0
-pyarrow==2.0.0
+pyarrow
```
4dahalibut, line 18: why did you unpin pyarrow

public_dropin_environments/r_lang/Dockerfile
4dahalibut: why do we need 3.7?

custom_model_runner/requirements.txt
yakov-g: We have different pyarrow versions pinned in the DRUM an in environments. This way there won't be version conflict.

4dahalibut: what happens if there is a conflict?

yakov-g: We use base envs which have older pip, which doesn't scream if there is conflict.
Now I created new virtual env, and there is a conflict. That's why I unpinned version in drum.

4dahalibut: I think we should leave it pinned. I could see this breaking things in the future

yakov-g: So I can pin it and degrade pip not to brake on conflicts.
Any how this is only temp measure, all this will be fixed as we need to create tests/envs for Py3.9

4dahalibut (approved):

yakov-g:

PR #566, opened 2022-02-01
[MODEL-7976] Create save/load_task helper functions

drenfr01:

## Summary

Two new helper functions, save_task and load_task, that handle pickling the CustomTask class and allow users to pass in objects, e.g. estimator, to ignore. We've modified the save and load methods in Serializable as well as the custom task templates to use these methods

## Rationale

To improve UX, we want to abstract away as much serialization as we can, even when the end user has to implement a custom serialization save and load hook, e.g. when saving a keras model. So we've introduced helper functions.


RodrigoGonzalez (approved):
LGTM

rvorobii (commented):
the implementation looks good.

regarding the naming:
cannot you just use `save`/`load` instead of `save_task`/`load_task`?
I see that the only thing that `save`/`load` do is call `save_task`/`load_task`

drenfr01:
> the implementation looks good.
> 
> regarding the naming: cannot you just use `save`/`load` instead of `save_task`/`load_task`? I see that the only thing that `save`/`load` do is call `save_task`/`load_task`

@rvorobii You are correct, but if you look at the example templates you'll see we split the function to allow abstraction for our end users. So they can overwrite the save method in the Serializable class, write their own custom serialization for their estimator (e.g. torch.save() for a pytorch model) and then use our save_task method to avoid having to write their own pickling code

PR #568, opened 2022-02-02
[RAPTOR-7659] make java_dropin_env_base use openjdk base image

yakov-g:
- created a new image in dockerhub datarobot/java-dropin-env-base:debian10-py3.7-java11
- make java_codegen env to use this image
- haven't done any optimizations to reduce image size (though it's 70 Mb less, than the previous base image)


docker/java_dropin_env_base/Dockerfile
```diff
@@ -1,42 +1,13 @@
 # This is the default base image for use with user models and workflows.
-FROM ubuntu:18.04
-ENV LC_ALL=en_US.UTF-8 TERM=xterm COLS=132 ROWS=43 DEBIAN_FRONTEND=noninteractive
+FROM openjdk:11.0.12-jre-slim-buster
```
zohar-mizrahi, line 2: Which OS is it based on?

docker/java_dropin_env_base/Dockerfile
```diff
@@ -1,42 +1,13 @@
...
+    pip3 install --upgrade pip && \
+    pip3 install --no-cache-dir wheel setuptools && \
+    # required for mlpiper
+    apt install -y python3-setuptools && \
+    chmod 707 /var/lib/nginx
 
 # do the trick to install drum deps
-RUN pip3 install --no-cache-dir datarobot-drum
-RUN pip3 uninstall -y datarobot-drum
+RUN pip3 install --no-cache-dir datarobot-drum && pip3 uninstall -y datarobot-drum
```
zohar-mizrahi, line 13: Don't we want to control which `datarobot-drum` version to install?

public_dropin_environments/java_codegen/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # This is the default base image for use with user models and workflows.
-FROM datarobot/java-dropin-env-base
+FROM datarobot/java-dropin-env-base:debian10-py3.7-java11
```
zohar-mizrahi, line 2: Is it really debian OS? Have we tested it and certified it? Is it documented anywhere?

docker/java_dropin_env_base/Dockerfile
yakov-g: `buster` is name for Debian 10

docker/java_dropin_env_base/Dockerfile
yakov-g: As this is base, I just install/uninstall DRUM to pull dependencies in order to potentially reduce build time when this base is used for actual envs.

zohar-mizrahi: Um ... but dependencies may be different depending on the drum version

public_dropin_environments/java_codegen/Dockerfile
yakov-g: We create this base so we know what is used. This is Debian

docker/java_dropin_env_base/Dockerfile
yakov-g: Yes. But then we actually install the latest DRUM in actual env. So if versions have been changed, they will be updated.

public_dropin_environments/java_codegen/Dockerfile
zohar-mizrahi: Have you ever tested Debian? I've never had experience with it. How stable it is, etc.

yakov-g: I haven't explicitly tested it. But we use it for legacy package images.
Quick google search says debian is more stable than ubuntu.

yakov-g:
So...
Our `h2o pojo` models have to be compiled, so base image should be based not on 
openjdk:11.0.12-jre-slim-buster

but on
openjdk:11.0.12-jdk-slim-buster

zohar-mizrahi (approved):

yakov-g:

yakov-g:

jmbannon (approved):

yakov-g:
> Talking to @4dahalibut , this is inference-only code. Please update codeowners. Removing the Core Modeling tag
Tag is assigned for all the public environments. Java is not used in training models. Do you want me to remove tag only for java env?
/public_dropin_environments                                         @datarobot/custom-models @datarobot/core-modeling

PR #573, opened 2022-02-03
Release v1.7.0

yakov-g:

## Summary
- release DRUM 1.7.0 - main addition is MLOps support
- rename base env `java_dropin_env_base` to `dropin_env_base`. This base has Java and Python which is required by MLOps, and is going to become the main base env
- inherit all py and java envs from  `datarobot/dropin-env-base:debian10-py3.7-jre11`; all the Py envs had and still have Py3.7; R env is still by it's own.
- fix pyarrow versions pinning as new base env has PIP version that checks versions compatibility

### DIFF
--- a/public_dropin_environments/python3_keras/Dockerfile
+++ b/public_dropin_environments/python3_keras/Dockerfile
@@ -1,6 +1,6 @@
 # This is the default base image for use with user models and workflows.
 # It contains a variety of common useful data-science packages and tools.
-FROM datarobot/python3-dropin-env-base
+FROM datarobot/dropin-env-base:debian10-py3.7-jre11
 
 # Install the list of core requirements, e.g. sklearn, numpy, pandas, flask.
 # **Don't modify this file!**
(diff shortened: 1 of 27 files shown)


RodrigoGonzalez (approved):
LGTM CORE MODELING

meltzg (approved):

RodrigoGonzalez (approved):
LGTM

4dahalibut (changes requested):
How much bigger are the new containers?

public_dropin_environments/python3_keras/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # This is the default base image for use with user models and workflows.
 # It contains a variety of common useful data-science packages and tools.
-FROM datarobot/python3-dropin-env-base
+FROM datarobot/dropin-env-base:debian10-py3.7-jre11
```
jmbannon, line 3: I feel like this will add a lot of unused bloat for custom (training) tasks. Do you know how much storage jre adds? Does it make sense to update all public envs to have this functionality?

yakov-g: https://hub.docker.com/r/datarobot/dropin-env-base/tags 
For py we've been using py3.7. slim, which was 150MB
This new image is not slim + jre, so it is 350 Mb.

As I understood, now we want all the envs to have MLOps support.

4dahalibut (approved):

PR #581, opened 2022-02-15
[MODEL-7979] Treat single col blank values as NAN in CSVs

jmbannon:

## Summary
When reading CSV input files, we should treat blank lines as NaNs instead of skipping over them. This does not include the last newline.

## Changes
- set skip_blank_lines = False when only 1 col is read
- unit test
- use existing CSV reading code in shared_fit_preprocessing

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/utils.py
+++ b/custom_model_runner/datarobot_drum/drum/utils.py
@@ -188,18 +188,9 @@ def shared_fit_preprocessing(fit_class):
         class_order: array specifying class order, or None
         row_weights: pd.Series of row weights, or None
     """
-    # read in data
-    if DrumUtils.endswith_extension_ignore_case(fit_class.input_filename, InputFormatExtension.MTX):
-        colnames = None
-        if fit_class.sparse_column_file:
-            colnames = [column.strip() for column in open(fit_class.sparse_column_file).readlines()]
-        df = pd.DataFrame.sparse.from_spmatrix(mmread(fit_class.input_filename), columns=colnames)
-    else:
-        df = pd.read_csv(fit_class.input_filename)
-        # If the DataFrame only contains a single column, treat blank lines as NANs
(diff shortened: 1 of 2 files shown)


custom_model_runner/datarobot_drum/drum/utils.py
```diff
@@ -348,8 +343,18 @@ def read_structured_input_data_as_df(binary_data, mimetype, sparse_colnames=None
                 df.fillna(value=np.nan, inplace=True)
 
                 return df
-            else:
-                return pd.read_csv(io.BytesIO(binary_data))
+            else:  # CSV format
+                df = pd.read_csv(io.BytesIO(binary_data))
+
+                # If the DataFrame only contains a single column, treat blank lines as NANs
+                if df.shape[1] == 1:
```
jasonjmcghee, line 350: is it possible for the shape to be 1d?

jmbannon: I don't think dataframes can be 1d

jmbannon: (this will always return a dataframe)

jasonjmcghee: yup you're good: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.shape.html

jasonjmcghee (commented):
You said _what_ in the PR description, but _why_ do we treat single col blank values as NaN?

jmbannon:
> You said _what_ in the PR description, but _why_ do we treat single col blank values as NaN?

Misha and a CFDS ran into this issue using drum locally. They thought that was the expected behavior.
It makes sense since if you have two cols, `,10` would translate to `nan,10`

jasonjmcghee (approved):

yakov-g (approved):

PR #593, opened 2022-03-14
[MODEL-8452] Organize unit tests

jmbannon:

## Summary
Breaks up the large `test_units.py` file, as well as move other unit tests laying around in various files under the `tests/drum/unit` directory into respective files that test specific parts of drum.

No code or tests were deleted, only moved/refactored/renamed. `test_fit_variety.py` did get deleted, but the tests were skipped with the comment: `This test has fallen into disrepair unfortunately` since 9/25/20. I think we can safely delete this one.

image

This should help development by being able to quickly run all tests in the unit test folder. We could even make this a separate test-suite that runs in PRs.


tests/drum/unit/model_metadata/test_model_metadata.py
```diff
@@ -853,3 +862,141 @@ def test_num_col_values(condition, value, fails):
...
+    assert isinstance(expected_as_str_list, list)
+
+    assert isinstance(expected_as_int, int)
+    assert isinstance(expected_as_str, str)
+
+
+@pytest.mark.parametrize(
+    "target_type, predictor_cls",
+    itertools.product([TargetType.TRANSFORM,], [PythonPredictor, RPredictor, JavaPredictor],),
+)
```
drenfr01, line 921: Interesting way to create a parameterized list! I hadn't seen that before. I wonder if Zach would approve :)

tests/drum/unit/test_language_predictors.py
```diff
@@ -0,0 +1,237 @@
+"""
+Copyright 2021 DataRobot, Inc. and its affiliates.
```
drenfr01, line 2: You're behind the times, man

tests/drum/unit/model_metadata/test_model_metadata.py
jmbannon: I personally prefer two separate `pytest.mark.parametrize`s

drenfr01 (approved):
Overall looks good to me! No need to restart the tests, but we should make a mental note to add the standard copyright to any new test files 

Also thank you for doing this! Particularly if / when we continue refactoring drum this will make a significant difference

rvorobii (approved):

PR #599, opened 2022-03-22
[MODEL-8487] Begin decoupling CLI arguments and fit logic

jmbannon:

## NOTE
- 618 lines of this PR are unit tests
- ~50 lines are license docstrings at the top of files
- Most of the other lines are moved code, will denote any significant changes in comments

## Summary
If we want users to ever use DRUM as a python module, we need to decouple CLI logic and drum fit/predict logic. This PR begins that work by decoupling the two for drum fit.

The class `DrumCLIAdapter` is introduced to take in CLI argument values and transform them into tangible python datatypes. The idea will be to use this as a translation layer between `cli` and `module` arguments. In the code itself, the 'modules' are model adapters, which hopefully one day we can make importable in the drum module.

`PythonModelAdapter` already exists, which only takes in a custom task path and target type. This PR also adds `RModelAdapter`, which we want to eventually make as similar to the python variant as possible - perhaps even inherit from the same class.

For drum fit specifically, `DrumCLIAdapter` will convert args/files/etc into X, y, row_weights, parameters. These are native python types that will be passed into custom tasks. `RModelAdapter` will be updated in the future to take these values as well. This will significantly help with code consolidation - we will be able to get rid of duplicate logic written in R in favor of `DrumCLIAdapter`'s implementation.

## Changes
- Add `DrumCLIAdapter` and `RModelAdapter` class (mostly moved from other files)
- Massive suite of unit tests for `DrumCLIAdapter`
- Refactor a few helper functions
- Add/update licenses for a few files
- Fix bug with R models where fit included the weights column in the dataframe
- Fix bug where we passed dense data to predict server when input fit data is sparse


custom_model_runner/datarobot_drum/drum/adapters/r/fit.R
```diff
@@ -134,16 +134,17 @@ process_weights <- function(X, weights_filename, weights, na_rows, sample_rows){
                   "is not one of the columns in your training data", sep=' '))
     }
     row_weights <- X[, weights, drop=FALSE]
+    X <- X[,!(colnames(X) %in% c(weights)), drop=FALSE]
```
jmbannon, line 137: We never separated row_weights from X in R, if X contained the row_weights column :grimacing:

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -560,23 +569,56 @@ def run_fit(self):
         DrumSchemaValidationException
             Raised when model metadata validation fails.
         """
-        self._prepare_fit()
+        cli_adapter = DrumCLIAdapter(
```
jmbannon, line 572: We use the DrumCLIAdapter to convert cli args (input filenames, target types, labels, etc) to dataframes, etc which will be passed to ModelAdapter. In the future, model adapter should act as a slim layer that takes in a custom task path + dataframes and runs fit/predict/score

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -613,19 +655,44 @@ def run_test_predict(self):
...
+                # TODO: Always force weights + target to be separate files with sparse input
+                if self.options.sparse_column_file:
+                    sparse_colnames = StructuredInputReadUtils.read_sparse_column_file_as_list(
+                        self.options.sparse_column_file
+                    )
+                    if self.options.row_weights in sparse_colnames:
+                        sparse_colnames.remove(self.options.row_weights)
+
+                        with open(self.options.sparse_column_file, "w") as file:
+                            file.write("\n".join(sparse_colnames))
```
jmbannon, line 682: Another case where we didn't drop row_weights from X before we performed inference :grimacing:

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -613,19 +655,44 @@ def run_test_predict(self):
...
             if self._get_fit_run_language() == RunLanguage.R:
                 df = handle_missing_colnames(df)
-            df.to_csv(__tempfile.name, index=False, line_terminator="\r\n")
+
+            if is_sparse_dataframe(df):
+                __tempfile = NamedTemporaryFile(suffix=".mtx")
+                mmwrite(__tempfile.name, df.sparse.to_coo())
+            else:
+                __tempfile = NamedTemporaryFile(suffix=".csv")
+                df.to_csv(__tempfile.name, index=False, line_terminator="\r\n")
```
jmbannon, line 693: `run_test_predict` somehow only passed dense data, and the sparse tests never failed

RodrigoGonzalez (approved):
LGTM CORE MODELING

custom_model_runner/datarobot_drum/drum/adapters/drum_cli_adapter.py
```diff
@@ -0,0 +1,303 @@
...
+        """
+        Returns
+        -------
+        pd.Series or None
+        """
+        if self.weights_filename:
+            weights = pd.read_csv(self.weights_filename, index_col=False).iloc[:, 0]
+        elif self.weights_name:
+            if self.weights_name not in self.input_dataframe.columns:
+                raise DrumCommonException(
```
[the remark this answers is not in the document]
jmbannon: Na I don't think so - printing errors is only needed for things that run in a separate process since we don't see the stacktrace

PR #604, opened 2022-03-24
[MODEL-8517] Use DrumCLIAdapter for drum score

jmbannon:

## Summary
This PR adds DrumCLIAdapter into the MLPiper drum score component for structured predicts. This helps consolidate how we read sparse column names to a single place. This makes it so we can pass down the sparse column names list instead of the binary data into the predictor. Drum server locations are updated to do the same.

## Rationale
The end-goal is to consolidate all shared logic between fit, score, and predict (both flask and werkzeug) to a single place, and to make the entrypoints to PythonModelAdapter/RModelAdapter take in dataframes/lists/etc instead of binary data. This will allow us to...
- Call predict without MLPiper / drum server
- Make drum importable as a module, and have users run fit/predict without using the CLI within their jupyter notebooks
  - Pass dataframes to their tasks to quickly test and iterate

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/adapters/drum_cli_adapter.py
+++ b/custom_model_runner/datarobot_drum/drum/adapters/drum_cli_adapter.py
@@ -84,7 +84,7 @@ class DrumCLIAdapter(object):
         """
         self.custom_task_folder_path = custom_task_folder_path
         self.target_type = target_type
-        self._input_filename = input_filename  # has setter property
+        self.input_filename = input_filename
         self.target_name = target_name
         self.target_filename = target_filename
         self.weights_name = weights_name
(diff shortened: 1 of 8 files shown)


custom_model_runner/datarobot_drum/drum/adapters/drum_cli_adapter.py
```diff
@@ -27,15 +27,15 @@
 
 class DrumCLIAdapter(object):
     """
-    This class acts as a layer between the DRUM CLI and running a custom task's fit method. It will convert the
-    cli arguments (mostly paths) into tangible variables to be passed into fit.
+    This class acts as a layer between DRUM CLI/MLPiper and running a custom task's fit/predict method. It will convert
```
jmbannon, line 30: I'll probably end up refactoring this class to have separate adapters for Fit/Score/Server that inherit from the same base class

RodrigoGonzalez (approved):
LGTM CORE MODELING

rvorobii (approved):

PR #614, opened 2022-03-31
[RAPTOR-3909] drop 2: unpin werkzeug, release DRUM 1.9.2

yakov-g:

## Summary
remove deprecated shutdown api

### DIFF
--- a/custom_model_runner/CHANGELOG.md
+++ b/custom_model_runner/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### [1.9.2] - 2022-04-01
 ##### Fixed
-- Unpin werkzeug
+- Pin werkzeug<2.2.0
 - Remove werkzeug shutdown API usage
 
 #### [1.9.1] - 2022-03-31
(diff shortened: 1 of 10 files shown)


rvorobii (approved):

custom_model_runner/requirements.txt
```diff
@@ -4,7 +4,6 @@ datarobot>=2.27.0
 trafaret!=1.1.0,<2.0,>=0.7
 docker>=4.2.2,<5.0.0
 flask
-werkzeug==2.0.3
```
akshoop, line 7: I would discourage this pin removal and instead suggest updating the hard pin. Based on the RAPTOR-3909 ticket description, I think the following pin would be better suited:
```
werkzeug>=2.1.1,<2.2.0
```

akshoop (commented):
I suggest instead of removing werkzeug pin to instead update the hard pin

custom_model_runner/requirements.txt
yakov-g: thanks...
I'll do `<2.2.0` just in case there are other issues

custom_model_runner/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+#### [1.9.2] - 2022-04-01
+##### Fixed
+- Unpin werkzeug
```
akshoop, line 9: I think this changelog note should be updated

akshoop (approved):
LGTM from Core Modeling

custom_model_runner/CHANGELOG.md
yakov-g: DONE

PR #626, opened 2022-05-13
Update Docs to reference removal of the target variable by DataRobot 

kindofluke:
## Summary
Documentation Change to Data Format section to mention that the target variable/column will be removed by DR. 

## Rationale

When models are tested in the Custom Model Registry, the target variable is removed from every dataset. This usually isn't a problem in most cases. However, some customers remove the target variable automatically through some other means like a SkLearn Column Transformer. DRUM has no way of knowing the data was changed prior to any hooks being invoked. This language in the docs will hopefully catch any errors.

See slack https://datarobot.slack.com/archives/CCE1XB7RU/p1652460660360309


elatt:
Closing due to inactivity.

PR #634, opened 2022-06-01
[MODEL-9100] Huggingface estimator template

scottp-ml:

## Summary
This adds an example estimator using the Huggingface library and the new Interface class.  

## Rationale
There are many people who are interested in using Huggingface in a custom task however doing so correctly is not obvious.  This provides a starting point with one of the more difficult situations since it requires preprocessing images data from b64 encoded strings first and then fine tuning the transformer.


rvorobii (approved):

scottp-ml:

scottp-ml:

PR #641, opened 2022-06-21
[MODEL-9340]

scottp-ml:

## Summary
Make sure that users are given actionable info when the typeschema is not valid.  For anything other than transforms output schema is not allowed, however the user was never given a clear message with how we handle the exception in fit.


gkoundry (approved):
LGTM Core Modeling

rvorobii (approved):

PR #657, opened 2022-06-29
[RAPTOR-8210] skip julia env functional test and move env

yakov-g:

## Summary
DRUM functional inference tests continue to fail on building Julia environment.
We don't have capacity to resolve this, and also Tim Whitaker - the one who wrote it and left, told me that no one uses it. (so likely Julia env can be removed)
I suggest to disable tests for now, and not update environment with  newer DRUM versions.

Also move this environment from `public_dropin_environments` folder to the `example_dropin_environments` to avoid installing it.


zohar-mizrahi (approved):

scottp-ml (approved):

meltzg (approved):

scottp-ml (approved):

yakov-g:

PR #660, opened 2022-07-05
[RAPTOR-8239] remove sagemaker sklearn extension from sklearn environment

yakov-g:

### DIFF
--- a/public_dropin_environments/python3_sklearn/env_info.json
+++ b/public_dropin_environments/python3_sklearn/env_info.json
@@ -1,8 +1,8 @@
 {
   "id": "5e8c889607389fe0f466c72d",
-  "name": "[DataRobot] Python 3 Scikit-Learn Drop-In",
+  "name": "[DataRobot] Python 3.9 Scikit-Learn Drop-In",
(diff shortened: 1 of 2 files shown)


eric-s-s (approved):

akshoop:
Ideally there will have to be a follow-up PR to update the various templates such as the sklearn model-template:
https://github.com/datarobot/datarobot-user-models/blob/309f55a6a7c7bf0c05827623b775d1afeda710e6/task_templates/3_pipelines/5_python3_sklearn_binary/create_pipeline.py#L9

yakov-g:
> Ideally there will have to be a follow-up PR to update the various templates such as the sklearn model-template:
> 
> https://github.com/datarobot/datarobot-user-models/blob/309f55a6a7c7bf0c05827623b775d1afeda710e6/task_templates/3_pipelines/5_python3_sklearn_binary/create_pipeline.py#L9

Btw, this example will work with tfidf vectorizer as only target field is text.

Let me do the cleanup in a separate PR

akshoop (approved):
LGTM from Core Modeling

PR #666, opened 2022-07-08
Making clear that classes need certain names

kindofluke:
## Summary
This is a small documentation change to the custom task tutorial which demonstrates how custom tasks can be created using just a `class` that inherits one of several interfaces. 

## Rationale
When writing my own custom task, I named it after what I was doing `CustomMultiClassSklearnModel` which seemed logical to me in my code base. I didn't think anything of this when I was writing the class but it would never load with DRUM. 

The reason is this code in `load_custom_hooks.py`

```python
            custom_module = __import__(CUSTOM_FILE_NAME)
            if getattr(custom_module, CUSTOM_PY_CLASS_NAME, None):
                self._load_custom_hooks_for_new_drum(custom_module)
            else:
                self._load_custom_hooks_for_legacy_drum(custom_module)
```
This means that the custom class must be named `CustomTask` in order for hooks to be loaded. It would be better to use logic to identify classes that subclass `CustomTaskInterface` but for now I just updated the documentation. 

I added the note in the image below. 
image


elatt:
Closing due to inactivity.

PR #668, opened 2022-07-14
[MODEL-9609]Point users to the official DR documentation for type schema

scottp-ml:

## Summary
The DataRobot docs site is the official documentation for custom tasks.  As such this PR removes the redundant documentation about type schema in order to prevent them from getting out of sync.


akshoop (approved):
LGTM from Core Modeling

rvorobii (approved):

PR #673, opened 2022-07-22
[MODEL-9648] ViT epochs as hyperparameter

scottp-ml:

## Summary
This adds epochs as a tunable hyperparameter for the ViT template.  Also I removed some unused imports from other templates.

### DIFF
--- a/task_templates/2_estimators/11_python_huggingface_vit/custom.py
+++ b/task_templates/2_estimators/11_python_huggingface_vit/custom.py
@@ -30,7 +30,13 @@ class DataSet(torch.utils.data.Dataset):
 
 
 class CustomTask(BinaryEstimatorInterface):
-    def fit(self, X, y, row_weights=None, **kwargs):
+    def fit(self, X, y, row_weights=None, parameters=None, **kwargs):
+        if parameters is None:
+            raise ValueError(
+                "Parameters were not passed into the custom task. "
+                "Ensure your model metadata contains hyperparameter definitions"
+            )
+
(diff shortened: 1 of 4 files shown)


task_templates/2_estimators/11_python_huggingface_vit/custom.py
```diff
@@ -30,7 +30,13 @@ def __len__(self):
 
 
 class CustomTask(BinaryEstimatorInterface):
-    def fit(self, X, y, row_weights=None, **kwargs):
+    def fit(self, X, y, row_weights=None, parameters=None, **kwargs):
```
yakov-g, line 33: If you want `parameters` to be required param, maybe make it positional t.e required?!

yakov-g: Actually,
maybe just pass `epochs` directly as `kwargs` and check it is not None. i.e.
```
epochs = kwargs.get("epochs")
if epochs is None:
   raise ValueError("epoch has to be provided in kwargs and can not be None")
```

scottp-ml: I believe all hyperparameters are added to parameters, they are not accessible individually.  This is passed by drum from reading the model-metadata

yakov-g: oh, I see, 
ok then

yakov-g (approved):

gkoundry (approved):
LGTM Core Modeling

PR #678, opened 2022-08-02
[MODEL-7482] support integer values as categorical values

scottp-ml:

## Summary
This allows integer values to be treated as categorical.  This happens in custom tasks when date derived features are used, which are passed along the categorical edge, but are integers.  
Additional change is to remove some unused numpy imports from task templates.


akshoop (approved):
LGTM from Core Modeling

rvorobii (approved):

PR #679, opened 2022-08-03
[MODEL-9684] log method added to custom task interface

scottp-ml:

## Summary
Printing debug or info in the predict step doesn't make it to the logs properly unless stdout is flushed.  This adds a simple utility to the interface class so that it is properly handled for the user and they don't need to know that detail.  This should make debugging tasks easier for users.

### DIFF
--- a/custom_model_runner/CHANGELOG.md
+++ b/custom_model_runner/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+#### [Current]
+##### Changed
+- added logging method to custom task interface
+
 #### [1.9.7] - 2022-08-02
 ##### Changed
 - pin flask<2.2.0 rpy2==3.5.2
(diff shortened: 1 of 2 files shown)


custom_model_runner/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+#### [Current]
```
akshoop, line 7: Do we want to keep version as "Current" or do we want to bump the version in:
https://github.com/datarobot/datarobot-user-models/blob/c4f07b60e875af3da50e25d18b0d3f2a864d4c48/custom_model_runner/datarobot_drum/drum/description.py#L7

akshoop (approved):
LGTM from Core Modeling with minor question

custom_model_runner/CHANGELOG.md
scottp-ml: I have two PRs open right now with changes, I will bump the version after both of those are in as its own PR.

akshoop: Sounds good thanks for the clarification 👍

rvorobii (approved):

PR #694, opened 2022-09-28
[RAPTOR-8525] Include DataRobot Python client in all relevant drop-in environments

zohar-mizrahi:

## Summary
In the context of [PBMP-5286](https://datarobot.atlassian.net/browse/PBMP-5286), it is required to include the DataRobot Python client in all python drop-in environments in other to allow the custom model's user code to use the DR client to access DataRobot API.

## Changes
* Add `datarobot==2.28.1` python package to the base drop-in environment
* Fix the README.md
* Fix the base image tag and reference it from all related images

## NOTE
**This PR can be merged only after the new base image (with the new tag) will be published in DockerHub.**

### DIFF
--- a/docker/dropin_env_base/README.md
+++ b/docker/dropin_env_base/README.md
@@ -1,16 +1,16 @@
 # Python/Java(JRE) drop-in environments base image
 Repository name: **datarobot/dropin-env-base**
-Latest date: 08-22-2022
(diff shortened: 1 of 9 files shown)


zohar-mizrahi:
Apparently the drum package is dependent on the DataRobot Python client (datarobot package). Therefore it is not required to add such dependency by the drop-in environments.

PR #698, opened 2022-10-15
[AGENT-3621] Adds support for extending DRUM server mode

elatt:

## Summary
Adds a new hook when the DRUM prediction server is starting up to look for a `custom_flask.py` file in the model-dir and if found, pass the `flask.Flask` application object to the `init_app` function in the found hook file. This allows the hook to customize the Flask application before we begin serving it.

I've also created a simple example in the template dir to illustrate how someone can use the new hook functionality to add simple authentication to the APIs DRUM exposes in server-mode.

## Rationale
The hook approach is familiar with users as they already use it to customize modeling tasks so now we have a similar way to allow users to customize the way DRUM server-mode behaves. The `init_app()` method name is a common function to call for extending a Flask application and `custom_flask.py` follows our naming conventions in this repo.


elatt:
jenkins run

elatt:
jenkins run

elatt:
jenkins run

elatt:

custom_model_runner/datarobot_drum/resource/components/Python/prediction_server/prediction_server.py
```diff
@@ -209,3 +212,24 @@ def terminate(self):
         terminate_op = getattr(self._predictor, "terminate", None)
         if callable(terminate_op):
             terminate_op()
+
+    def load_flask_extensions(self, app):
+        custom_file_paths = list(Path(self._code_dir).rglob("{}.py".format(FLASK_EXT_FILE_NAME)))
+        assert len(custom_file_paths) <= 1
```
scottp-ml, line 218: It would be nice to have a useful user facing message for this.

scottp-ml (approved):

custom_model_runner/datarobot_drum/resource/components/Python/prediction_server/prediction_server.py
elatt: Yeah, that makes sense. I copied the logic from the `custom.py` hook code and didn't think much about it but I also don't like using `assert` to catch bad user input. I'll go update that code as well: https://github.com/datarobot/datarobot-user-models/blob/master/custom_model_runner/datarobot_drum/drum/model_adapter.py#L139-L141

scottp-ml: Thanks

elatt:
Jenkins run test_integration_general

elatt:
Jenkins run test_integration_r_environment

rvorobii (commented):
I like the idea of being able to customize the drum webserver

I'm leaving this PR for @yakov-g  to review & approve 

my thoughts:
1. "flask" is an implementation detail. we used to support uwsgi + nginx in addition to flask, not sure what's the current status of this. so, we shouldn't mention flask, but be more generic instead. let's hear what @yakov-g thinks
2. we don't need a separate custom_flask.py file -- I'd just add `init_webserver_app` function to `custom.py`

rvorobii:
or maybe custom_flask.py is fine? let's hear from @yakov-g

yakov-g:
Regarding uwsgi + nginx: in this case Flask is wrapped deep inside mlpiper: https://github.com/datarobot/mlpiper/blob/1053748d804eac4e72ca0e775e5eab6f01ad55be/mlpiper-py/mlpiper/components/restful/uwsgi_broker.py#L310

So could you mention, that Flask customization is not available in `production` mode. 
You can do it in the following PR

yakov-g (approved):

PR #703, opened 2022-10-20
[RAPTOR-8303] fix tests, comments cleanup

yakov-g:

## Summary
Though tests in my prev PR passed and it was merged., following PRs started to fail.

I can not figure out the exact reason. Probably this is due to packages cross dependency, so change that to what it was before tests splitting.


yakov-g:

akshoop (approved):

scottp-ml (approved):

PR #729, opened 2023-01-26
[AGENT-3993] Update RuntimeParameters.get() to support default val

elatt:

## Summary
Adds a _default_ value to `RuntimeParameters.get()`. Also bumps the version of the package to a dev version in preparation for release.

## Rationale
Similar to the dict.get() method, I think it can be useful to have a mode that doesn't raise but instead allows a user to specify a default. The usecase I imagine is with credentials that may not always be required so a `RuntimeParameters.get('API_KEY', None)` call could be used.

### DIFF
--- a/custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
+++ b/custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
@@ -68,7 +68,9 @@ class RuntimeParameters:
         runtime_param_key = cls.namespaced_param_name(key)
         if runtime_param_key not in os.environ:
             if fallback is cls._UNSET:
-                raise ValueError(f"Runtime parameter '{key}' does not exist and no fallback provided!")
+                raise ValueError(
+                    f"Runtime parameter '{key}' does not exist and no fallback provided!"
+                )
             else:
                 return fallback
 
--- a/custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
+++ b/custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
@@ -38,22 +38,22 @@ class RuntimeParameters:
     _UNSET = object()
 
     @classmethod
-    def get(cls, key, default=_UNSET):
+    def get(cls, key, fallback=_UNSET):
         """
         Fetches the value of a runtime parameter as set by the platform. A ValueError is
-        raised if the parameter is not set unless if a default argument was provided.
+        raised if the parameter is not set and no fallback argument was provided.
 
         Parameters
         ----------
         key: str
             The name of the runtime parameter
-        default: ANY (optional)
+        fallback: ANY (optional)
             If specified, will be returned if no value has been set by the platform
 
 
         Returns
         -------
-        The value of the runtime parameter or the default (if specified)
+        The value of the runtime parameter or the fallback (if specified)
 
 
         Raises
@@ -67,10 +67,10 @@ class RuntimeParameters:
         """
         runtime_param_key = cls.namespaced_param_name(key)
         if runtime_param_key not in os.environ:
-            if default is cls._UNSET:
-                raise ValueError(f"Runtime parameter '{key}' does not exist!")
+            if fallback is cls._UNSET:
+                raise ValueError(f"Runtime parameter '{key}' does not exist and no fallback provided!")
             else:
-                return default
+                return fallback
 
         try:
             env_value = json.loads(os.environ[runtime_param_key])
--- a/custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
+++ b/custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
@@ -33,11 +33,44 @@ class RuntimeParameters:
 
     PARAM_PREFIX = "MLOPS_RUNTIME_PARAM"
 
+    # Used to determine if a user has specified a default or not since None is a valid
+    # user input.
+    _UNSET = object()
+
     @classmethod
-    def get(cls, key):
+    def get(cls, key, default=_UNSET):
+        """
+        Fetches the value of a runtime parameter as set by the platform. A ValueError is
+        raised if the parameter is not set unless if a default argument was provided.
+
+        Parameters
+        ----------
+        key: str
+            The name of the runtime parameter
+        default: ANY (optional)
+            If specified, will be returned if no value has been set by the platform
+
+
+        Returns
+        -------
+        The value of the runtime parameter or the default (if specified)
+
+
+        Raises
+        ------
+        ValueError
+            Raised when the parameter key was not set by the platform
+        InvalidJsonException
+            Raised if there were issues decoding the value of the parameter
+        InvalidRuntimeParam
+            Raised if the value of the parameter doesn't match the declared type
+        """
         runtime_param_key = cls.namespaced_param_name(key)
         if runtime_param_key not in os.environ:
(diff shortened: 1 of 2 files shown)


elatt:
jenkins run test_integration_r_environment test_integration_general

elatt:

akshoop:

custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
```diff
@@ -33,11 +33,44 @@ class RuntimeParameters:
 
     PARAM_PREFIX = "MLOPS_RUNTIME_PARAM"
 
+    # Used to determine if a user has specified a default or not since None is a valid
+    # user input.
+    _UNSET = object()
+
     @classmethod
-    def get(cls, key):
+    def get(cls, key, default=_UNSET):
```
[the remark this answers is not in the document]
eric-s-s: since a default is required if the key is missing otherwise error, perhaps two methods:

- `get_or_default(key, default)`
- `strict_get(key)`

the names are first off the top of my head, but you get the point.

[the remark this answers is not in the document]
rvorobii: > since a default is required

do we have to make the default required? I'd keep the dict semantics of `get` + `[]` if possible.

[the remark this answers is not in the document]
elatt: This is basically the same behavior as `ConfigParser.get()` in the stdlib: If you ask for an option that isn't found in the config file and you didn't specify a `fallback` option the it will raise `NoOptionError`, otherwise it will return the fallback. https://github.com/python/cpython/blob/3.11/Lib/configparser.py#L781

If we don't want to move forward with this approach, I'd say I'm just going to close this PR out and wait for more user feedback.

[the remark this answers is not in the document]
rvorobii: good point about being consistent with `ConfigParser.get()`.

rvorobii (approved):

custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
```diff
@@ -33,11 +33,44 @@ class RuntimeParameters:
 
     PARAM_PREFIX = "MLOPS_RUNTIME_PARAM"
 
+    # Used to determine if a user has specified a default or not since None is a valid
```
eric-s-s, line 36: do we need the comment?

custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
```diff
@@ -33,11 +33,44 @@ class RuntimeParameters:
...
+    # user input.
+    _UNSET = object()
+
     @classmethod
-    def get(cls, key):
+    def get(cls, key, default=_UNSET):
+        """
+        Fetches the value of a runtime parameter as set by the platform. A ValueError is
+        raised if the parameter is not set unless if a default argument was provided.
```
eric-s-s, line 44:
```suggestion
        raised if the parameter is not set and no default argument was provided.
```

custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
```diff
@@ -33,11 +33,44 @@ class RuntimeParameters:
...
+        InvalidJsonException
+            Raised if there were issues decoding the value of the parameter
+        InvalidRuntimeParam
+            Raised if the value of the parameter doesn't match the declared type
+        """
         runtime_param_key = cls.namespaced_param_name(key)
         if runtime_param_key not in os.environ:
-            raise ValueError(f"Runtime parameter '{key}' does not exist!")
+            if default is cls._UNSET:
+                raise ValueError(f"Runtime parameter '{key}' does not exist!")
```
eric-s-s, line 71:
```suggestion
                raise ValueError(f"Runtime parameter '{key}' does not exist and not default/fallback provided!")
```

custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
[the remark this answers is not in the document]
eric-s-s: If you want to follow ConfigParser.get, which seems fine, then maybe change the kwarg to `fallback`?  also, update the error.  the issue i see with `get` is that many users will first assume `dict.get` where `None` is the default fallback.  so maybe the method needs a more descriptive name?  :shrug:

eric-s-s (approved):
a few docstring nits, otherwise :+1:

custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
```diff
@@ -33,11 +33,44 @@ class RuntimeParameters:
...
+    def get(cls, key, default=_UNSET):
+        """
+        Fetches the value of a runtime parameter as set by the platform. A ValueError is
+        raised if the parameter is not set unless if a default argument was provided.
+
+        Parameters
+        ----------
+        key: str
+            The name of the runtime parameter
+        default: ANY (optional)
```
elatt, line 50:
```suggestion
        fallback: ANY (optional)
```

custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
elatt, line 41:
```suggestion
    def get(cls, key, fallback=_UNSET):
```

custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
```diff
@@ -33,11 +33,44 @@ class RuntimeParameters:
...
+            If specified, will be returned if no value has been set by the platform
+
+
+        Returns
+        -------
+        The value of the runtime parameter or the default (if specified)
```
elatt, line 56:
```suggestion
        The value of the runtime parameter or the fallback (if specified)
```

custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
```diff
@@ -33,11 +33,44 @@ class RuntimeParameters:
...
+            Raised when the parameter key was not set by the platform
+        InvalidJsonException
+            Raised if there were issues decoding the value of the parameter
+        InvalidRuntimeParam
+            Raised if the value of the parameter doesn't match the declared type
+        """
         runtime_param_key = cls.namespaced_param_name(key)
         if runtime_param_key not in os.environ:
-            raise ValueError(f"Runtime parameter '{key}' does not exist!")
+            if default is cls._UNSET:
```
elatt, line 70:
```suggestion
            if fallback is cls._UNSET:
```

custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
```diff
@@ -33,11 +33,44 @@ class RuntimeParameters:
...
+                raise ValueError(f"Runtime parameter '{key}' does not exist!")
+            else:
+                return default
```
elatt, line 73:
```suggestion
                return fallback
```

yakov-g:
jenkins run test_integration_java_environment

elatt:
jenkins run 0-all-check-status

PR #732, opened 2023-01-30
[AGENT-4009] Fix some bugs found during testing RuntimeParams

elatt:

## Summary
Update the way `setup.py` imports code from the target module to only import the modules directly needed. Also fix the schema for the `model-metadata.yaml` file to support runtime parameters. Finally, make sure production code uses `strictyaml` as that is used elsewhere and since we don't directly require pyaml, we can't rely on it being installed.

## Rationale
The root module of the `datarobot_drum` package now imports `RuntimeParameters` class which requires `trafaret` and `strictyaml` but it seems like a waste to list those as `setup_requires` when instead we can just make the setup script more stand-alone.

### DIFF
--- a/custom_model_runner/setup.py
+++ b/custom_model_runner/setup.py
@@ -9,24 +9,18 @@ from pathlib import Path
 from setuptools import setup, find_packages
 
 
-def direct_import(path):
-    # Direct imports are needed because datarobot_drum/__init__.py imports other modules
-    # that depend on 3rd party libraries and `setup.py` **must** be able to run in a blank
-    # virutalenv. This code snippet was adapted from and simply loads the module directly
-    # without loading all its parents:
-    #   https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly
-    module_path = path.resolve()
-    spec = importlib.util.spec_from_file_location(module_path.stem, module_path)
-    module = importlib.util.module_from_spec(spec)
-    spec.loader.exec_module(module)
-    return module
-
-
 # The directory containing this file
 root = Path(__file__).absolute().parent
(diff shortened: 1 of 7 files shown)


yakov-g:
jenkins run test_integration_r_environment test_integration_general

elatt:

yakov-g (approved):

elatt:
jenkins run test_integration_general test_integration_r_environment

akshoop:

elatt:

custom_model_runner/setup.py
```diff
@@ -4,34 +4,49 @@
...
 """
+import importlib.util
+from pathlib import Path
 from setuptools import setup, find_packages
-import os
 
-from datarobot_drum.drum.description import version, project_name
-from datarobot_drum.drum.enum import SupportedFrameworks, extra_deps
+
+def direct_import(path):
```
eric-s-s, line 12: how did this work before?

eric-s-s (approved):

elatt:
Jenkins run test_integration_general

custom_model_runner/setup.py
elatt: https://github.com/psf/requests/blob/15585909c3dd3014e4083961c8a404709450151c/setup.py#L78

Hmm maybe that is a little more elegant. Since I still have a failing test I may adapt our code.

yakov-g:
jenkins run this

yakov-g:
jenkins run Jenkins run test_integration_general

yakov-g:
jenkins run ontinuous-integration/jenkins/pr-merge

PR #733, opened 2023-01-30
[RAPTOR-9013] update drum for Julia environment

yakov-g:

## Summary
Update drum in the Julia environment.
We removed Julia environment from the public environment, as we don't have anyone specialized in Julia, but the env is still available in the examples.

Older drum doesn't pin flask version, so there will be problems if someone decides to play with the env.


yakov-g:

yakov-g:
jenkins run test_integration_java_environment

klichukb (approved):

rvorobii (approved):

PR #736, opened 2023-02-06
[AGENT-4034] Improve how we inject runtime params during local testing

elatt:

## Summary
- Bump version to 1.10.0rc1
- No longer convert credential values to snake_case
- Remove fallback argument from `get()`
- Make sure all defined params have env var injected

## Rationale
This change makes the local exprience more in line with the behavior when run in the MLOps custom model infrastructure:
- we no longer snake_case the credentials payload so we shouldn't do that locally either
- we no longer omit params that have a `None` value so it is pointless to have a fallback in the `get()` function
- we now will utilize the default value from the definition when running locally if no value exists in the values file

### DIFF
--- a/custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
+++ b/custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
@@ -6,7 +6,6 @@ Released under the terms of DataRobot Tool and Utility Agreement.
 """
 import json
 import os
-import re
 from collections import namedtuple
 
 import trafaret as t
@@ -118,8 +117,8 @@ class RuntimeParametersLoader:
         try:
             with open(values_filepath, encoding="utf-8") as file:
                 self._yaml_content = yaml.safe_load(file)
-                if not self._yaml_content:
-                    raise InvalidEmptyYamlContent("Runtime parameter values YAML file is empty!")
+            if not self._yaml_content:
+                raise InvalidEmptyYamlContent("Runtime parameter values YAML file is empty!")
         except yaml.YAMLError as exc:
             raise InvalidYamlContent(f"Invalid runtime parameter values YAML content! {str(exc)}")
         except FileNotFoundError:
@@ -131,6 +130,8 @@ class RuntimeParametersLoader:
         try:
             with open(os.path.join(code_dir, MODEL_CONFIG_FILENAME)) as file:
                 model_metadata = yaml.safe_load(file)
+            if not model_metadata:
+                raise InvalidEmptyYamlContent("Model-metadata YAML file is empty!")
         except yaml.YAMLError as exc:
             raise InvalidYamlContent(f"Invalid model-metadata YAML content! {str(exc)}")
         except FileNotFoundError:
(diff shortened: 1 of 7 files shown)


elatt:

akshoop:

custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
```diff
@@ -109,57 +101,80 @@ class RuntimeParametersLoader:
...
-    def __init__(self, values_filepath):
+    ParameterDefinition = namedtuple("ParameterDefinition", ["name", "type", "default"])
+
+    def __init__(self, values_filepath, code_dir):
         if not values_filepath:
             raise InvalidInputFilePath("Empty runtime parameter values file path!")
+        if not code_dir:
+            raise InvalidInputFilePath("Empty code-dir path!")
+
+        self._load_parameter_definitions(code_dir)
```
zohar-mizrahi, line 115: My understanding was that for local development the user does not have to provide a parameter definitions file, to ease his development. However, I do agree that is it much more sustainable and robust to require it.

zohar-mizrahi (approved):

custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
elatt: Hmm yeah, that is a good point that it makes things much simpler only have to worry about one file. However, _when I did_ have both files in place when I was doing some local dev, I was tripped up by the fact that my defaults defined in the master file were not being applied as they would when actually deployed on DataRobot.

So I guess I would feel strongly that _if_ the `model-metadata.yaml` file was present, then we should utilize it for local testing; however, I could see an argument for if the file didn't exist, then we can just rely on the values file alone. But, this is a little moot because the user will need to define their params eventually so why not just make them do it up front? I've only made two models so far with runtime params so I don't have a good sense of what the ideal workflow would be yet🤔  I think it is easier to have it required and later relax that restriction than have it relaxed and later have it required and break some peoples workflows so I'm leaning towards this way for now.

custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters.py
```diff
@@ -109,57 +101,80 @@ class RuntimeParametersLoader:
...
+                data = RuntimeParameterDefinitionTrafaret.check(parameter)
+            except t.DataError as exc:
+                raise ErrorLoadingRuntimeParameter(f"Failed to load runtime parameter: {str(exc)}")
+            self._parameter_definitions[data["name"]] = self.ParameterDefinition(**data)
+
     def setup_environment_variables(self):
         credential_payload_trafaret = RuntimeParameterCredentialPayloadTrafaret()
         string_payload_trafaret = RuntimeParameterStringPayloadTrafaret()
-        for param_key, param_value in self._yaml_content.items():
+        for param_key, param_definition in self._parameter_definitions.items():
```
eric-s-s, line 159: so this means that all parameter definitions always have a default value, correct?  is it possible for the user's model metadata to have no default set?  if so, is it None?

elatt: Yup, the default _default_ 😉 is `None`

eric-s-s: :scream:

eric-s-s (approved):

PR #756, opened 2023-04-11
[RAPTOR-9684] bump pandas<=2.0.3

yakov-g:

## Summary
As per GenAI team request, bump pandas limit to <=2.0.3
All the envs will still have pandas 1.5.3 by default as it is installed in the base images. 
But I checked that all the tests pass if explicitly set pandas to 2.0.3 in the envs.
@mxpoliakov


mxpoliakov:
@yakov-g thank you very much! If the `pandas` update went without issues, we could probably unpin `pyarrow` (latest version 12.0.1). Also, it would be interesting to try updating `Pillow` to 10.0.0.

rvorobii (approved):

akshoop (approved):
LGTM from Core Modeling, CC @scottp-ml @eric-s-s for awareness

PR #758, opened 2023-04-12
[MODEL-11091]bug fix for dropped rows in R during predict/transform

scottp-ml:

## Summary
Empty rows are dropped from the dataset during transform step in R.


yakov-g:
jenkins run test_integration_general test_integration_python_keras

PR #776, opened 2023-05-31
[RAPTOR-9417] Upgrade dr-usertool to v3

rvorobii:
The latest dr-usertool requires users to be a part of an organization.

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -44,12 +44,18 @@ def pytest_configure(config):
         admin_api_key = get_admin_api_key()
 
         # User credentials
+        org_name = "local-custom-model-tests-org-{}".format(suffix)
         user_username = "local-custom-model-tests-{}@datarobot.com".format(suffix)
         user_password = "Lkjkljnm988989jkr5645tv_{}".format(suffix)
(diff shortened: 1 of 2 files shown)


zohar-mizrahi (approved):

tests/functional/conftest.py
```diff
@@ -51,6 +51,17 @@ def pytest_configure(config):
...
         )
 
+        # Add organization
+        org_name = "local-custom-model-tests-org"
+        try:
+            # ensure it exists
+            DataRobotUserDatabase.add_organization(environment=env, organization_name=org_name)
+        except ValueError:
+            # already exists
+            pass
```
JarheadStever, line 61: Hey @rvorobii I saw this pattern used elsewhere too, but I think that catching the ValueError here is dangerous and unnecessary. `add_organization()` should return `true` even if the organization already exists, but will raise a `ValueError` if there is no admin API key. This was a change I made before the 3.0.0 release. datarobot/dr-usertool@c6c4143#diff-fe01c37518f26141ae7e0ca068f497f8e44cc2f1e1071195533c2325412c0be5R588-R592

PR #795, opened 2023-07-24
[MODEL-11933] fixed misleading tests

eric-s-s:

## Summary

While reviewing code, I found tests that claim to test things that they don't actually test and aren't in the code base.  so i fixed the tests.

The basic problem is that the original tests went in a `for` loop.  instead of testing each missing key in isolation, they tested an ever decreasing dict

original: 

```
payload = {
            "credentialType": "s3",
            "region": "us-west",
            "awsAccessKeyId": "123aaa",
            "awsSecretAccessKey": "3425sdd",
             "awsSessionToken": "123aaa",
        }
```

before first test in loop

```
payload = {
            "region": "us-west",
            "awsAccessKeyId": "123aaa",
            "awsSecretAccessKey": "3425sdd",
             "awsSessionToken": "123aaa",
        }
```

then

```
payload = {
            "region": "us-west",
            "awsSecretAccessKey": "3425sdd",
             "awsSessionToken": "123aaa",
        }
```

so each loop correctly fails not because of missing key, but because `"credentialType"` is not present

### DIFF
--- a/tests/drum/unit/test_runtime_parameters.py
+++ b/tests/drum/unit/test_runtime_parameters.py
@@ -28,15 +28,7 @@ class TestRuntimeParameters:
         "runtime_param_type, payload",
         [
             (RuntimeParameterTypes.STRING, "Some string value"),
-            (
-                RuntimeParameterTypes.CREDENTIAL,
-                {
-                    "credentialType": "s3",
-                    "awsAccessKeyId": "123aaa",
-                    "awsSecretAccessKey": "3425sdd",
-                    "awsSessionToken": "12345abcde",
-                },
-            ),
+            ( RuntimeParameterTypes.CREDENTIAL, { "credentialType": "askdfjsdk" }),
             (
                 RuntimeParameterTypes.CREDENTIAL,
                 {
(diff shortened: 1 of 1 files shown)


tests/drum/unit/test_runtime_parameters.py
```diff
@@ -28,15 +28,7 @@ class TestRuntimeParameters:
...
-            (
-                RuntimeParameterTypes.CREDENTIAL,
-                {
-                    "credentialType": "s3",
-                    "awsAccessKeyId": "123aaa",
-                    "awsSecretAccessKey": "3425sdd",
-                    "awsSessionToken": "12345abcde",
-                },
-            ),
+            ( RuntimeParameterTypes.CREDENTIAL, { "credentialType": "askdfjsdk" }),
```
klichukb, line 31: Is this a valid runtime param?

eric-s-s: yes-ish.  this is a valid schema, and the code will pass, but it's not actually a valid param.

klichukb: So why the change? s3 also seemed fine-ish? :)

scottp-ml (approved):

tests/drum/unit/test_runtime_parameters.py
eric-s-s: 1. we have nice s3 examples elsewhere in the same test file
2. a garbage parameter in the test helps show that the code allows arbitrary strings. 

the only thing required for a credential env var is that it has a non-empty credential type.  this test demonstrates the bare minimum and answers the question I had, which was, "What are the actual code requirements for this piece of code".  When I saw the original test and looked at the code, it _seemed_ like anything was ok,  but there was no test confirming that.  then there was also the misleading test that seemed to say the code did much more than it actually could.  I'd like to be able to read the test and quickly determine what the limits are.  an arbitrary value does this better than some expected value that isn't actually checked by the code

zohar-mizrahi: It was intentionally decided to not to enforce any credential string/type at the drum level for future proof. (cc @elatt )

zohar-mizrahi (approved):

tests/drum/unit/test_runtime_parameters.py
elatt: Yes, since drum is sort of disconnected from datarobot, I felt it best to be loosely coupled and not having to chase after DR every time a new credential type was added.

That being said, I don't see any value in tests having garbage data in them. I'm of the belief tests should have values as close the real thing as reasonable. You're testing an implementation detail... we decided (for now) to allow arbitrary strings so when some other team adds a new credential type, it doesn't break drum right away but I feel we should be free to change this.

elatt (changes requested):
I don't see value in this change as is. We shouldn't be asserting that we allow a garbage value. We should only be asserting that we allow good values that come from DR. I think that's the only guarantee we provide.

The for-loop stuff is a little confusing and agree we can do without.

tests/drum/unit/test_runtime_parameters.py
eric-s-s: done

eric-s-s:
jarvis please rerun failed

elatt (approved):
Thanks, LGTM 👍

eric-s-s:
jarvis test test_integration_general

PR #803, opened 2023-08-04
[RAPTOR-9675] base GenAI env on Python 3.11

yakov-g:

## Summary
* Create new base environment: UBI8.8 + Java11(JDK) + Py3.11 + all the required packages. Decision to use UBI came from that it has Java11 and Py3.11 packages, so no need to build from source, which results in 1GB less image that fro Debian11. (Also UBI is used widely across DR)
* rename `python3_genai` -> `python3_genai_39`
* Create a separate GenAI environment that uses Python3.11, named it EXPERIMENTAL
* Update DRUM to 1.10.9 in both GenAI py3.9 and Py3.11 envs.

Also I temporary switched all the environments to use new UBI Py3.11 image, and ran tests.
 70% tests passed. Some tests failed because of deps conflict or/and artifacts that require update because of deps bump.
As a result I think new UBI+Py3.11 image is in a good shape to be a base for envs update to Py3.11. Some envs will require packages tune up.

@mxpoliakov


yakov-g:
jenkins run test_integration_r_environment test_inference_model_templates

docker/dropin_env_base/Dockerfile
```diff
@@ -3,7 +3,14 @@ FROM openjdk:11.0.16-jre-slim-bullseye
...
 
-RUN apt update && apt upgrade -y && apt install -y python3 python3-pip python3-dev python3-distutils fontconfig nginx && \
+RUN apt update && apt upgrade -y && apt install -y wget build-essential libncursesw5-dev libssl-dev libsqlite3-dev tk-dev libgdbm-dev libc6-dev libbz2-dev libffi-dev zlib1g-dev fontconfig nginx && \
+    cd /tmp/ && \
+    wget https://www.python.org/ftp/python/3.11.4/Python-3.11.4.tgz && \
+    tar -xzvf Python-3.11.4.tgz && \
+    cd Python-3.11.4/ && \
+    ./configure --enable-optimizations && \
+    make -j `nproc` && \
+    make install && \
```
zohar-mizrahi, line 13: @yakov-g do you know if installing Python from source code results in a bigger image size? If so, what's the difference? (If there is a significant difference, I'd recommend using the multi-stage build:
https://docs.docker.com/build/building/multi-stage/#name-your-build-stages)

yakov-g: Yes, new image is 1.1GB bigger. 
It's probably not only because of python but all the toolchain.

So how stages should help? Every stage would be just a separate image, which doesn't nee to be rebuilt?

Also, do you want me to use stages in this PR? Base image is already pushed to dockerhub.

yakov-g: Eventually, after installin Py3.11 from source images became 7.3Gb vs 6.3 with py 3.9.

So I implemented py3.11 installation baed on the UBI image. it is ~6.4GB

yakov-g:
jenkins run test_integration_general test_integration_r_environment

docker/dropin_env_base_jdk_ubi/build_image.sh
```diff
@@ -0,0 +1,23 @@
...
+#
+# All rights reserved.
+# This is proprietary source code of DataRobot, Inc. and its affiliates.
+#
+# Released under the terms of DataRobot Tool and Utility Agreement.
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+IMAGE_NAME=datarobot/dropin-env-base-jdk
+IMAGE_TAG=ubi8.8-py3.11-jdk11.0.20-drum1.10.9-mlops9.1.3
```
zohar-mizrahi, line 12: No for this PR, but maybe consider having all the related versions in a single file and read the versions in all required places. This way, you'll always make a change in one place only.

public_dropin_environments/python3_genai_py311/Dockerfile
zohar-mizrahi, line 1: I'd recommend to rename it to `python311_genai`.

yakov-g:
jenkins run test_integration_general test_integration_python_keras test_integration_r_environment

docker/dropin_env_base_jdk_ubi/build_image.sh
yakov-g: ok

public_dropin_environments/python3_genai_py311/Dockerfile
yakov-g: DONE

yakov-g:

zohar-mizrahi (approved):

PR #824, opened 2023-09-28
[BUZOK-454] Sync 3.11 GenAI env with DataRobot Buzok

mxpoliakov:

## Summary
This PR adds dependencies needed for the DR GenAI project:

- Update `langchain`
- Adds `optimum`, `onnx`, `onnxruntime` to support optimized embeddings. 
- Adds `google-cloud-aiplatform` and `aws-request-signer` to enable working with Amazon and Google LLMs. 
- Adds `pydantic` and `aiofiles` since they are used in DR-created custom models.


yakov-g:
What about 3.9 genai env. Do we want to keep it? Should we update it?

yakov-g (approved):

mxpoliakov:
> What about 3.9 genai env. Do we want to keep it? Should we update it?

We don't use it for DR GenAI, so there is no need to update it.

mxpoliakov:
@yakov-g it requires 2 reviews, can we get someone to sign it off?

aslisabanci (approved):

PR #841, opened 2023-10-25
[RAPTOR-9976] clean exception handling logic

yakov-g:

## Summary
So,
the original error
 https://datarobot.slack.com/archives/CCE1XB7RU/p1697809688880589?thread_ts=1697808116.835149&cid=CCE1XB7RU
says: TypeError: No constructor defined.
Which is probably error in the implementation of the exception we are trying to re-throw.

So instead of re-throwing with an alternative message,   log and re-throw the original one.

Probably I should change all the similar cases in this module.

@rvorobii @zohar-mizrahi @eric-s-s 
what do you think?

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/model_adapter.py
+++ b/custom_model_runner/datarobot_drum/drum/model_adapter.py
@@ -10,6 +10,7 @@ import sys
 import textwrap
 from inspect import signature
 from pathlib import Path
+from typing import NoReturn
 
 import pandas as pd
 from scipy.sparse import issparse
@@ -43,6 +44,7 @@ from datarobot_drum.drum.enum import (
 )
 from datarobot_drum.drum.exceptions import (
     DrumCommonException,
+    DrumException,
     DrumTransformException,
     DrumSerializationError,
 )
@@ -51,6 +53,10 @@ from datarobot_drum.drum.utils.drum_utils import DrumUtils
 from datarobot_drum.custom_task_interfaces.custom_task_interface import CustomTaskInterface
 
 RUNNING_LANG_MSG = "Running environment language: Python."
+
+
+class DrumPythonModelAdapterError(DrumException):
+    """Raised in case of error in PythonModelAdapter"""
 
 
 class PythonModelAdapter:
@@ -79,6 +85,10 @@ class PythonModelAdapter:
         self._custom_task_class = None
         self._custom_task_class_instance = None
 
+    def _log_and_raise_final_error(self, exc: Exception, message: str) -> NoReturn:
+        self._logger.error(f"{message} Exception: {exc!r}")
+        raise DrumPythonModelAdapterError(f"{message} Exception: {exc!r}")
+
     @property
     def is_custom_task_class(self):
         """
@@ -100,6 +110,7 @@ class PythonModelAdapter:
             raise type(e)("Couldn't naively apply transformer:" " {}".format(e)).with_traceback(
                 sys.exc_info()[2]
             ) from None
+
         return output_data
 
     def _load_custom_hooks_for_new_drum(self, custom_module):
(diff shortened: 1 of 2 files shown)


yakov-g:

akshoop:
Adding back core modeling review label and asking Eric to take a look

eric-s-s (changes requested):
I _like_ this change!

Requested change: Once we've settled on a way forward, please fix for all the instances in this file.  a quick git grep showed that this is the only file where we have this pattern.  this might also be worth pulling  out to some method. so that future change will be easier.

```python
log_and_raise_final_error(exc: Exception, message: str) -> NoReturn: ...
```

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -520,9 +520,10 @@ def _predict_legacy_drum(self, data, model, **kwargs):
                 # noinspection PyCallingNonCallable
                 predictions_df = self._custom_hooks.get(CustomHooks.SCORE)(data, model, **kwargs)
             except Exception as exc:
-                raise type(exc)(
+                self._logger.error(
                     "Model score hook failed to make predictions. Exception: {}".format(exc)
-                ).with_traceback(sys.exc_info()[2]) from None
+                )
+                raise
```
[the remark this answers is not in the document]
rvorobii: from user perspective it's important to have a traceback so that user can debug the model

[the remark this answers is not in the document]
eric-s-s: right.  when you raise from another error, it gives the traceback

```python
def do_thing():
    raise ValueError("oops")

try:
    do_thing()
except ValueError:
    raise RuntimeError("well that didn't work")
```

output

```
Traceback (most recent call last):
  File "/home/eric.shaw/workspace/DataRobot/_delme.py", line 5, in <module>
    do_thing()
  File "/home/eric.shaw/workspace/DataRobot/_delme.py", line 2, in do_thing
    raise ValueError("oops")
ValueError: oops

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/eric.shaw/workspace/DataRobot/_delme.py", line 7, in <module>
    raise RuntimeError("well that didn't work")
RuntimeError: well that didn't work
```

yakov-g:

custom_model_runner/datarobot_drum/drum/model_adapter.py
```diff
@@ -79,6 +85,10 @@ def __init__(self, model_dir, target_type=None):
         self._custom_task_class = None
         self._custom_task_class_instance = None
 
+    def _log_and_raise_final_error(self, exc: Exception, message: str) -> NoReturn:
+        self._logger.error(f"{message} Exception: {exc!r}")
+        raise DrumPythonModelAdapterError(f"{message} Exception: {exc!r}")
```
rvorobii, line 90: will the traceback of the original exception be printed to logs?

e.g.
- don't we need `exc_info=True` for logger?
- don't we need `raise DrumPythonModelAdapterError(...) from exc` here?

eric-s-s: you want `logger.exception`

https://github.com/python/cpython/blob/3.12/Lib/logging/__init__.py#L1558

```python
    def error(self, msg, *args, **kwargs):
        """
        Log 'msg % args' with severity 'ERROR'.

        To pass exception information, use the keyword argument exc_info with
        a true value, e.g.

        logger.error("Houston, we have a %s", "major problem", exc_info=1)
        """
        if self.isEnabledFor(ERROR):
            self._log(ERROR, msg, args, **kwargs)

    def exception(self, msg, *args, exc_info=True, **kwargs):
        """
        Convenience method for logging an ERROR with exception information.
        """
        self.error(msg, *args, exc_info=exc_info, **kwargs)
```

eric-s-s: you shouldn't need the `from` to get the stacktrace, but it shouldn't hurt either.

eric-s-s (approved):
:100: looks good, thanks!  

nit:  add exc_info=True or logger.exception

yakov-g:

yakov-g:
jenkins run continuous-integration/jenkins/pr-merge

rvorobii (approved):

PR #848, opened 2023-10-31
Add clustering example

kindofluke:
# Cluster Model Examples

## Summary
We have recently had a few requests to demonstrate how clustering models might be supported in DataRobot as custom models. For most use cases, after the model is trained the number of clusters is known making it suitable for use as a multiclass. However, libraries like SKLearn do not output clustering results the same way that multiclass outputs are structured. For instance, there is no "probability" score as is required by our multiclass models. 

This PR adds two examples for clustering models as multi-class. The first uses SkLearn and simply assigned the probabilities as 1 or 0. Thus the output is `[1,0,0]`. This has the benefit of being very simple and fully testable by DRUM. 

The second model is an unstructured model that includes MLOps reporting. This provides a good practical example of using these monitored models.


model_templates/python3_sklearn_clustering_mlops_reporting/README.md
```diff
@@ -0,0 +1,48 @@
...
+        "Sepal.Width":3.1,
+        "Petal.Length":1.5,
+        "Petal.Width":0.2
+    }
+]
+```
+
+When this model is deployed you would call by sending this JSON String in the body of your POST request. 
+
+### Note
```
yakov-g, line 46: You still should be able to add an example command with `drum score`
Or maybe add an example of the command with `drum server` 
like here:
https://github.com/datarobot/datarobot-user-models/tree/master/model_templates/python3_unstructured
Up to you

yakov-g (approved):

yakov-g:
I see this pr is a pull request from a fork repo.
I'm not sure it will work.
Could you please create a branch in the repository itself and create a pull request.

elatt:
Closing due to inactivity.

PR #854, opened 2023-11-02
[NCA-1696] Added Streamlit environment

demchukilya:

### DIFF
--- a/public_custom_application_environments/python39_streamlit/Dockerfile
+++ b/public_custom_application_environments/python39_streamlit/Dockerfile
@@ -1,0 +1,19 @@
+FROM python:3.9-slim
+
+WORKDIR /app
+
+RUN apt-get update && apt-get install -y \
+    build-essential \
+    curl \
+    software-properties-common \
+    git \
+    && rm -rf /var/lib/apt/lists/*
+
+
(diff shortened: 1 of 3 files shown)


public_custom_application_environments/python39_streamlit/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM python:3.9-slim
+
+WORKDIR /app
+
+RUN apt-get update && apt-get install -y \
+    build-essential \
+    curl \
+    software-properties-common \
+    git \
```
ben-cutler-datarobot, line 9: I'm not totally sure we need all these dependencies. They came from an example streamlit app on Streamlit's site which does a 'git clone'

public_custom_application_environments/python39_streamlit/README.md
ben-cutler-datarobot, line 1: You can add Dr streamlit via a git link 😃

ben-cutler-datarobot (approved):
Looks good overall, but a few thoughts

public_custom_application_environments/python39_streamlit/Dockerfile
klichukb: I agree, please aim for as compact envs as possible.

PR #862, opened 2023-11-08
[RAPTOR-10068] add drapps binary

yakov-g:

## Summary
As per request form @liororama , adding `drapps` binary and tests for it.

### DIFF
--- a/custom_model_runner/setup.py
+++ b/custom_model_runner/setup.py
@@ -68,7 +68,7 @@ setup(
         "datarobot_drum.resource.pipelines": ["*"],
         "datarobot_drum.resource.default_typeschema": ["*.yaml"],
     },
-    scripts=["bin/drum", "../drapps/drapps.py"],
+    scripts=["bin/drum", "../drapps/drapps.py", "../drapps/drapps"],
     install_requires=requirements,
     extras_require=extras_require,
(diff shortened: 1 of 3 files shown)


ben-cutler-datarobot (approved):

akshoop:
FYI @eric-s-s @scottp-ml for review on behalf of Core Modeling

demchukilya (approved):

gkoundry (approved):
LGTM Core Modeling

scottp-ml (approved):

custom_model_runner/setup.py
```diff
@@ -68,7 +68,7 @@
         "datarobot_drum.resource.pipelines": ["*"],
         "datarobot_drum.resource.default_typeschema": ["*.yaml"],
     },
-    scripts=["bin/drum", "../drapps/drapps.py"],
+    scripts=["bin/drum", "../drapps/drapps.py", "../drapps/drapps"],
```
eric-s-s, line 71: why do we have both?  can we keep the new one and get rid of the old one?

yakov-g: No, we can't.
This script `drapps.py` is outside of the package, so it is not packaged automatically. I have to put it into scripts.

eric-s-s (approved):

PR #869, opened 2023-11-14
[RAPTOR-10097] add multiple artifacts negative tests

yakov-g:

## Summary
As these checks were removed from functional tests 
https://github.com/datarobot/datarobot-user-models/pull/865

add them as unit tests for Python; and integration tests for R.


tests/drum/test_other_cases_per_framework.py
```diff
@@ -311,3 +312,47 @@ def test_set_language(
                 != -1
             )
             assert case
+
+    @pytest.mark.parametrize(
```
eric-s-s, line 316: is this an integration test?  if so, maybe move to tests/integration?  i'm not sure of the test structure here.

eric-s-s (approved):

rvorobii (approved):

PR #871, opened 2023-11-16
[RAPTOR-10117] properly name functional/e2e tests

yakov-g:

## Summary
1. Move tests/functional -> tests/e2e - these are the tests that run with DataRobot app
2. Move tests/dum/constants.py - > tests/constants.
3. Move tests/drum -> tests/functional (as most of the tests in this folder run checks by running cli `drum`)
4. Rename integration scripts -> functional


eric-s-s:
as per previous discussion i think this:

> Move tests/functional -> tests/functional_with_dr

would be better moved to `tests/e2e`  or maybe a similar name.  There's no reason the name can't be fully descriptive and it would be better for us and future devs.  so while `e2e` or `acceptance_tests` are _fine_, it might be better to really spell out what these tests are for.

- tests/e2e/
- tests/acceptance_tests
- tests/contract_tests (making sure that they adhere to a contract with DR :shrug: )
- tests/e2e_with_full_datarobot_app
- tests/acceptance_tests_against_running_datarobot_instance
- any combination of above.

@scottp-ml @yakov-g

yakov-g:
@eric-s-s updated test names - check tests run.
Still need to add separate job for integration.

rvorobii (approved):

PR #875, opened 2023-11-23
[MMM-15567] introduce an environment that supports public custom metric

m-borkowski:

## Summary
I would like to introduce a new environment that supports public custom metrics (custom metrics templates).
Basically from a specific template we will create a custom job that will run on a schedule (hosted custom metric), to do this we need to have an environment that supports metric code.


rvorobii:
@yakov-g please review

public_custom_metrics_templates_environments/python39_llm_templates/README.md
```diff
@@ -0,0 +1,16 @@
+# Python 3.9 Custom Metrics Templates Drop-In Environment
+
+This environment is used as the default environment in custom metrics templates.
+Hosted custom metrics created from templates have this environment attached.
```
dlski, line 4:
```suggestion
Hosted custom metrics created from templates will run in this environment.
```

dlski (commented):
Just one question - in terms of IBS (next-gen infra) does this support proper user IDs and file system permissions?

m-borkowski:
> Just one question - in terms of IBS (next-gen infra) does this support proper user IDs and file system permissions?

I've tested this image with proposed custom metrics templates, and everything looks OK.
Image has ingested all necessary ENVs like `api_key` and `base_url`.
Files are read and run without any issues.

yakov-g:
Hey @m-borkowski ,
Several notes:
1. I think there is an agreement that the env marked is [DataRabot] becomes responsibility of DR to be maintained and updated. Do we want to mark this env as such right away? We may put [Experimental] tag for now.
2. We put evns into this repo to maintain them faster with every DRUM update. But in the beginning there were only envs added by CustomModels teams. Now there are envs wihich we(CustomModels) know nothing about, and which have other dependencies in DR. For example: with the envs we had from the very beginning, we roll out them on staging, then after a week we roll out them on prod. For GenAI env there are some additional deps which should be installed into DR before publishing env.
3. dependabot is going to scream about all the new packages this env has in the requirements,file.
4. Would it make more sense to put it into MMM related repo or create just envs related repo ASAP?

m-borkowski:
@yakov-g 
hey Yakov, thanks for the response
1. I will add  [Experimental] tag 
Sure, I understand, I am going to close this PR and open a new one in some  mmm-only related repo.
Next I will request the system admin to add it on staging.

PR #893, opened 2023-12-19
[AGENT-5214] Add support for numeric runtime parameters

dulcardo:

## Summary
Adding new runtime parameter type numeric, this is very similar to boolean type added perviously (https://github.com/datarobot/datarobot-user-models/pull/887)

Additionally we are adding two fields specific for numeric runtime, max/min values that are optional, when provided there is a extra validation in DRUM as well in DR backend to enforce that the provided values is between the range

yaml configuration example:
```
name: runtime-params-example
type: inference
targetType: regression

runtimeParameterDefinitions:
  - fieldName: number1
    type: numeric
    defaultValue: 0
    minValue: -100
    maxValue: 100

  - fieldName: number2
    type: numeric
    defaultValue: 0
```

* Add unit/functional tests
* update `model_templates/python3_sklearn_runtime_params/model-metadata.yaml` template to include bool/numeric runtime params

### DIFF
--- a/custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters_schema.py
+++ b/custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters_schema.py
@@ -14,6 +14,7 @@ class RuntimeParameterTypes(Enum):
     BOOLEAN = "boolean"
     CREDENTIAL = "credential"
     DEPLOYMENT = "deployment"
+    NUMERIC = "numeric"
 
 
 class NativeEnumTrafaret(t.Enum):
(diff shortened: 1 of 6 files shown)


zohar-mizrahi (approved):

custom_model_runner/datarobot_drum/runtime_parameters/runtime_parameters_schema.py
```diff
@@ -75,5 +85,7 @@ def __init__(self):
         t.Key("fieldName", to_name="name"): t.String,
         t.Key("type"): NativeEnumTrafaret(RuntimeParameterTypes),
         t.Key("defaultValue", optional=True, default=None, to_name="default"): t.Any,
+        t.Key("minValue", optional=True, default=None, to_name="min_value"): t.Any,
```
akshoop, line 88: Should the values be Any type or should they be Int type?

akshoop (approved):
LGTM from Core Modeling with minor question, CC @eric-s-s @scottp-ml for awareness

PR #908, opened 2024-01-04
[AGENT-5226] add new field to indicate if a runtime parameter is mandatory

dulcardo:

## Summary

Adding new field for runtime parameters, to specify that a field must be set before registration. This field `allowEmpty` is optional and set to `True` by default.

The check will be done in the app, before registering any custom model, the runtime params that has `allowEmpty=False` need to have a current value, either a default value or override.


dulcardo:
jenkins run test_functional_r_environment

yakov-g (approved):

eric-s-s (approved):
:+1:

PR #926, opened 2024-02-05
[RAPTOR-10333] Handle redundant quotation marks in target name when extracting extra model output

zohar-mizrahi:
## Summary
It turns out that DataRobot backend adds redundant quotation marks to the target name, when it sets up the `TARGET_NAME` environment variable for the text-generation model. Consequently, if the user tries to set the output column hard-coded, without reading the `TARGET_NAME` from the environment, the extraction of the additional columns will fail. 

For instance if the user sets the target column in the UI to `response` and try to return a DataFrame with the response column name hard-coded from the `score` method in `custom.py`, than the whole prediction will fail, saying that there is no `"response"` column in the output. But if the user reads the target column from the environment variable `TARGET_NAME` everything will work as expected, because the value in this environment variable contains the redundant quotation marks (`"response"`).

To fix this discrepancy, in drum layer, the target name is first being used as is. If it does not exist in the response DataFrame, there is an attempt to remove the quotation marks (if exists) and continue.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
+++ b/custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
@@ -616,7 +616,18 @@ class PythonModelAdapter(AbstractModelAdapter):
                 predictions_df = result_df
         else:
             if len(result_df.columns) > 1:
-                target_column = self._target_name if self._target_name else PRED_COLUMN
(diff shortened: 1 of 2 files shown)


datarobot-devinfragit3:

scottp-ml (approved):

yakov-g (approved):
Should those extra quotes be fixed in the DR?

PR #944, opened 2024-03-11
AGENT-5349 add NVIDIA Triton Server dropin environment

alexromanyuk:


datarobot-devinfragit3:

public_dropin_environments/triton_server/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# This is the default base image for use with user models and workflows.
+FROM nvcr.io/nvidia/tritonserver:22.04-py3
```
elatt, line 2: Is this right?
```suggestion
FROM nvcr.io/nvidia/tritonserver:24.02-py3
```

yakov-g:
@alexromanyuk @DrewParallel 
**public_dropin_environments** holds environments which are historically maintained by Custom Models team. We run logical (framework related artifacts) and functional(drum) tests on them.
We periodically update those environments when deps or drum are bumped. 
Also these envs are package into installer for OnPrem installations. 

The only relatively new env in that folder was GenAI (which I would probably also move out).

So I'd suggest to put new nvidia related envs into a separate folder, `public_dropin_environments_gpu` ?

public_dropin_environments/triton_server/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+FROM nvcr.io/nvidia/tritonserver:24.02-py3
+
+# Install the list of core requirements, e.g. sklearn, numpy, pandas, flask.
+# **Don't modify this file!**
+COPY dr_requirements.txt dr_requirements.txt
+
+# '--upgrade-strategy eager' will upgrade installed dependencies
+# according to package requirements or to the latest
+RUN pip3 install -U pip && \
+    pip3 install -U --upgrade-strategy eager --no-cache-dir -r dr_requirements.txt  && \
```
elatt, line 11:
```suggestion
    pip3 install -U --upgrade-strategy eager --no-cache-dir -r dr_requirements.txt  && \
```
`-U --upgrade-strategy eager` is a little dangerous right? Since we don't control the upstream env I think we should be as isloated as possible.

public_dropin_environments/triton_server/start_server.sh
```diff
@@ -0,0 +1,25 @@
...
+    echo "Environment variables:"
+    env
+fi
+
+export MODEL_DIR="${CODE_DIR}/model_repository/"
+
+echo
+echo "Executing command: tritonserver --model-repository=${MODEL_DIR}"
+echo
+exec nohup tritonserver --model-repository=${MODEL_DIR} > log.txt 2>&1 &
```
elatt, line 20:
```suggestion
nohup tritonserver --model-repository=${MODEL_DIR} > log.txt 2>&1 &
```
I don't think `exec` is needed here -- I'm a little surprised it works TBH. Can you explain why you need it?

Also, I'm worried about sending all output to the `log.txt`. From my experience playing around with this stuff, if there is a problem with the model file, the output will be from Trition explaining what is wrong and with this line, I don't see how the user will ever see that? What is the problem with just leaving the output going to stdout/stderr -- does this result in the output getting intermixed, is it still readable?

public_dropin_environments/triton_server/dr_requirements.txt
```diff
@@ -0,0 +1 @@
+requests
```
elatt, line 1: Am I missing something, how does drum get installed, doesn't that need to be listed as a requirement because we aren't using the standard drop-in base env?

Also, I think we should freeze these deps so re-builds are deterministic.

public_dropin_environments/triton_server/README.md
```diff
@@ -0,0 +1,44 @@
...
+    densenet_labels.txt
+    1/
+      model.onnx
+```
+
+This environment makes the following assumption about your serialized model:
+- The data sent to custom model can be used to make predictions without
+additional pre-processing
+- There is a single model and model version present
+- Model is expected to have one of the target types: `Unstructured` or `TextGeneration`
```
elatt, line 44: Is this right? This is just the PyTorch, ONNX, etc env.

public_dropin_environments/triton_server/README.md
```diff
@@ -0,0 +1,44 @@
+# NVIDIA Triton Inference Server Drop-In Template Environment
+
+This template environment can be used to create artifact-only custom models with 
+the following Triton Inference Server backends: ONNX, vLLM, TensorRT-LLM.
```
elatt, line 4: This env doesn't support TensorRT-LLM or vLLM right? it's just the standard PY3 env from your base image. This is what upstream says, I think we should include similar language
> The xx.yy-py3 image contains the Triton inference server with support for Tensorflow, PyTorch, TensorRT, ONNX and OpenVINO models.

elatt (changes requested):
My main objection is that I worry we need a way to surface errors from `tritionserver` to the user.

public_dropin_environments/triton_server/start_server.sh
```diff
@@ -0,0 +1,25 @@
...
+#
+# All rights reserved.
+# This is proprietary source code of DataRobot, Inc. and its affiliates.
+#
+# Released under the terms of DataRobot Tool and Utility Agreement.
+echo "Starting Custom Model environment with Triton inference server"
+
+if [ "${ENABLE_CUSTOM_MODEL_RUNTIME_ENV_DUMP}" = 1 ]; then
+    echo "Environment variables:"
+    env
```
elatt, line 12:
```suggestion
    env

    echo
    echo "Running NVIDIA init scripts..."
    echo
    /opt/nvidia/nvidia_entrypoint.sh /bin/true
```
What do you think about running these if we are in some sort of debugging mode?

rvorobii:
@yakov-g is already assigned to review this PR, so I'll remove "Needs Review: Custom Models" label

public_dropin_environments/triton_server/dr_requirements.txt
alexromanyuk: Drum should be listed here. I removed it during the local testing. 
Added the `datarobot-drum==1.10.20` which is the version without any Triton support, but we'll update it later.

alexromanyuk: pined requests to `requests==2.31.0`.

public_dropin_environments/triton_server/Dockerfile
alexromanyuk: I don't think we need the "eager" strategy here. It's here because I just reused the Dockerfile from existing dropin env. Will drop

public_dropin_environments/triton_server/start_server.sh
alexromanyuk: The `exec` has no effect here, the `nohup &` is sufficient to run Triton in the background process. Removed the `exec` command.

public_dropin_environments/triton_server/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    rm -rf dr_requirements.txt
+
+# Copy the drop-in environment code into the correct directory
+# Code from the custom model tarball can overwrite the code here
+ENV HOME=/opt CODE_DIR=/opt/code ADDRESS=0.0.0.0:8080
```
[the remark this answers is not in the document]
alexromanyuk: removed the `HOME=/opt` env var

public_dropin_environments/triton_server/README.md
alexromanyuk: Yes, it's ONNX only, will update

public_dropin_environments/triton_server/start_server.sh
alexromanyuk: It's nice to have. I think, we may want to run it w/o any special env vars. I'll leave it as is for now.

alexromanyuk: And you're using the `/bin/true` here, because of this:
```
# This script can either be a wrapper around arbitrary command lines,
# or it will simply exec bash if no arguments were given
if [[ $# -eq 0 ]]; then
  exec "/bin/bash"
else
  exec "$@"
fi
```

public_dropin_environments/triton_server/README.md
alexromanyuk: updated

public_dropin_environments/triton_server/start_server.sh
alexromanyuk: > What is the problem with just leaving the output going to stdout/stderr -- does this result in the output getting intermixed, is it still readable

I just wanted to see output from Flask proxy more, than output from Triton. But for user's containers, I agree Triton logs are more important. I'll remove output redirection

public_dropin_environments/triton_server/start_server.sh
elatt: Yeah, if you don't have _something_ then it goes into a shell and your container will just hang.

elatt (approved):

yakov-g (approved):

yakov-g:
@alexromanyuk please set codeowners for this new folder. maybe i the next PR

alexromanyuk:
@yakov-g I'll have two more PRs soon, will update code owners there.

PR #946, opened 2024-03-14
[RAPTOR-10347] Propagate error with strict yaml

dulcardo:

## Summary
Yaml with flow style is not allowed by StrictYaml,  https://hitchdev.com/strictyaml/why/flow-style-removed/. so providing something like is not allowed.

```
name: runtime-params-example
type: inference
targetType: regression

runtimeParameterDefinitions: []
```

This PR propagates the error so it can be visible from user running in custom models.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/model_metadata.py
+++ b/custom_model_runner/datarobot_drum/drum/model_metadata.py
@@ -22,6 +22,7 @@ from strictyaml import (
     Int,
     Any,
     StrictYAMLError,
+    YAMLValidationError,
 )
 from typing import Optional as PythonTypingOptional, List, Dict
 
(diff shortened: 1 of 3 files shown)


yakov-g (approved):

gkoundry (approved):
LGTM Core Modeling

custom_model_runner/datarobot_drum/drum/model_metadata.py
```diff
@@ -151,12 +163,20 @@ def read_model_metadata_yaml(code_dir) -> PythonTypingOptional[dict]:
...
                 if "found a blank string" in e.problem:
                     print("The model_metadata.yaml file appears to be empty.")
                 else:
                     print(e)
                 raise SystemExit(1)
+            except StrictYAMLError as e:
+                raise DrumFormatSchemaException(
+                    "\nStrictYAMLError: The current format does not comply with strict yaml rules."
+                    " (Empty list on fields are not allowed)\n{}".format(e)
+                )
```
elatt, line 176: Why for some exceptions we print and exit but for this one we raise a new exception class? From the looks of it, these all appear to be similar classes of errors (e.g. bad user input) so I'm surprised we are behaving differently in this way. Should all of them be raising an exception -- is DRUM server swallowing the errors we print?

dulcardo: Yes the prints error are not propagated to logs shows in the DataRobot App, when some of these error happen the logs display in the UI only show `Error drum: 1`.

We could consolidate all exceptions into one common that is a good idea, but I guess we should raise exceptions for all error related to user input.

PR #951, opened 2024-03-19
AGENT-5461 add DRUM proxy to NVIDIA Nemo Inference Microservices (NIM)

alexromanyuk:
## Summary
PR adds the `NemoPredictor` which is turned on for text generation model types and only if the `--with-nemo-server` flag is set. The `NemoPredictor` starts the nemollm inference server in a subprocess and then proxies completion requests sent to the `/predict` route.

## Tesing
All the testing is done manually, automated tests will be added by separate PRs. 

Run DRUM
```
export MLOPS_RUNTIME_PARAM_chat_context='{"type": "boolean", "payload": true}'

export MODEL_NAME="generic_llm"
export NEMO_HOST="http://35.172.164.204"
export TARGET_NAME="promptText"
export GPU_COUNT=1

# for local testing any code directory will work, as it's not used by the NemoPredictor,
# (unless you have a custom.py, e.g. to download model)
drum server --target-type=textgeneration \
  --code-dir /home/alexandr.romanyuk/workspace/external/triton-demo/nemo \
  --address localhost:8080 \
  --gpu-predictor=nemo
```
Request completions:
```
import requests

endpoint = 'http://localhost:8080/predict/'
headers = { 'accept': 'application/json' }
response = requests.post(endpoint, headers=headers, files=[('X', open('prompt_chat.txt'))])
print(response.text)
```
Chat prompt:
```
promptText,assistant
"tell me a joke","Sure! Here's one: Why don't scientists trust atoms? Because they make up everything."
"another one, please","Of course! How about this one: What do you call an alligator in a vest? An investigative journalist."
"try one more time, but no more jokes about scientists or aligators",
```
Response:
```
{"predictions": [
  "My apologies for that previous answer. Let me try again with something else...
   Did you hear about the scarecrow who won an award? He was outstanding in his field!"
]}
```


custom_model_runner/datarobot_drum/drum/gpu_predictors/nemo_predictor.py
```diff
@@ -0,0 +1,126 @@
...
+DEFAULT_MODEL_NAME = "generic-llm"
+
+
+class NemoPredictor(BaseLanguagePredictor):
+    def __init__(self):
+        super(NemoPredictor, self).__init__()
+        self.logger = logging.getLogger(LOGGER_NAME_PREFIX + "." + __name__)
+        self.prompt_field = None
+        self.triton_host = None
+        self.triton_http_port = None
```
alexromanyuk, line 36: Those are currently used only for local dev, so naming does not matter much yet.

public_dropin_gpu_environments/nemollm_inference_ms/Dockerfile
```diff
@@ -12,8 +12,9 @@ RUN groupmod -g 1000 nemo && \
 # try and unpack any layers with out-of-bound UIDs
 FROM scratch
 COPY --from=0 / /
+ARG BASE_PATH=/opt/tritonserver/bin:/usr/local/cmake/bin:/usr/local/cmake/bin:/usr/local/mpi/bin:/usr/local/nvidia/bin:/usr/local/cuda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/ucx/bin:/opt/tritonserver/bin
 
-ENV PATH=/opt/tritonserver/bin:/usr/local/cmake/bin:/usr/local/cmake/bin:/usr/local/mpi/bin:/usr/local/nvidia/bin:/usr/local/cuda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/ucx/bin:/opt/tritonserver/bin \
+ENV PATH=${BASE_PATH} \
```
elatt, line 17: Why this? e.g. why set this as an ARG? -- with the latest base image, hopefully this stuff all goes away because the UID is now 1000 for the nemo user.

public_dropin_gpu_environments/nemollm_inference_ms/Dockerfile
```diff
@@ -71,10 +72,12 @@ RUN apt-get update && apt-get install -y \
   && rm -rf /var/lib/apt/lists/*
 
 USER 1000
+ENV DATAROBOT_VENV_PATH=/home/nemo/dr
 ENV PIP_NO_CACHE_DIR=1
-RUN python -m venv /home/nemo/dr && /home/nemo/dr/bin/pip install -U pip
+ENV PATH=${BASE_PATH}
```
elatt, line 77: Why are you setting this again? You already set it on line 17?

public_dropin_gpu_environments/nemollm_inference_ms/Dockerfile
alexromanyuk: > with the latest base image

The latest image changes the required repository structure. Models compiled earlier for A10 are not working out-of-the box. I'll retry those latter.

custom_model_runner/datarobot_drum/drum/gpu_predictors/nemo_predictor.py
```diff
@@ -0,0 +1,239 @@
...
+import logging
+import os
+import signal
+import subprocess
+import atexit
+
+import numpy as np
+from openai import OpenAI
+from threading import Thread
+from datarobot_drum import RuntimeParameters
```
elatt, line 18: *nit, can you fixup these imports -- `threading` should move up to top section and then you should add a 3rd section for `numpy` and `openai` (e.g. 3rd party libraries)

custom_model_runner/datarobot_drum/drum/gpu_predictors/nemo_predictor.py
```diff
@@ -0,0 +1,239 @@
...
+import csv
+import io
+import logging
+import os
+import signal
+import subprocess
+import atexit
+
+import numpy as np
+from openai import OpenAI
```
elatt, line 16: How do we ensure that OpenAI is installed? Do we need to add an optional dependency to the drum package?

public_dropin_gpu_environments/nemollm_inference_ms/Dockerfile
alexromanyuk: env vars were exported for `root` in the build, then you changed to the nemo user. is docker expected to set the same env vars for both users?

I had issues with the path, nemo user was unable to find `mpirun`.

custom_model_runner/datarobot_drum/drum/gpu_predictors/nemo_predictor.py
```diff
@@ -0,0 +1,239 @@
...
+from datarobot_drum.drum.enum import (
+    LOGGER_NAME_PREFIX,
+    PayloadFormat,
+    StructuredDtoKeys,
+)
+from datarobot_drum.drum.exceptions import DrumCommonException
+from datarobot_drum.drum.language_predictors.base_language_predictor import BaseLanguagePredictor
+from datarobot_drum.resource.drum_server_utils import wait_for_server, DrumServerProcess
+
+RUNNING_LANG_MSG = "Running environment: Nemo Inference Microservices."
```
elatt, line 30:
```suggestion
RUNNING_LANG_MSG = "Running environment: NeMo Inferencing Microservice."
```
This appears to be the official name? https://developer.nvidia.com/blog/nvidia-nim-offers-optimized-inference-microservices-for-deploying-ai-models-at-scale/

custom_model_runner/datarobot_drum/drum/gpu_predictors/nemo_predictor.py
alexromanyuk: currently it's not supposed to be useable without the drop-in environment.

>  Do we need to add an optional dependency to the drum package

yes

custom_model_runner/datarobot_drum/drum/gpu_predictors/nemo_predictor.py
```diff
@@ -0,0 +1,239 @@
...
+    SYSTEM = "system"
+    ASSISTANT = "assistant"
+    USER = "user"
+
+
+class NemoPredictor(BaseLanguagePredictor):
+    def __init__(self):
+        super(NemoPredictor, self).__init__()
+        self.logger = logging.getLogger(LOGGER_NAME_PREFIX + "." + __name__)
+        self.datarobot_venv_path = os.environ.get("DATAROBOT_VENV_PATH", "/home/nemo/dr")
```
elatt, line 44: Aren't we running inside of the venv already? e.g. didn't we install and start drum up inside of the venv so why do we need to know it's path? We can deduce that from the location of the python binary we are running as right?

alexromanyuk: updated to os.environ.get('VIRTUAL_ENV')

custom_model_runner/datarobot_drum/drum/gpu_predictors/nemo_predictor.py
```diff
@@ -0,0 +1,239 @@
...
+        # expected to be set for text generation type
+        self.prompt_field = os.environ.get("TARGET_NAME")
```
elatt, line 45: From my experience, this isn't the prompt field, this is the response field. Buzok appears to use a runtime param to define the prompt field.

alexromanyuk: ok, buzok has the `TARGET_NAME` and the `PROMPT_COLUMN_NAME`. but why, isn't that confusing?
buzok-model

alexromanyuk: text-gen-model-target-name

alexromanyuk: the `TARGET_NAME` env var is set from this UI field. What Buzok does is a question :question:

alexromanyuk:
jenkins rerun failed

alexromanyuk:
jarvis run test_functional_general

alexromanyuk:
jarvis run test_functional_python_keras

scottp-ml (approved):

custom_model_runner/datarobot_drum/drum/gpu_predictors/nemo_predictor.py
alexromanyuk: updated code to read a prompt filed name from the `MLOPS_RUNTIME_PARAM_prompt_column_name`. 

Buzok uses upper case for parameter names, I chose to use lowercase to be similar to [OpenAI request parameters](https://platform.openai.com/docs/api-reference/chat/create).

custom_model_runner/datarobot_drum/drum/gpu_predictors/nemo_predictor.py
alexromanyuk: I'm not adding this dependency into DRUM for now. It may be useful only for local dev work, when you have DRUM on local machine and NIM server on remote.

alexromanyuk:
jenkins run test_functional_general

alexromanyuk:
jenkins run test_functional_python_keras

custom_model_runner/datarobot_drum/drum/gpu_predictors/nemo_predictor.py
```diff
@@ -0,0 +1,250 @@
...
+        self.nemo_thread.start()
+        self._check_nemo_health()
+        self.nim_client = OpenAI(base_url=f"{self.nemo_host}:{self.openai_port}/v1", api_key="fake")
+
+    @staticmethod
+    def get_optional_parameter(key, default_value=None):
+        try:
+            return RuntimeParameters.get(key)
+        except ValueError:
+            return default_value
```
elatt, line 86: We need to think about this more. Either we need another interface for this or we need to add support for Execution Environments to support **defining** runtime params, and then the custom model will be a combination of the runtime param defs from the selected environment version and also from any defined in the model itself.

custom_model_runner/datarobot_drum/drum/gpu_predictors/nemo_predictor.py
```diff
@@ -0,0 +1,250 @@
...
+        # update the path so nemollm process can find its libraries
+        env = os.environ.copy()
+        datarobot_venv_path = os.environ.get("VIRTUAL_ENV")
+        env["PATH"] = env["PATH"].replace(f"{datarobot_venv_path}/bin:", "")
+        with subprocess.Popen(
+            cmd,
+            env=env,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            preexec_fn=os.setsid,
```
elatt, line 208: What is `setsid` for ?  Maybe worth a comment because it's not obvious why we need to run in a new process group?

custom_model_runner/datarobot_drum/drum/gpu_predictors/nemo_predictor.py
```diff
@@ -0,0 +1,250 @@
...
+        ) as p:
+            nemo_process.process = p
+            for line in p.stdout:
+                self.logger.info(line[:-1])
+
```
elatt, line 213: Is this necessary/preferred? Does the output not show up well if we just let Triton log to the normal stderr/out?

custom_model_runner/datarobot_drum/drum/gpu_predictors/nemo_predictor.py
alexromanyuk: yeah, nim creates a bunch of subprocesses, 

```
$  ps x -o  "%p %r %y %x %c "
    PID    PGID TTY          TIME COMMAND         
    590     590 pts/0    00:00:00 bash            
   1688    1688 pts/3    00:00:00 bash            
   1898    1898 pts/2    00:00:00 bash            
   2123    2123 ?        00:00:00 run_nemo.sh     
   2130    2123 ?        00:00:00 nemollm_inferen 
   2133    2123 ?        00:00:01 python3         
   2134    2123 ?        00:00:00 python3         
   2135    2123 ?        00:00:00 dash            
   2136    2123 ?        00:00:00 mpirun          
   2140    2140 ?        00:04:38 tritonserver    
   2159    2140 ?        00:00:03 triton_python_b 
   2268    2268 pts/0    00:00:00 ps      
```

public_dropin_gpu_environments/nemollm_inference_ms/Dockerfile
elatt: Oh hmm, I thought env vars were global to the image and not based on `USER` 😕

elatt (approved):

custom_model_runner/datarobot_drum/drum/gpu_predictors/nemo_predictor.py
alexromanyuk: I'll create a ticket to track that work

custom_model_runner/datarobot_drum/drum/args_parser.py
```diff
@@ -683,10 +683,24 @@ def _reg_arg_dr_api_access(*parsers):
     @staticmethod
     def _reg_arg_triton_server_access(*parsers):
         for parser in parsers:
-            parser.add_argument(
+            server_group = parser.add_mutually_exclusive_group(required=False)
+            server_group.add_argument(
```
yakov-g, line 687: Probably should be organized as a single param with multiple choices.
--with-nvidia-server=triton/nemo

yakov-g: Especially when you add mutually exclusive.
Could you redo this in this PR?

elatt: We plan to add at least one more server too though: e.g. vLLM which has no relation to nvidia, so maybe `--with-nvidia-server` needs to be more generic: `--with-inference-server` ?

yakov-g: Or even more general: add a param to specify predictor instead of resolving it by language/artifacts.
--predictor==Python/R/Java/Triton/Nemo.

yakov-g: Maybe it worth introducing `--predictor` param. If specified, ignore language-based resolution and just used what is provided

custom_model_runner/datarobot_drum/resource/components/Python/generic_predictor/generic_predictor.py
```diff
@@ -70,12 +72,18 @@ def configure(self, params):
             from datarobot_drum.drum.language_predictors.r_predictor.r_predictor import RPredictor
 
             self._predictor = RPredictor()
-        elif self._run_language == RunLanguage.OTHER and self._with_triton_server:
-            from datarobot_drum.drum.language_predictors.triton_predictor.triton_predictor import (
+        elif self._with_triton_server and self.cli_adapter.target_type == TargetType.UNSTRUCTURED:
```
yakov-g, line 75: I would add triton/nemo + target type check in the args parser.

custom_model_runner/datarobot_drum/drum/args_parser.py
elatt: 🤔 but from inside custom models workshop... customers don't have control over CLI flags right? So we kinda need to auto-detect stuff? Or are we saying that the env we are building will be locked to a specific predictor? I'll have to defer to Alex on that (I'm not sure if that model works with Triton-python env)

yakov-g: Most of the CLI params are also easily duplicated with ENV vars, we have this mechanism built in into DRUM.

By default, our dropin envs have only deps that are reflected in the env names: sklearn, pytorch. So if you upload onnx artifact into sklearn env it likley won't work. Though you can still install additional packages and it will.

So you are right about locking, maybe not everything should be locked.
But in the PR Alex tries to resolve predictor based on param provided and target type, which seems not right. It would be better to provide predictor and verify, that target type is suitable to that predictor.

yakov-g (approved):
@alexromanyuk 
I approve, but please implement the `--predictor` option I mentioned in comments.
I think I also mentioned this at the time you added Triton predictor.

custom_model_runner/datarobot_drum/drum/args_parser.py
elatt: > Most of the CLI params are also easily duplicated with ENV vars, we have this mechanism built in into DRUM

How do our users set env vars though? (other than via runtime params but users shouldn't actually know those are env vars as it is an implementation detail).

Otherwise I think Alex will have to explain his vision for how he wants envs to work with these new inference servers. I think the Triton-py3 env is most complicated because we want a single env that has Triton-py3 installed but Triton can handle Pytorch, Onnx, Tensorflow, etc all from this one env so maybe that's where things are not aligning perfectly?

alexromanyuk: > customers don't have control over CLI flags right? So we kinda need to auto-detect stuff based on the files we see?

We don't need to autodetect artifacts. User selects a correct drop-in environment, where the `start-server.sh` sets the `--predictor=triton` or `--predictor=nemo` flag.

alexromanyuk: > Or even more general: add a param to specify predictor instead of resolving it by language/artifacts.
--predictor==Python/R/Java/Triton/Nemo.

@elatt @yakov-g I'm going to add the `--gpu-predictor` param and it will have only the `triton/nemo/vllm` choices. I know, Triton may work without the gpu, but the main reason we adding it - is the GPU support.

alexromanyuk: @yakov-g I've added the `--gpu-predictor=triton|nemo|vllm` option. If you think it would better to have the generic `--predictor` option, I'll add the change it in the followup.

custom_model_runner/datarobot_drum/resource/components/Python/generic_predictor/generic_predictor.py
alexromanyuk: added the `--gpu-predictor=triton|nemo|vllm` option.

PR #949, opened 2024-03-15
[AGENT-5372] Bump version to 1.10.21

dulcardo:
## Summary

Bump version to 1.10.21

* add new env vars for runtime parameter file
* Bug fixes for disallowed model-metadata file.
* Adds support for `datarobot-mlops` 10.0+

PR #952, opened 2024-03-19
[AGENT-5464] Prepare code for new import path in 10.0 version of mlops library

elatt:

## Summary
Add support for `datarobot_mlops.mlops` and `datarobot.mlops.mlops` import thats so we can support pre and post 10.0 versions of the `datarobot-mlops` library.

## Rationale
Make it smoother when we bump the version of mlops so we can support old and new import paths.

### DIFF
--- a/docker/dropin_env_base/requirements.txt
+++ b/docker/dropin_env_base/requirements.txt
@@ -1,2 +1,2 @@
 datarobot-drum==1.10.14
-datarobot-mlops==9.2.8
+datarobot-mlops==10.0.3a2
--- a/docker/dropin_env_base/requirements.txt
+++ b/docker/dropin_env_base/requirements.txt
@@ -1,2 +1,2 @@
 datarobot-drum==1.10.14
-datarobot-mlops==10.0.3a2
+datarobot-mlops==9.2.8
--- a/custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
+++ b/custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
@@ -34,6 +34,7 @@ mlops_import_error = None
 MLOps = None
 try:
     from datarobot_mlops.mlops import MLOps
+
     mlops_loaded = True
 except ImportError as e:
     mlops_import_error = "Error importing MLOps python module: {}".format(e)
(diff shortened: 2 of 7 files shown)


rvorobii:
@yakov-g please review

akshoop (approved):
LGTM, CC @eric-s-s @scottp-ml

dulcardo (approved):

docker/dropin_env_base/requirements.txt
```diff
@@ -1,2 +1,2 @@
 datarobot-drum==1.10.14
-datarobot-mlops==9.2.8
+datarobot-mlops==10.0.3a2
```
yakov-g, line 2: Is it ok to use it in pair with Java mlops-agent==9.2.8 ?

This is the agent installed on this envs. It is installed in the base images.

yakov-g (approved):
Please read my comment regarding mlops-agent

yakov-g:
You'll need to adjust mlops installation here:
jenkins/test_functional_general.sh

docker/dropin_env_base/requirements.txt
elatt: It _should_ be ok in this instance because we haven't really done much (if any?) agent API changes this release. But it would be safer if we kept things at the same major version.

yakov-g: But I see the latest version of mlops-agent on Maven is 9.2.8

elatt: Yeah so while in theory this should work for these version skews, I'm pretty sure in actuality we don't support doing such a thing: https://github.com/datarobot/tracking-agent/blob/187a4bc83c56cd638a7921e380a319ba14a534d2/mlops-py/datarobot_mlops/agent.py#L438-L443

I think I'm going to have to go back to a prior change I had where instead of bumping the version and locking into the new import path, I can make a more backward/forward compatible change (but it's a bit uglier).

elatt:
jenkins run test_functional_java_environment

elatt:
jenkins run 0-all-check-status

elatt:
jenkins run pipelineController:all-expected-tasks

elatt:
I'd like to try and merge this before https://github.com/datarobot/datarobot-user-models/pull/949

elatt:
Jenkins run test_functional_general

elatt:
jenkins run test_functional_general

elatt:
@yakov-g I need some help with this PR. I've rolled back the majority of changes -- the latest state should just make it compatible with both the 9.x and 10.x versions of `datarobot-mlops` but not bump any versions of MLOPs library yet. However, the tests seem to be hitting an import error that I don't understand why (because I attempt to import both the old and new paths).

elatt:
jenkins run test_functional_general

PR #953, opened 2024-03-19
AGENT-5461 add Nemo Inference Server drop-in environment

alexromanyuk:


public_dropin_gpu_environments/nemo_inference_ms/Dockerfile
```diff
@@ -0,0 +1,84 @@
+# This is the default base image for use with user models and workflows.
+FROM nvcr.io/ohlfw0olaadg/ea-participants/nemollm-inference-ms:24.01
```
alexromanyuk, line 2: **TODO** 
This image should be uploaded into our [docker hub
](https://hub.docker.com/orgs/datarobotdev/repositories)

public_dropin_gpu_environments/nemollm_inference_ms/start_server.sh
```diff
@@ -0,0 +1,54 @@
...
+# All rights reserved.
+# This is proprietary source code of DataRobot, Inc. and its affiliates.
+#
+# Released under the terms of DataRobot Tool and Utility Agreement.
+echo "Starting Custom Model environment with NIM"
+set -e
+
+if [ -z "${$MODEL_NAME}" ]; then
+    echo "The MODEL_NAME must be set in runtime parameters"
+fi
```
elatt, line 13: Runtime params don't work like this (it's JSON encoded and has the prefix `MLOPS_RUNTIME_PARAM_...` or something like that (we shouldn't rely on these details though) -- the only official way to support this is using the DRUM code though.

elatt: Why not just hard code the model name? From what I can tell, the only thing it is used for is to simulate the OpenAI API (which requires a model name because the same library can talk to a multitude of different models OpenAI currently has but with DRUM we will only be serving one model). As long as this startup code, and the _drum proxy code_ agree on the model name, I think it can be hard coded right?

public_dropin_gpu_environments/nemollm_inference_ms/start_server.sh
```diff
@@ -0,0 +1,54 @@
...
+echo "Starting DRUM server in background..."
+echo
+nohup drum server --with-triton-server "$@" &
+
+if [[ -e $CODE_DIR/custom.py ]]; then
+    echo
+    echo "Running custom.py..."
+    echo
+    python ${CODE_DIR}/custom.py
+fi
```
elatt, line 37: I don't think we will need this now that we are using DRUM because drum has it's own support for hooks that we can leverage to download models.

public_dropin_gpu_environments/nemollm_inference_ms/start_server.sh
```diff
@@ -0,0 +1,54 @@
...
+export NEMO_PORT=9998
+export OPENAI_PORT=9999
+export HEALTH_PORT=8081
+
+exec nemollm_inference_ms --model $MODEL_NAME \
+    --log_level=info \
+    --health_port=$HEALTH_PORT \
+    --openai_port=$OPENAI_PORT \
+    --nemo_port=$NEMO_PORT \
+    --num_gpus=$(nvidia-smi -L | wc -l)
```
elatt, line 54: I'd pull this line out from inline so you can check the output of the command to be sure we detected at least one GPU, otherwise I think we should exit with error (unlike pure Triton, I don't think there is any CPU only usecase for NIM).

elatt (changes requested):
Mostly looks good, just a few small things to fix before merging.

public_dropin_gpu_environments/nemollm_inference_ms/Dockerfile
```diff
@@ -0,0 +1,84 @@
...
+# it is out-of-range for our rootless buildkit image builder.
+RUN groupmod -g 1000 nemo && \
+  usermod -u 1000 -g 1000 nemo && \
+  chown -R 1000.1000 /home/nemo /model-store && \
+  rm /var/log/lastlog
+
+# Need to do this to squash our hack above so rootless buildkit won't
+# try and unpack any layers with out-of-bound UIDs
+FROM scratch
+COPY --from=0 / /
```
elatt, line 14: FWIW, this was an assumption that I never tested. Now that we are out of the time crunch for the demo, it is probably worth validating this is necessary because it would simplify things if it wasn't needed.

public_dropin_gpu_environments/nemollm_inference_ms/start_server.sh
alexromanyuk: Now we have two different "custom models" with support of downloading models:
- Triton's custom model which is served by Triton's python backend (e.g. vLLM model, uses python backend)
- DRUM custom model, with hooks.

alexromanyuk: removed this if-block.

public_dropin_gpu_environments/nemollm_inference_ms/start_server.sh
alexromanyuk: done

public_dropin_gpu_environments/nemollm_inference_ms/start_server.sh
alexromanyuk: > Why not just hard code the model name? 

Ah, cool. I didn't knew that you can set anything as a model name. I'll hardcode model name then for NIM.

For Triton, you have to specify the correct model name for API requests, which I'm reading from the `config.pbtxt`.

public_dropin_gpu_environments/nemollm_inference_ms/Dockerfile
alexromanyuk: without layers squash the image builder fails:
```
{"timestamp":"2024-03-20 18:04:58.294","level":"INFO","thread":"pool-7-thread-1","logger":"com.datarobot.imagebuilder.util.Logger","message":"#7 ERROR: mount callback failed on /run/user/1000/containerd-mount2453652930: failed to Lchown \"/run/user/1000/containerd-mount2453652930/home/nemo\" for UID 4474987, GID 4474987: lchown /run/user/1000/containerd-mount2453652930/home/nemo: invalid argument (Hint: try increasing the number of subordinate IDs in /etc/subuid and /etc/subgid)","context":"default"}
```

alexromanyuk: https://kibana-o11y.aws.us.stg.int.datarobot.com/app/r/s/05Ldj?auth_provider_hint=anonymous1

public_dropin_gpu_environments/nemollm_inference_ms/start_server.sh
```diff
@@ -32,25 +27,26 @@ if [ "${ENABLE_CUSTOM_MODEL_RUNTIME_ENV_DUMP}" = 1 ]; then
...
     --nemo_port=$NEMO_PORT \
-    --num_gpus=$(nvidia-smi -L | wc -l)
+    --num_gpus=$GPU_COUNT &
+
+
+echo
+echo "Starting DRUM server..."
+echo
+source /home/nemo/dr/bin/activate
+exec drum server --with-nemo-server "$@"
```
alexromanyuk, line 52: Perhaps I'll need to start NIM from instide of the DRUM, after the model is loaded. I'm experimenting with that, but it will be a separate PR.

elatt (approved):

public_dropin_gpu_environments/nemollm_inference_ms/start_server.sh
```diff
@@ -0,0 +1,52 @@
...
+# This is proprietary source code of DataRobot, Inc. and its affiliates.
+#
+# Released under the terms of DataRobot Tool and Utility Agreement.
+echo "Starting Custom Model environment with NIM"
+set -e
+
+GPU_COUNT=$(nvidia-smi -L | wc -l)
+echo "GPU count: $GPU_COUNT"
+
+if [ "$GPU_COUNT" -eq 0 ]; then
```
elatt, line 14: How does shell treat empty string, does that map to zero? I'm thinking of the case where smi binary doesn't even exist in image so you will end up with an empty string.

alexromanyuk: we counting lines
```
GPU_COUNT=$("" | wc -l)
echo "GPU count: $GPU_COUNT"
```
echoes
```
GPU count: 0
```

public_dropin_gpu_environments/nemollm_inference_ms/start_server.sh
elatt: Yeah, I forgot about that because I found too that it seems NIM just had one shot to start and have models there. Be careful of starting from DRUM though because I installed drum and stuff into dedicated venv due to package conflicts so from drum, you will want to deactivate venv before calling command (I think subprocess module supports pre-hook)

public_dropin_gpu_environments/nemollm_inference_ms/start_server.sh
alexromanyuk: > case where smi binary doesn't even exist

then startup script should exit, because of `set -e`.

elatt: I didn't think set -e took effect for captured commands (I.e. $(command...))

But I did a quick test and you are right, wc outputs zero so we are all good.

PR #964, opened 2024-04-01
[BUZOK-2294] Remove buzok from owners

tsdaemon:
## Summary
Remove buzok from owners of GenAI environment

## Rationale
Buzok no longer depends on GenAI custom environment: we have migrated to use an execution envrionment defined in our repository.

### DIFF
--- a/DRCODEOWNERS
+++ b/DRCODEOWNERS
@@ -27,8 +27,6 @@
 /custom_model_runner/datarobot_drum/resource/default_typeschema   @datarobot/core-modeling
 /public_custom_application_environments                           @datarobot/custom-models @datarobot/applications
 /public_dropin_environments/java_codegen                          @datarobot/custom-models
-/public_dropin_environments/python311_genai                       @datarobot/custom-models @datarobot/buzok
-/public_dropin_environments/python39_genai                        @datarobot/custom-models @datarobot/buzok
 /public_dropin_environments/python3_keras                         @datarobot/custom-models @datarobot/core-modeling
 /public_dropin_environments/python3_onnx                          @datarobot/custom-models
 /public_dropin_environments/python3_pmml                          @datarobot/custom-models @datarobot/core-modeling


scottp-ml (approved):

mxpoliakov (approved):

klichukb:
Does this mean that effectively the definition of these same environments moves to a different repo? I'm looking to understand if the definitions here are even needed, since maintaining them will fall on our team then.

klichukb:
Actually this creates more questions. When we ship onprem at this moment, we package environments from this repo, does Buzok distribute their environments now separately?

tsdaemon:
Good question @klichukb. I am not sure what is the history of those two envs: I suppose users are also using them for their custom models not generated in Buzok. I don't know whether or not we still need to have them, and who should maintain them if we do. @dr-nate-daly-pm @iamfeniak I need your help with that

tsdaemon:
> Actually this creates more questions. When we ship onprem at this moment, we package environments from this repo, does Buzok distribute their environments now separately?

Are we packaging these envs? I believe when we discussed the GenAI environments on-premise you said that these environments are not packaged, so we followed the approach you use for [custom templates](https://github.com/datarobot/datarobot-custom-templates): we implemented our own migration job which will automatically create or update an execution environment from definition in our repository.

tsdaemon:
Ok sorry, this was not discussed with you @klichukb, my memory betrays me. It was @liororama who suggested that we need to follow that approach.

klichukb:
Okay, sound reasonable, also the standard installation system will be reworked for onprem.

I would say merging this is fine, but we can consider moving this env to "example_dropin_environments", or similar directory, that doesnt include things that we automatically install/ship.

klichukb (approved):
@yakov-g what are your thoughts? Lets move the env to example dropin envs?

yakov-g:
@klichukb 
After we discussed this in person.
Let's do:
* delete Py3.9 GenAI and mark it as [DEPRECATED] on Staging/Prod
* move Py3.11 Gen AI to examples

@tsdaemon 
> we implemented our own migration job which will automatically create or update an execution environment from definition in our repository.

Do I understand correct that you are going to roll out Py3.11 GenAI on Staging/Prod by yourself?

tsdaemon:
@yakov-g yep, we are doing that already in our pipelines

yakov-g:
> @yakov-g yep, we are doing that already in our pipelines

This means, that if we keep this environment in current(DRUM) repository, its context will be very different from what is actually installed, as it is maintained in your repo.
So i believe we should just have a README and and a link to buzok repository.

tsdaemon:
I am good with moving it to examples: Buzok no longer rely or use this environment in any way, but since it was used by CFDS and users before, its still a good thing to have it available.

PR #968, opened 2024-04-05
[RAPTOR-10612] add log stacktrace when there is exception

yakov-g:

## Summary
We don't log exception stacktrace when DRUM is running in server mode.  Instead we start error server with only exception message.

Thus, add stacktrace logging to DRUM.


zohar-mizrahi (approved):

akshoop (approved):

yakov-g:
> Could you show an example snippet log or screenshot that shows the logged traceback text?

image

PR #985, opened 2024-04-17
[APPLS-2170] Rename custom apps streamlit env

kijimoshib:
Rename custom apps streamlit env

### DIFF
--- a/public_custom_application_environments/python39_streamlit/env_info.json
+++ b/public_custom_application_environments/python39_streamlit/env_info.json
@@ -1,6 +1,6 @@
 {
   "id": "6542cd582a9d3d51bf4ac71e",
-  "name": "[Experimental] Python 3.9 Streamlit",
+  "name": "[DataRobot] Python 3.9 Streamlit",
   "description": "Base image for Streamlit custom application with DR components package.",
   "programmingLanguage": "python",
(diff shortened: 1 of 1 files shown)


ben-cutler-datarobot (commented):
I wonder if it would make more sense to make a new datarobot env for the non-experimental container; for example I don't think we absolutely need to ship `wordcloud` or datarobot 3.0.2 while we are on 3.3.2

I also think changing an env name out from under customers could cause confusion. What do you think?

kijimoshib:
> I wonder if it would make more sense to make a new datarobot env for the non-experimental container; for example I don't think we absolutely need to ship `wordcloud` or datarobot 3.0.2 while we are on 3.3.2
> 
> I also think changing an env name out from under customers could cause confusion. What do you think?


I think updating dependencies could be done in following PR's and should include required testing passed.
For current moment we have tested environment and need to rename it accordingly.

zohar-mizrahi (approved):

kijimoshib:
> I wonder if it would make more sense to make a new datarobot env for the non-experimental container; for example I don't think we absolutely need to ship `wordcloud` or datarobot 3.0.2 while we are on 3.3.2
> 
> I also think changing an env name out from under customers could cause confusion. What do you think?

You’re right. CLI customers are using env name and this could cause confusion.  I'm going to close this.

PR #1033, opened 2024-05-06
[AGENT-5696] Add runtime-parameter-file to docker command

dulcardo:

## Summary

Add the mapping of runtime parameter file when running drum with `--docker` param, this  allows to provide runtime parameter file either as env var or argument.

now we can execute the following commands
using args:
```
drum score --docker cm_pps_66048257e5c67e60324307c7 --runtime-params-file /tmp/runtime.yaml --code-dir /tmp/model_dir/  --input /tmp/juniors_3_year_stats_regression_inference.csv
```

using env var:
```
 RUNTIME_PARAMS_FILE=/tmp/runtime.yaml drum score --docker cm_pps_66048257e5c67e60324307c7  --code-dir /tmp/mode_dir/  --input /tmp/juniors_3_year_stats_regression_inference.csv
```


custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -942,6 +943,21 @@ def _prepare_docker_command(self, options, run_mode, raw_arguments):
             )
             docker_cmd_args += " -p {port}:{port}".format(port=port)
 
+        if "runtime_params_file" in options and options.runtime_params_file:
```
eric-s-s, line 946: this is extremely suspicious.  don't we set all of this stuff in `enums.py`?

eric-s-s: see `datarobot_drum.drum.args_parser.CMRunnerArgsRegistry._reg_arg_custom_task_user_secrets` for an example.

also see: `unit.datarobot_drum.drum.test_args_parser.TestUserSecretsArgs.test_set_user_secrets_from_env_vars`

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -942,6 +943,21 @@ def _prepare_docker_command(self, options, run_mode, raw_arguments):
...
+            docker_cmd_args += ' -v "{}":{}'.format(
+                options.runtime_params_file, in_docker_runtime_parameters_file
+            )
+            DrumUtils.replace_cmd_argument_value(
+                in_docker_cmd_list,
+                ArgumentsOptions.RUNTIME_PARAMS_FILE,
+                in_docker_runtime_parameters_file,
+            )
+            # check if runtime params is provided through env vars
+            if "RUNTIME_PARAMS_FILE" in os.environ:
```
zohar-mizrahi, line 956: What would be the behavior if the user provides the runtime params file both as a command line argument and as an environment variable? IMHO, the command line argument should have precedence over the environment variable and therefore, you should add an `else` clause on top of this code block.

tests/functional/test_runtime_parameters.py
```diff
@@ -216,6 +218,10 @@ def _test_custom_model_with_runtime_params(
             cmd += f"--runtime-params-file {runtime_param_values_stream.name}"
         else:
             env["RUNTIME_PARAMS_FILE"] = runtime_param_values_stream.name
+
+        if docker:
```
eric-s-s, line 222: is this only testable through functional tests?  if there's no way to unit test this, then it would be worthwhile to create one.

eric-s-s (changes requested):
We are missing some test coverage.

1. please test that `runtime_params_file` is _always part of the `options`
2. please unit test this instead of functional test it.  we _should_ have some unit tests of creating the docker command.  and if we don't, then it's time to create some.

custom_model_runner/datarobot_drum/drum/drum.py
eric-s-s: we should already be getting this from the `options` please test that and remove this.

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -888,6 +888,7 @@ def _prepare_docker_command(self, options, run_mode, raw_arguments):
         in_docker_fit_output_dir = "/opt/fit_output_dir"
         in_docker_fit_target_filename = "/opt/fit_target.csv"
         in_docker_fit_row_weights_filename = "/opt/fit_row_weights.csv"
+        in_docker_runtime_parameters_file = "/opt/runtime_parameters.yaml"
```
eric-s-s, line 891: what does all of the `in_docker_` stuff mean?  do you mean that this is the where the runtime parameters file is mounted in the docker container?  if so, maybe these should change to `file_name_docker_mounted_location` or ... something?  asking chat gpt gives me:

- docker_mounted_params_file="/opt/params.yaml"
- docker_params_file_location="/opt/params.yaml"
- docker_params_file_mount_point="/opt/params.yaml"
- docker_params_file_path="/opt/params.yaml"
- docker_params_file_destination="/opt/params.yaml"

maybe like these but move `docker`

- params_file_docker_location
- params_file_location_in_docker

:shrug: not sure what the best answer is, but `in_docker_` was certainly confusing for a first-time reader.

custom_model_runner/datarobot_drum/drum/drum.py
eric-s-s: similar to the previous comment, the reliance on a string here is extremely suspicious.  and I'm pretty sure this is already done inside of the args parser.

dulcardo: the command line argument has precedence over the env var, and this is already part of the argument parsing, but what I need to add is the subcommand, so this is fixed, check :
* first, check if we need to update the current cmd , if so update the cmd by `DrumUtils.replace_cmd_argument_value`
* if is provided by env var update the `-e PARAM_FILE=/path/to/file.yaml`

custom_model_runner/datarobot_drum/drum/drum.py
dulcardo: I will have to do this renaming of all vars in separate PR, including a set of Unit test for each particular case

tests/unit/datarobot_drum/runtime_parameters/test_runtime_parameters.py
```diff
@@ -275,3 +277,140 @@ def test_setup_success(
 
                 actual_value = RuntimeParameters.get(param_name)
                 assert actual_value == expected_value
+
```
eric-s-s, line 280: `unit/datarobot_drum/runtime_parameters/test_runtime_parameters.py`  tests `datarobot_drum/runtime_parameters/runtime_parameters.py`  file.  these tests are in `drum.py` and so should be in the `test_drum.py` file.

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -942,6 +943,24 @@ def _prepare_docker_command(self, options, run_mode, raw_arguments):
             )
             docker_cmd_args += " -p {port}:{port}".format(port=port)
 
+        if (
```
eric-s-s, line 946: from our discussion use any of:

```python
        if getattr(options, ArgumentsOptions.RUNTIME_PARAMS_FILE, None):
            docker_cmd_args += f' -v "{options.runtime_params_file}":{in_docker_runtime_parameters_file}'
            DrumUtils.delete_cmd_argument(in_docker_cmd_list, ArgumentsOptions.RUNTIME_PARAMS_FILE)
            in_docker_cmd_list.extend([ArgumentsOptions.RUNTIME_PARAMS_FILE, in_docker_runtime_parameters_file])
```

dulcardo: `ArgumentsOptions.RUNTIME_PARAMS_FILE` -> "--runtime-params-file", while options contains "runtime_params_file".

I updated this to use 
```
if getattr(options, ArgumentOptionsEnvVars.RUNTIME_PARAMS_FILE.lower(), None):
```

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -942,6 +943,15 @@ def _prepare_docker_command(self, options, run_mode, raw_arguments):
             )
             docker_cmd_args += " -p {port}:{port}".format(port=port)
 
+        if getattr(options, ArgumentOptionsEnvVars.RUNTIME_PARAMS_FILE.lower(), None):
```
eric-s-s, line 946: i think that this doesn't _really_ work either.  I think it would good to have a method

```python
def get_arg_option(options: Namespace, arg_option: ArgumentOptions):
    value_in_namespace = arg_option.lstrip("-").replace("-", "_")
    return getattr(options, value_in_namespace, None)
```

and put this in args_parser and test it and then start using that method instead of hopes and prayers in the Namespace.

tests/unit/datarobot_drum/drum/test_args_parser.py
```diff
@@ -654,3 +659,16 @@ def test_read_triton_server_configs_success(
                 minimal_score_cmd_args,
                 is_env_variable,
             )
+
+
+class TestRuntimeParametersArgs:
+    @pytest.mark.parametrize("args", ["score_args", "server_args"])
+    def test_runtime_params_valid(self, request, args):
```
eric-s-s, line 666: nice!  please add for all of the positive cases

```
        CMRunnerArgsRegistry._reg_args_runtime_parameters_file(
            score_parser, perf_test_parser, server_parser, validation_parser
        )
```

custom_model_runner/datarobot_drum/drum/drum.py
eric-s-s: yeah.  this is definitely out of scope

tests/unit/datarobot_drum/drum/test_drum.py
```diff
@@ -495,3 +500,119 @@ def test_output_in_code_dir(self):
         output_code_dir = "/test/code/is/here/output"
         assert not output_in_code_dir(code_dir, output_other)
         assert output_in_code_dir(code_dir, output_code_dir)
+
+
+class TestRuntimeParametersDockerCommand:
```
eric-s-s, line 505: looks good :100:

tests/unit/datarobot_drum/drum/test_args_parser.py
dulcardo: done

elatt (approved):

tests/unit/datarobot_drum/drum/test_args_parser.py
```diff
@@ -654,3 +669,41 @@ def test_read_triton_server_configs_success(
...
+
+    @pytest.mark.parametrize("args", ["fit_args", "push_args"])
+    def test_get_options_invalid(self, request, args):
+        args = request.getfixturevalue(args)
+        options = get_args_parser_options(args)
+        arg_option = CMRunnerArgsRegistry.get_arg_option(
+            options, ArgumentsOptions.RUNTIME_PARAMS_FILE
+        )
+        assert arg_option is None
+        assert arg_option not in options
```
eric-s-s, line 694: this assert is `assert None not in options`.  did you mean `assert "runtime_params_file" not in options`?

eric-s-s (approved):
nice!

custom_model_runner/datarobot_drum/drum/drum.py
eric-s-s: keeping this unresolved because the similar code in `main.py` would be worth addressing in the future

tests/unit/datarobot_drum/drum/test_args_parser.py
dulcardo: ohh my bad, yes it `assert "runtime_params_file" not in options

PR #1050, opened 2024-05-31
[AGENT-5795] Update vLLM env for large LLMs

elatt:

## Summary
- Pass gpu count to vLLM server
- Add `autoawq` library to vllm env
- Bump version of vllm env
- Allow setting `--model` param via engine config file

## Rationale
To run larger LLMs (e.g. 70b+ param) one needs to split the model across multiple GPUs. This adds automatically handling of that so if we detect multiple GPUs, we configure vLLM accordingly. I'm also adding the `autoawq` package as it is used for quantization which is related running large LLMs (on smaller GPUs with minimal degredation to quality). Finally, users may want to set CLI flags via config file instead of runtime params so only hard code the most core params.


datarobot-devinfragit3:

rvorobii (approved):

dulcardo (approved):

PR #1054, opened 2024-06-12
Add short-circuit to NIM env

elatt:

## Summary
- Update health check in Nemo predictor to return 513 if NIM server has crashed
- Bump DRUM version to v1.11.5
- Release new envs
- Add notes to add short-circuit to Triton env

## Rationale
Returning a 513 after the NIM server has crashed allows us to fail-fast and return status back to the user.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/gpu_predictors/triton_predictor.py
+++ b/custom_model_runner/datarobot_drum/drum/gpu_predictors/triton_predictor.py
@@ -114,12 +114,14 @@ class TritonPredictor(BaseLanguagePredictor):
         return self.health_check()
 
     def health_check(self):
+        # TODO: would be good to respond with 513 status code if Triton server has crashed or
+        #   if we can detect some other terminal error relating to loading the model provided.
         try:
             triton_health_url = f"{self.triton_host}:{self.triton_http_port}/v2/health/ready"
             response = requests.get(triton_health_url, timeout=5)
             return {"message": response.text}, response.status_code
         except Timeout:
-            return {"message": "Timeout waiting for Nemo health route to respond."}, 503
+            return {"message": "Timeout waiting for Triton health route to respond."}, 503
 
     def _transform(self, **kwargs):
         raise DrumCommonException("Transform feature is not supported")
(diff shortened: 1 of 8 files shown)


custom_model_runner/datarobot_drum/drum/gpu_predictors/triton_predictor.py
```diff
@@ -114,12 +114,14 @@ def readiness_probe(self):
         return self.health_check()
 
     def health_check(self):
+        # TODO: would be good to respond with 513 status code if Triton server has crashed or
+        #   if we can detect some other terminal error relating to loading the model provided.
```
elatt, line 118: @alexromanyuk from your experience, is there an easy way to add this to Triton env? My first question would be, how does Triton behave if it sees an invalid model in its model directory:
- Does it terminate
- if not, are there any health routes that can differentiate between it starting up vs. being ready but having no models to serve?
- any other options to detect a terminal error?

alexromanyuk: seems I missed this question. may be not relevant now, but as far as I remember - Triton will terminate if there are no models to serve or that single model can't be loaded.

> are there any health routes that can differentiate between it starting up vs. being ready but having no models to serve?

The `/health/live` - server started and the `/health/ready` - models loaded and ready for inference. see here 
https://docs.nvidia.com/deeplearning/triton-inference-server/archives/triton_inference_server_1120/triton-inference-server-guide/docs/http_grpc_api.html#health

alexromanyuk (approved):

custom_model_runner/datarobot_drum/drum/gpu_predictors/triton_predictor.py
alexromanyuk: and this one
> Performing an HTTP GET to /api/status returns status information about the server and all the models being served. Performing an HTTP GET to /api/status/\<model name\> returns information about the server and the single model specified by <model name>.

dulcardo (approved):

PR #1059, opened 2024-06-18
[CODEGEN-1913] Fix nginx permissions

baekdahl:

## Summary
Change `chmod` to be recursive and add `/var/log/nginx` to make sure that nginx can write to required paths when running as non-root user.
Nginx needs to be able to write to `/var/lib/nginx/temp` and `/var/log/nginx` but currently we only add write permissions for `/var/lib/nginx`. 

## Rationale
We want to be able to run nginx when DRUM is running in production mode.


yakov-g (approved):

DrewParallel (approved):

yakov-g:
I had to rebase on master and force push

PR #1060, opened 2024-06-18
[RAPTOR-10986] Set proper dependency for numpy in the runner and java-codegen env

zohar-mizrahi:
## Summary
I turned out that `datarobot-drum` is not compatible with `numpy>=2.0.0`. The latter was release on Jun-16, 2024. Ever since then, the https://ci1.devinfra.drdev.io/job/Custom_Models_drop_in_environment_tests/ started to fail.

It is required to restrict the numpy version to <2.0.0


yakov-g (approved):

zohar-mizrahi:
jarvis I manually verified this

PR #1068, opened 2024-06-26
[CODEGEN-1917] Copy mlpiper into drum

baekdahl:

## Summary
* Copy mlpiper python parts into drum.
* Remove mlpiper dependency.
* Remove some mlpiper tests that are broken because mlpiper-java wasn't copied.

## Rationale
We want to move away from mlpiper in the future because it is no longer maintained and a lot of the functionality is not used in drum. 
This can be seen as the first step of getting rid of mlpiper, copying it into drum to enable easier refactoring.


akshoop:
CC @eric-s-s @scottp-ml for awareness on behalf of Core Modeling

yakov-g (approved):

yakov-g:
Super cool!
Thank you!

eric-s-s:
I am not clear how this allows for easier change.  the ticket: https://datarobot.atlassian.net/browse/CODEGEN-1917 does not have any details or a link to an epic or a link to a PBMP or a tech spec.  Please add details to the ticket and put it in an epic that gives more context for this change.  This looks like it will be a big  lift and I'd like to know more.

Also, 100% in favor of the general idea.

baekdahl:
@eric-s-s I have updated the ticket with some more information. The associated PBMP is [PBMP-6561](https://datarobot.atlassian.net/browse/PBMP-6561).

> I am not clear how this allows for easier change

In my opinion, making changes are faster when working in one repo. When mlpiper is in DRUM, we can also make changes that will break other uses of mlpiper than what is needed for DRUM.

>This looks like it will be a big lift and I'd like to know more.

The big lift of actually refactoring DRUM is out of scope for this project.

eric-s-s (changes requested):
For a large change like this, I would like to see the reverse action done.

1. build the functionality that will replace similar mlpiper code.
2. move functionality from using mlpiper to new pieces
3. remove mlpiper

Adding 25,000 lines seems like it should not be a change here.  Bringing in this much code seems dangerous.  There's no way for anyone to review this code and see what changes are happening.  Then, when you go to update MLPiper code, there's no way to review that either as you would need complete knowledge of what's added and missing in order to understand the changes.  If the actual code we are importing is small, then we don't need to do this.  If we are importing a large amount of mlpiper, then we should update the mlpiper repo.  Moving off of mlpiper could be a very good move, but pulling in the entire library to do that is not really moving off of mlpiper.

scottp-ml (changes requested):
I do not like the idea of moving a library into DRUM.  This seems like the wrong move.

PR #1083, opened 2024-07-08
[RAPTOR-11093] bump envs scikit-learn to 1.3.2

yakov-g:

## Summary
scikit-learn==1.0.2 in our public envs is quite old, so updating it to 1.3.2

There are some e2e tests that use artifacts, so I had to retrain them with the bumped version.
Added comments and adjusted some code.
The most of the existing artifacts/custom.py were ~2 years old.

DRUM bump is not required.


shopin32 (approved):

tsdaemon (approved):
LGTM

rvorobii (approved):

tests/fixtures/transform_custom.py
```diff
@@ -29,3 +36,55 @@ def transform(X, transformer, y=None):
         )
     else:
         return pd.DataFrame(transformed)
+
```
eric-s-s, line 39: can we remove the top-level functions if we have this?

eric-s-s: same question for the other fixtures.  so in this one I'm asking if we can delete L#18-L#38

tests/functional/test_fit_per_framework.py
```diff
@@ -712,6 +716,8 @@ def test_fit_sh(
                 env["CLASS_LABELS_FILE"] = f.name
 
         if parameters:
+            # I don't understand why hypersearch params are taken from a separate json
```
eric-s-s, line 719: This looks like a good discussion to have.  let's do it in conversation here, but please remove this comment when we are done.

scottp-ml (changes requested):

tests/fixtures/transform_custom.py
```diff
@@ -29,3 +36,55 @@ def transform(X, transformer, y=None):
...
+        artifact_directory: str
+            Path to the directory to save the serialized artifact(s) to.
+
+        Returns
+        -------
+        cls
+            The deserialized object
+        """
+        # only one artifact is expected in the folder: sklearn_transform.pkl or sklearn_transform_dense.pkl
+        artifact_file = glob.glob(os.path.join(artifact_directory, "sklearn_transform*.pkl"))[0]
```
eric-s-s, line 60: but how will it know which one to choose?  does it matter if it's dense or not?  if not, why not excplicitly choose one.

tests/fixtures/transform_custom.py
```diff
@@ -29,3 +36,55 @@ def transform(X, transformer, y=None):
         )
     else:
         return pd.DataFrame(transformed)
+
+
+class CustomTask(TransformerInterface):
+    @classmethod
+    def load_task(cls, artifact_directory):
```
eric-s-s, line 43: can we solve this issue by having unique artifact directories instead of unique file names?

so something like

`tests/fixtures/drop_in_model_artifacts/sklearn_bin/<default_filename>`

tests/fixtures/transform_custom.py
eric-s-s: I _think_ this means that you've defined the task twice in a single file and there's a real question of what will happen when you try to run this, and with two things here, that behavior could change at any time.

yakov-g: Sorry, I don't understand what `twice` refers to.
If I do:
`drum fit --code-dir task_templates/1_transforms/3_python3_sklearn_transform --input tests/testdata/10k_diabetes_sample.csv --target-type transform --target readmitted --output ~/custom_task`

In the output dir I get:
ls -la  ~/custom_task/
total 44
-rw-rw-r-- 1 yakov.goldberg yakov.goldberg  1762 Jul  9 15:37 create_transform_pipeline.py
-rw-rw-r-- 1 yakov.goldberg yakov.goldberg  1762 Apr 14 23:22 custom.py
-rw-rw-r-- 1 yakov.goldberg yakov.goldberg 22703 Jul 10 10:08 drum_artifact.pkl
-rw-rw-r-- 1 yakov.goldberg yakov.goldberg  3099 Nov 16  2023 model-metadata.yaml
drwxrwxr-x 2 yakov.goldberg yakov.goldberg  4096 Jul 10 10:08 __pycache__
-rw-rw-r-- 1 yakov.goldberg yakov.goldberg   138 Jul 10 10:08 README.md


drum score works when i predict on that folder.
custom.py contains CustomTask

tests/fixtures/transform_custom.py
```diff
@@ -29,3 +36,55 @@ def transform(X, transformer, y=None):
         )
     else:
         return pd.DataFrame(transformed)
+
+
+class CustomTask(TransformerInterface):
```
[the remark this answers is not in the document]
yakov-g: @scottp-ml I commented ^

tests/functional/test_fit_per_framework.py
yakov-g: Let's do this...
So I still don't understand why json is needed there :)

tests/fixtures/transform_custom.py
yakov-g: This default_filename seems wrong to me, or I just don't know/understand some logic behind this.

If adjustment is needed, I'd better copy `custom_artifact_name` to `default_filename` when constructing code_dir for a test.

tests/fixtures/transform_custom.py
yakov-g: During the test only artifact is copied into the code_dir.

tests/fixtures/transform_custom.py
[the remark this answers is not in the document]
scottp-ml: If you are going to add a version with the interface you should get rid of the other version above.

tests/fixtures/transform_custom.py
scottp-ml: The custom task is defined twice in one file.  Once with the old style hooks and again with the new interface class.  That is confusing and I am actually shocked that it works at all.

yakov-g: oh,
why shouldn't it work... drum code just loads `CustomTask`

tests/fixtures/transform_custom.py
scottp-ml: In normal usage of custom tasks users shouldn't need to worry about the filename for their pickled task, or even that it is pickled. Normally an empty output directory is used to dump the results of fit, and then that can be moved to wherever predict happens.  We also have a `save` and a `load` method for more compicated situations such as serializing keras or pytorch models.  The way we have these pickle files just saved and reused is an abnormal use case specific to how these tests were implemented.

tests/fixtures/transform_custom.py
scottp-ml: Yes, you can remove the old method that isn't used anymore if you are using the interface class.

tests/functional/test_fit_per_framework.py
eric-s-s: this looks like it's setting the actual values to run hyper parameters on.  model metadata contains the definition for the hyper parameters with  allowed values, etc.  but this actually sets the hyperparameters search space.

eric-s-s: it's very much like launch configuration.  the model metadata tells you what launch configs are allowed and what the defaults are, and then, when running drum, you can pass in actual set values via a json file.

tests/fixtures/transform_custom.py
yakov-g: removed old methods

tests/functional/test_fit_per_framework.py
yakov-g: updated comment

tests/fixtures/transform_custom.py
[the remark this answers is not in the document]
yakov-g: Got rid of standalone methods

tests/functional/test_fit_per_framework.py
```diff
@@ -716,8 +716,8 @@ def test_fit_sh(
                 env["CLASS_LABELS_FILE"] = f.name
 
         if parameters:
-            # I don't understand why hypersearch params are taken from a separate json
-            # instead of the model-metadata in python3_sklearn_binary_hyperparameters
+            # model-metadata configs which values are allowed for hyperparameters.
```
eric-s-s, line 719: can we just remove this comment?  It doesn't add new information.

yakov-g: removed

tests/fixtures/transform_custom.py
yakov-g: So I changed tests to rename artifacts to a provided name. For these test cases I provide drum_artifact.pkl

scottp-ml (approved):

PR #1088, opened 2024-07-19
[RAPTOR-11194] Remove dependency from requirements_test.txt

rvorobii:

## Summary
Remove unused test dependency (it's causing installation issues).


yakov-g (approved):

yakov-g:
These tests are not run now

yakov-g:
I'll create if it is possible to move it to a separate file: requrements_tests_e2e

requirements_test.txt
```diff
@@ -1,5 +1,4 @@
 datarobot==3.1.0
-datarobot-bp-workshop==0.2.6
```
eric-s-s, line 2: @yakov-g @scottp-ml starting this as an arbitrary code comment so that we can have a conversation instead of random top-level messages

> These tests are not run now

why aren't they run?

yakov-g: They are not run, because we switched to harness, so now we don't install DR during tests and can not run e2e tests.

They were turned off together with pipeline controller. 
I knew that inference models will be still somehow covered by: [Custom_Models_drop_in_environment_tests](https://ci1.devinfra.drdev.io/job/Custom_Models_drop_in_environment_tests/4629/)
I admit, this is my mistake.

Regarding inference models: E2E tests didn't add much value in terms of DRUM testing, because they do the following: start DR, install env, run predictions, which is kind of making sure that dropin envs are installable and functioning.

In functional drum tests we build an env image and run tests inside of it, so we make sure env has all the deps.
We should improve this testing by adding cases where we run `start_server.sh`.

`_Probably_` we should have daily tests which install dropin envs in DR and run predictions, to make sure that envs are fully functioning at every given moment (but not sure).

scottp-ml: I think it is safe at this point to actually remove the e2e test for task templates, and then the dependency can be removed.

eric-s-s: so it looks like this is part of wider change that involves updating these tests and/or deleting them.  this probably should have been a wider discussion when we switched off the tests.  Both @scottp-ml and I would be in favor of killing e2e tests, but we should coordinate about what we have and what our needs are.  I'd be in favor of closing this PR and working out what we need to do for e2e tests and then updating/deleting this as part of a wider change.

eric-s-s: is there a particular need to remove this dependency?

yakov-g: Let me add some changes to this PR, and you maybe you'll approve.

yakov-g: Yes.
This still runs as part of DTS. Only inference models.

yakov-g: @eric-s-s @scottp-ml 
I separated requirements files for inference and training.
Inference still runs as a part of DTS. https://ci1.devinfra.drdev.io/job/Custom_Models_drop_in_environment_tests/4632/display/redirect

scottp-ml (approved):

eric-s-s (approved):
lgtm

PR #1089, opened 2024-07-19
[RAPTOR-11201] cleanup final exception logging

yakov-g:

## Summary
Clean up final exception logging a bit:
* printing exception type/message does't bring any value, just print traceback.
* remove senseless  "Looks like there is a problem." message and an instruction  to use --verbose.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/runtime.py
+++ b/custom_model_runner/datarobot_drum/drum/runtime.py
@@ -42,15 +42,7 @@ class DrumRuntime:
             # exception occurred before args were parsed
             return False  # propagate exception further
 
-        msg = "Looks like there is a problem."
-        if not self.options.verbose:
-            msg += f" To get more output information try to run locally(not in DataRobot) with: '{ArgumentsOptions.VERBOSE}'."
-        logger_drum.warning(colored(msg, "yellow"))
-
-        if exc_value:
-            logger_drum.error(exc_value)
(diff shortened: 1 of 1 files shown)


zohar-mizrahi (approved):

custom_model_runner/datarobot_drum/drum/runtime.py
```diff
@@ -42,15 +42,7 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
...
 
-        msg = "Looks like there is a problem."
-        if not self.options.verbose:
-            msg += f" To get more output information try to run locally(not in DataRobot) with: '{ArgumentsOptions.VERBOSE}'."
-        logger_drum.warning(colored(msg, "yellow"))
-
-        if exc_value:
-            logger_drum.error(exc_value)
-        logger_drum.error(exc_type)
-        logger_drum.error("".join(traceback.format_list(traceback.extract_tb(exc_traceback))))
```
akshoop, line 53: If you remove line seems like `exc_traceback` will not get printed, or would it get printed somewhere else?

yakov-g: it will be printed because I explicitly call logger's `exception`  method, which takes `message` as an argument, which I provide as empty.

akshoop (approved):
LGTM

PR #1090, opened 2024-07-23
[CODEGEN-1924] Add chat hook

baekdahl:

## Summary
This adds a new hook `chat` which can be implemented by models to add support for [OpenAIs chat completion API](https://platform.openai.com/docs/api-reference/chat/create). A new route `/chat/completions/` route is added to `drum server` which will invoke this chat hook.

Exception handling in drum server is changed to allow raising HTTPException which will get serialized to JSON. This adds possibility for Custom Model to raise an error when a bad request is received and have that returned as a HTTP 4xx instead of only HTTP 500 response which is currently the case.

Support for Custom Model testing will be added in [CODEGEN-1925](https://datarobot.atlassian.net/browse/CODEGEN-1925).

## Rationale
We want to add Chat API support for Custom Models so we can improve GenAI applications with support for features like chat history and streaming response.

[CODEGEN-1925]: https://datarobot.atlassian.net/browse/CODEGEN-1925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


requirements_test_unit.txt
```diff
@@ -1,3 +1,4 @@
 pytest
 responses
 scikit-learn==1.3.2
+openai
```
akshoop, line 4: Consider pinning to specific version of `openai`

custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
```diff
@@ -746,6 +747,9 @@ def predict_unstructured(self, model, data, **kwargs):
         PythonModelAdapter._validate_unstructured_predictions(predictions)
         return predictions
 
+    def chat(self, model, completion_create_params):
```
akshoop, line 750: nit, more verbose param name:
```suggestion
    def chat(self, model, chat_completion_create_params):
```

akshoop (approved):
Overall LGTM from Core Modeling

akshoop:
FYI @eric-s-s @scottp-ml for heads up and awareness

custom_model_runner/datarobot_drum/resource/components/Python/prediction_server/prediction_server.py
```diff
@@ -216,6 +232,11 @@ def stats():
             self._stats_collector.stats_reset()
             return ret_dict, HTTP_200_OK
 
+        @model_api.errorhandler(HTTPException)
```
klichukb, line 235: Can you please explain why is it useful to route http exceptions as 4xx to user?

A 4xx back to the user means that user's request to the custom model has lead to the problem (http exception), so user should fix the request.

This will in general capture all http exceptions, even if its non-openai related code and an unexpected http exception, which will still return to the user as 4xx, whereas it should probably be 500 in that case.

baekdahl: Just to clarify, all `HTTPException`s won't be returned as 4xx, they will return the status code associated with the exception so if e.g. `InternalServerError` is raised, that would return 500. 

The motivation is that user code can then return an appropriate error by raising the relevant subclass of `HTTPException`.

requirements_test_unit.txt
baekdahl: Will do

custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
baekdahl: The param name was chosen to align with [type names in openai library](https://github.com/openai/openai-python/blob/1ed0e35679438cc7409f2e3aeb36d637b3c579c0/src/openai/types/chat/completion_create_params.py#L282):

akshoop: Got it 👍

custom_model_runner/datarobot_drum/resource/components/Python/prediction_server/prediction_server.py
klichukb: Does this mean that if there is an 401 http exception with the POST /predict to a custom model, the /predict would propagate back HTTP 401 to the user?

baekdahl: Yes, if `Unauthorized` is raised somewhere it would. Are you thinking of a specific situation?

baekdahl: I need to be able to return 400 from the chat hook in case the request was rejected, e.g. if the input was blocked by a guard model. 

An alternative could be to have one specific exception for this case which would be added to the errorhandler instead of `HTTPException`

klichukb: Sounds good to me, if you have a specific case in mind, I would start with this.

Would do you think about the following idea:

```
class PassthroughHttpException(HttpException):
    pass
```

then

```
    def _handle_exception(self, ex):
        self._logger.error(ex)

        # pass through HTTP errors
        if isinstance(ex, PassthroughHTTPException):
            return jsonify(error=ex.description), ex.code
        ... # else
```

then wherever you reach out outside and can generate this:

```
try:
   ... do stuff
except HttpException as exc:
    if exc.code == 400:
        raise PassthroughHTTPException(.... wrap exc...)
    reraise
```

this can live in drum and be generic

baekdahl: Thanks. I'll come up with something similar

baekdahl: Did it a bit different - maybe even simpler. Let me know what you think

klichukb: How would DrumBadRequest be triggered? Manually by the user in the custom.py (I can only see an example in the test). Does this mean user needs to wrap their code to raise this specific error? (and document this)

baekdahl: Yes, the user will trigger this manually - It's not that they would be calling some function/library that would raise an `HTTPException` if that is what you are referring to, but I can see why I might be been giving that impression.

klichukb: I can see the idea, but my concern is that this is a new concept that model coders will need to be documented for DRUM, and also, it only works within chat()

If you raise this error in score(), score_unstructured() it will act as a regular exception. Maybe if we're introducing a new concept, at least provide the same logic in score and score_unstructured(). Its also a question whether this would be only supported in Python.

cc @yakov-g

baekdahl: I can add the same exception handling to score, score_unstructured but it sounds like you would rather that it is done differently?

Do you have any suggestion for a different way to return 400 from chat hook?

klichukb: Overall I like the idea of turning regular http errors into respective responses.

How about supporting passing through the exceptions for the native HTTPExecution class, but simply based on the response code.

If this is a HTTPException 400 - pass through to the user, otherwise act same as now. This would let user use the regular HTTPException class.

This would be support in score(), score_unstructured() as well that if you raise this error, it will be properly returned to the user.

Later if we need to support other codes returned, we can expand that. What do you think?

klichukb: I think this would be your original suggested code but only for 400 for now, but also the same behavior in score() and score_unstructured().

baekdahl: Good point. I'll try to do that then.

yakov-g: What about to start moving in the direction to implement interfaces for hooks and define custom.py using OOP as it is already done for CustomTasks.

https://github.com/datarobot/datarobot-user-models/blob/master/task_templates/1_transforms/3_python3_sklearn_transform/custom.py
https://github.com/datarobot/datarobot-user-models/blob/master/custom_model_runner/datarobot_drum/custom_task_interfaces/custom_task_interface.py
https://github.com/datarobot/datarobot-user-[models/blob/master/custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py ](https://github.com/datarobot/datarobot-user-models/blob/master/custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py#L230)

what if add smth like this:
```
interfaces collection in drum
class ChatInterface():
   @abstract_method()
   def chat()

   def do_chat()
        try:
           chat()
        except HHTPException()
           raise DrumError...

custom.py
 -----------
class MyChat()
    def chat()
          do_stuff()
```

Seems this may help answering multiple questions:
1. Cleaner interface for customers.
2. Easier, cleaner, more structured documentation.
3. Better separation: one interface may support only `chat`, another only `score` 
4. Exceptions may be seamlessly handled in a base class
etc.

Not related to this PR:
We should build a DRUM development roadmap. 
I think there are four teams who currently works or recently worked.
 on DRUM.

yakov-g: @klichukb 
Regarding your question: whether it should be supported only in Python.
It seems like everything moves towards Py-only way:
- textgeneration is not supported by R
- guard lib is not implemented for R (because of the previous reason)
- mlops R implementation was dropped

So if we implement interfaces for Py hooks, we could improve error handling for Py only, leaving R with legacy hooks.
Overall we need to get Product input on Rlang usage/vision.

I'm preparing an overview of DRUM condition/technical debts

custom_model_runner/datarobot_drum/resource/components/Python/prediction_server/prediction_server.py
```diff
@@ -202,6 +205,21 @@ def predict_unstructured():
             finally:
                 self._post_predict_and_transform()
 
+            return (response, response_status)
+
+        @model_api.route("/chat/completions", methods=["POST"])
```
alexromanyuk, line 210: @baekdahl , 
I think, we should also expose the `/v1/chat/completions`, in addition to the `/chat/completions`. Because it seems to be a part of OpenAI API standard. It's possible that in the future they'll change API to v2 and we'll have to be backwards compatible.

Predictions Gateway is configured to proxy either request:
```
# base_url="https://staging.datarobot.com/api/v2/deployments/669917cd46a0f93cd756e53d/"
base_url="https://staging.datarobot.com/api/v2/deployments/669917cd46a0f93cd756e53d/v1/"

client = OpenAI(
    base_url=base_url,
    api_key=os.environ.get('DATAROBOT_API_TOKEN'),
    http_client=httpx.Client(follow_redirects=True)
)



completion = client.chat.completions.create(
  model="generic_llm",
  messages=[
    {"role": "system", "content": "You are a stadup comic."},
    {"role": "user", "content": "Tell me a joke about programming"},
    {"content": "Why did the programmer quit his job?\n\nBecause he didn't get arrays! (get a raise)", "role": 'assistant'},
    {"role": "user", "content": "Can you write a longer joke on the same topic?"},
  ]
)

print(completion.choices[0].message)
```

baekdahl: Makes sense. I've added the `v1` route as well.

custom_model_runner/datarobot_drum/resource/components/Python/prediction_server/prediction_server.py
zohar-mizrahi: FWIW - the best for us would be to keep the approach of encapsulating the custom mode's exception with the 5xx status code. It clearly indicates that the error came from the custom model rather than our system.
However, it's may be a non standard way for clients. So, a special handling for the specific 400 error code seems to be fine for me but needs to be tested carefully (I assume that the chat query doesn't go through the predictions server, correct me if I'm wrong). Given the special meaning of the 400 status code (.e.g server is not able to process the query due to a client's invalid request - syntax error or invalid parameters, etc) I think it may work.

baekdahl: > . So, a special handling for the specific 400 error code seems to be fine for me but needs to be tested carefully (I assume that the chat query doesn't go through the predictions server, correct me if I'm wrong)

Could you elaborate on what you mean by "tested carefully" ? I have one for exception [here](https://github.com/datarobot/datarobot-user-models/pull/1090/files#diff-fa761b5019598affba12ea617aa16e96226808fd28807686db20d60677a94c73R188).

You are right, chat requests are bypassing prediction server.

zohar-mizrahi: As for careful testing I mean to verify all the paths in the backend that handle the responses from DRUM and make sure nothing is broken. Also, you may need to go through the documentation in DRUM and update wherever necessary. I think there's also related documentation in DataRobot about DRUM. cc: @yakov-g .

baekdahl: So this would be in the predictions API or also other places?

zohar-mizrahi: I can't find it in DataRobot repo. Maybe @yakov-g can help.

As for the new hook, you may need to add the related information [here](https://github.com/datarobot/datarobot-user-models/blob/master/DEFINE-INFERENCE-MODEL.md)

baekdahl: I have added some info to that readme file

yakov-g: Hooks related docks are here:
https://github.com/datarobot/datarobot-docs/blob/021eef9632375ae536f8944cf17bfb1a7b9a6004/datarobot_docs/en/mlops/deployment/custom-models/custom-model-assembly/structured-custom-models.md?plain=1#L2

Please talk to @n-aylward , but I believe no need to add docs while feature is not public, but surely worth creating a doc ticket for the project.

baekdahl: I have mentioned this in the [DOC ticket](https://datarobot.atlassian.net/browse/DOC-7037) associated with the project.

yakov-g (approved):

kindofluke:
@baekdahl should https://github.com/datarobot/datarobot-user-models/blob/ac3ca9530d8bd627a2d52964f06a1d07bccc2a96/custom_model_runner/drum_server_api.yaml#L186 this also be updated to have the chat hook?

baekdahl:
@kindofluke https://github.com/datarobot/datarobot-user-models/pull/1119

kindofluke:
Thanks

PR #1091, opened 2024-07-23
[CODEGEТ-1918] Copy RESTfulComponent from mlpiper to allow support configuration of uwsig threads

yura-ant:

## Summary
Copy  RESTfulComponent, UwsgiBroker, WSGI_INI_CONTENT from mlpiper and use it version instead of mlpiper versions:
https://github.com/datarobot/mlpiper/blob/master/mlpiper-py/mlpiper/components/restful_component.py
https://github.com/datarobot/mlpiper/blob/master/mlpiper-py/mlpiper/components/restful/uwsgi_broker.py
https://github.com/datarobot/mlpiper/blob/master/mlpiper-py/mlpiper/components/restful/uwsgi_ini_template.py

Functionality is not changed but we have ability to modify them now to allow improve configuration of uWSGI in separate PR

### DIFF
--- a/custom_model_runner/datarobot_drum/resource/components/Python/uwsgi_component/restful_component.py
+++ b/custom_model_runner/datarobot_drum/resource/components/Python/uwsgi_component/restful_component.py
@@ -1,0 +1,350 @@
+"""
+The RESTfulComponent class is used to create RESTful model serving components.
+It is an abstract class that is supposed to be extended and the sub-class is
+supposed to implement its abstract methods.
+
+The class provide the core functionality of the model restful serving, which
+includes setup and preparation of the environment, configurations as well as
+executing of required services (.e.g uwsig, nginx)
+"""
+
+import abc
+import os
+import socket
+import tempfile
+import time
+import logging
+import subprocess
+import sys
+
+from mlpiper.pipeline import java_mapping
+from mlpiper.model.model_env import ModelEnv
+from mlpiper.ml_engine.rest_model_serving_engine import RestModelServingEngine
+from mlpiper.common import constants
+from mlpiper.components import parameter
+from mlpiper.common.mlpiper_exception import MLPiperException
+from mlpiper.components.connectable_component import ConnectableComponent
+from mlpiper.components.restful.constants import (
+    SharedConstants,
+    ComponentConstants,
+    UwsgiConstants,
+    NginxConstants,
+)
+from mlpiper.components.restful.constants import RestfulConstants
+from mlpiper.components.restful.uwsgi_broker import UwsgiBroker
+from mlpiper.components.restful.nginx_broker import NginxBroker
+from mlpiper.components.restful.flask_route import FlaskRoute
+from mlpiper.components.restful.metric import Metric, MetricType, MetricRelation
+
+
+class RESTfulComponent(ConnectableComponent):
+    _uuid_engine = None
+    _stats_path_filename = None
+    _stats_count = 0
+
+    def __init__(self, engine):
+        # When this component is created from uwsgi entry point, engine param is None.
+        # So check for uwsgi installation only once from pipeline/component creation,
+        # when engine is not None.
+        if engine is not None:
+            if subprocess.run(
+                [sys.executable, "-m", "pip", "show", "uwsgi"],
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL
+            ).returncode != 0:
+                err_msg = "'uwsgi' package is required. " \
+                          "If you are running on a Windows system, " \
+                          "you may not be able to use RESTful component. " \
+                          "Check how to install uwsgi on Windows. " \
+                          "To install uwsgi on a Mac/Linux system, call " \
+                          "`pip install uwsgi` or `pip install mlpiper[uwsgi]`"
+                raise MLPiperException(err_msg)
+
+        super(RESTfulComponent, self).__init__(
+            engine if engine else RestModelServingEngine("uwsgi-context")
+        )
+        self._dry_run = False
+        self._wsgi_broker = None
+        self._nginx_broker = None
+        self._wid = None
+        self._wuuid = None
+
+        self._total_stat_requests = Metric(
+            "mlpiper.restful.stat_requests",
+            title="Total number of stat requests",
+            metric_type=MetricType.COUNTER,
+            value_type=int,
+            metric_relation=MetricRelation.SUM_OF,
+        )
+
+    def set_wid(self, wid):
+        self._wid = wid
+
+    def get_wid(self):
+        return self._wid
+
+    def set_wuuid(self, wuuid):
+        self._wuuid = wuuid
+
+    def _validate_output(self, objs):
+        pass
+
+    def _post_validation(self, objs):
+        pass
+
+    def configure(self, params):
+        super(RESTfulComponent, self).configure(params)
+
+    def _materialize(self, parent_data_objs, user_data):
+        monitor_info = {
+            UwsgiConstants.MONITOR_ERROR_KEY: None,
+            UwsgiConstants.MONITOR_THREAD_KEY: None,
+        }
+        self._setup(self._ml_engine.pipeline_name, monitor_info)
+        self._wait_and_monitor_errors(monitor_info)
+
+    def _setup(self, pipeline_name, monitor_info):
+        target_path = tempfile.mkdtemp(
+            dir=ComponentConstants.TMP_RESTFUL_ROOT,
+            prefix=ComponentConstants.TMP_RESTFUL_DIR_PREFIX,
+        )
+        os.chmod(target_path, 0o777)
+
+        fd, stats_path_filename = tempfile.mkstemp(
+            dir=ComponentConstants.TMP_RESTFUL_ROOT,
+            prefix=ComponentConstants.TMP_RESTFUL_DIR_PREFIX,
+        )
+        os.chmod(stats_path_filename, 0o777)
+
+        self._logger.debug("Path for stats {}".format(stats_path_filename))
+
+        shared_conf = {
+            SharedConstants.TARGET_PATH_KEY: target_path,
+            SharedConstants.SOCK_FILENAME_KEY: UwsgiConstants.SOCK_FILENAME,
+            SharedConstants.STATS_SOCK_FILENAME_KEY: UwsgiConstants.STATS_SOCK_FILENAME,
+            SharedConstants.STANDALONE: self._ml_engine.standalone,
+            SharedConstants.STATS_PATH_FILENAME_KEY: stats_path_filename,
+        }
+
+        log_format = self._params.get(
+            ComponentConstants.LOG_FORMAT_KEY, ComponentConstants.DEFAULT_LOG_FORMAT
+        )
+
+        log_level_param = self._params.get(
+            ComponentConstants.LOG_LEVEL_KEY, ComponentConstants.DEFAULT_LOG_LEVEL
+        ).lower()
+        log_level = constants.LOG_LEVELS.get(log_level_param, logging.INFO)
+        self._logger.debug(
+            "log_level_param: {}, log_level: {}, level_constants: {}".format(
+                log_level_param, log_level, constants.LOG_LEVELS
+            )
+        )
+
+        stats_reporting_interval_sec = self._params.get(
+            ComponentConstants.STATS_REPORTING_INTERVAL_SEC,
+            ComponentConstants.DEFAULT_STATS_REPORTING_INTERVAL_SEC,
+        )
+
+        model_filepath_key = java_mapping.RESERVED_KEYS[
+            ComponentConstants.INPUT_MODEL_TAG_NAME
+        ]
+        self._params[model_filepath_key] = ModelEnv(
+            self._params[model_filepath_key], self._ml_engine.standalone
+        ).model_filepath
+
+        single_uwsgi_worker = parameter.str2bool(
+            self._params.get(ComponentConstants.SINGLE_UWSGI_WORKER_KEY, False)
+        )
+
+        # Used for uwsgi UDP logging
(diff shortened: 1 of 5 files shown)


yakov-g:
So this is just copying files into this repo, right? DRUM will not use files yet?

yakov-g:
Could you please push into the branch again to trigger tests.

yakov-g:
> So this is just copying files into this repo, right? DRUM will not use files yet?

We should completely remove: mlpiper usage from `uwsgi_serving.py` and remove `component.json` , which is component configuration for mlpiper. And also create a boilerplate to instantiate comopnents without mlpiper.

yura-ant:
> So this is just copying files into this repo, right? DRUM will not use files yet?

There are two commits in this PR. The first one just copy file and not use them. The second one use them via modify import to those copied files. I expect all test should pass as logic of those files are not changed.

yura-ant:
> > So this is just copying files into this repo, right? DRUM will not use files yet?
> 
> We should completely remove: mlpiper usage from `uwsgi_serving.py` and remove `component.json` , which is component configuration for mlpiper. And also create a boilerplate to instantiate comopnents without mlpiper.

In scope of https://docs.google.com/document/d/1yq-T7-pmjW49RsvbslNIl3BF_GRbYLti2McVmZgd_To/edit#heading=h.8vln2nt42rgh I have no plans to remove mlpliper. I've added this as possible future work. I like idea to remove it but I afraid it could lead to much more changes as uwsgi_serving.py is not  just  class which use mlpiper. Also I copied classes where I plan to do changes to allow configure uWSGI threads. But those files as well have deps to  other mlpipers files which I not copied.  In scope of  https://datarobot.atlassian.net/browse/PBMP-6561 I would like to do minimum acceptable changes  to support configure threads

yura-ant:
> Could you please push into the branch again to trigger tests.
just bump version

yakov-g:
In mlpiper in this folder: https://github.com/datarobot/mlpiper/tree/master/mlpiper-py/mlpiper/components - we have components of the **_mlpiper pipeline_**, one of them is RESTfulComponent (restful_component.py )
In DRUM we have UwsgiComponent(RESTfulComponent) https://github.com/datarobot/datarobot-user-models/blob/master/custom_model_runner/datarobot_drum/resource/components/Python/uwsgi_component/uwsgi_serving.py - in just implements RESTfulComponent.

Back to mlpiper: 
In https://github.com/datarobot/mlpiper/tree/master/mlpiper-py/mlpiper/components/restful - we have whole orthogonal implementation of uwsgi/nginx not related to mlpiper at all (it uses some utils/exceptions, should be changed to DRUM exception, or even its own). (It just can be a separate project).

We should move all the https://github.com/datarobot/mlpiper/tree/master/mlpiper-py/mlpiper/components/restful somewhere into DRUM and keep it a separate  **uwsgi/nginx** implementation. For example to: https://github.com/datarobot/datarobot-user-models/tree/master/custom_model_runner/datarobot_drum/resource/components/restful.. Or even on higher levels.

As for the RESTfulComponent https://github.com/datarobot/mlpiper/blob/master/mlpiper-py/mlpiper/components/restful_component.py, we should move it next to UwsgiComponent(RESTfulComponent) https://github.com/datarobot/datarobot-user-models/blob/master/custom_model_runner/datarobot_drum/resource/components/Python/uwsgi_component/uwsgi_serving.py  (as you did). I would clean it and merge together.

In this way: we will use mlpiper's pipleline construction functionality and component's functionality(nginx) will be in DRUM

I don't think we should take some files from the standalone uwsgi/nginx implementation
https://github.com/datarobot/mlpiper/blob/master/mlpiper-py/mlpiper/components/restful_component.py
https://github.com/datarobot/mlpiper/blob/master/mlpiper-py/mlpiper/components/restful/uwsgi_broker.py
https://github.com/datarobot/mlpiper/blob/master/mlpiper-py/mlpiper/components/restful/uwsgi_ini_template.py
and copy them directly into  `uwsgi_component/uwsgi_serving.py`

In this case we use  mlpiper's pipleline construction functionality and in pararralel use some uswgi/nginx functionality. Which is not good imho.

@zohar-mizrahi ?

yakov-g:
Also, somehow tests don't like your PR, please rebase it on master.

yura-ant:
> In mlpiper in this folder: https://github.com/datarobot/mlpiper/tree/master/mlpiper-py/mlpiper/components - we have components of the **_mlpiper pipeline_**, one of them is RESTfulComponent (restful_component.py ) In DRUM we have UwsgiComponent(RESTfulComponent) https://github.com/datarobot/datarobot-user-models/blob/master/custom_model_runner/datarobot_drum/resource/components/Python/uwsgi_component/uwsgi_serving.py - it just implements RESTfulComponent.
> 
> Back to mlpiper: In https://github.com/datarobot/mlpiper/tree/master/mlpiper-py/mlpiper/components/restful - we have whole orthogonal to mlpiper implementation of uwsgi/nginx. (it uses some utils/exceptions, should be changed to DRUM exception, or even its own). (It just can be a separate project).
> 
> We should move all the https://github.com/datarobot/mlpiper/tree/master/mlpiper-py/mlpiper/components/restful somewhere into DRUM and keep it a separate **uwsgi/nginx** implementation. For example to: https://github.com/datarobot/datarobot-user-models/tree/master/custom_model_runner/datarobot_drum/resource/components/restful.. Or even to the higher levels.
> 
> As for the RESTfulComponent https://github.com/datarobot/mlpiper/blob/master/mlpiper-py/mlpiper/components/restful_component.py, we should move it next to UwsgiComponent(RESTfulComponent) https://github.com/datarobot/datarobot-user-models/blob/master/custom_model_runner/datarobot_drum/resource/components/Python/uwsgi_component/uwsgi_serving.py (as you did). I would clean it and merge together.
> 
> In this way: we will use mlpiper's pipleline construction functionality and component's functionality(nginx) will be in DRUM
> 
> I don't think we should take some files from the standalone uwsgi/nginx implementation https://github.com/datarobot/mlpiper/blob/master/mlpiper-py/mlpiper/components/restful_component.py https://github.com/datarobot/mlpiper/blob/master/mlpiper-py/mlpiper/components/restful/uwsgi_broker.py https://github.com/datarobot/mlpiper/blob/master/mlpiper-py/mlpiper/components/restful/uwsgi_ini_template.py and copy them directly into `uwsgi_component/` which defines `pipeline` logic, In this case we try use mlpiper's _**pipleline construction functionality**_ and in parallel use some uswgi/nginx functionality. Which is not good imho.
> 
> @zohar-mizrahi ?

UwsgiComponent already provides ability in DRUM to configure uwsgi/nginx configuration via mlpiper and we have no plans to change that functionality in scope of that story. We actually utilise that functionality to run DRUM with enabled threads to improve cpu utilization for proxy model use case. In our POC it was done via environment variable UWSGI_THREADS as that option was missed in [uwsgi_ini_template.py](https://github.com/datarobot/mlpiper/blob/master/mlpiper-py/mlpiper/components/restful/uwsgi_ini_template.py) to allow set it in transparent way.
As result  in scope of that story we want just have all uwsgi options in UWSGI.ini file and not use environment variable like UWSGI_THREADS.

We copy just files which plan to improve a bit with minimum changes to avoid any change to existing functionality. As mlpiper repo is prohibited to update. Other case I would do those changes in mlpiper repo.
I can move those 3 files to any location  if needed  please advise but I would like to avoid any refactoring of those files to keep them close to original.

To remove/refactor full (or big part) of mlpiper looks like as separate tech debt story which require proper investigation and competency which I don't have at the moment. 

What about to have a meeting if we need to clarify any other items?

yura-ant:
All tests pass expect formatting of mlpiper files. I would like to avoid formatting of those file via # pylint: disable-all  to keep them close to original files or I better format it?

eric-s-s (changes requested):
based off of these changes, what is the actual aim here?  I see that https://github.com/datarobot/datarobot-user-models/pull/1091/commits/5fa4d6dd25462230abdb56b5f289c500c88ad7e2  is pulling this in and copying it.  

is the plan to then modify these specific pieces to change the eventual functionality?  It looks like that's the plan.  

## requested change

No matter then end use, this still needs to also pull in the unit tests for these components.  Especially if you plan on changing them, we'll want those test in place.

custom_model_runner/datarobot_drum/resource/components/Python/uwsgi_component/restful_component.py
```diff
@@ -0,0 +1,351 @@
...
+                          "`pip install uwsgi` or `pip install mlpiper[uwsgi]`"
+                raise MLPiperException(err_msg)
+
+        super(RESTfulComponent, self).__init__(
+            engine if engine else RestModelServingEngine("uwsgi-context")
+        )
+        self._dry_run = False
+        self._wsgi_broker = None
+        self._nginx_broker = None
+        self._wid = None
```
eric-s-s, line 70: what is `wid` ? and `wuuid` ?

eric-s-s:
jira ticket: https://datarobot.atlassian.net/browse/CODEGEN-1918

project: https://datarobot.atlassian.net/browse/PBMP-6561
tech spec: https://docs.google.com/document/d/1yq-T7-pmjW49RsvbslNIl3BF_GRbYLti2McVmZgd_To/edit

yakov-g:
@yura-ant 
We can have a meeting.
Sorry, but I still can not agree with the approach.
1. By copying those files we just introduce more imports from mlpiper while our goal is to stop using it.
2. The fact that you don't want to format added files and _keep them them close to the original_ introduces even more (mental?) binding between the drum and mlpiper, one always have to look back at the same files in mlpiper. It doesn't seem viable.

Another approach I could suggest.
0. Very very likely there will be a situation **_oh we need to copy one more file to DRUM_**
So
1. You **_copy_** all the uwsgi/nginx logic into DRUM maybe assuming it is WIP project.
2. Duplicate in DRUM RestfulComponent and UwsgiComponent or merge them together (UwsgiComponent2 + RestfulComponent2 ). It will be new (but truly just duplicated into DRUM)  component.
3. Add additional env var (similar to PRODUCTION) which make server use new UwsgiComponent2.
4. You change/improve/test/etc. 
5. Then we just switch RestfulComponent to this RestfulComponent2. Anyhow, as long as I'm aware, no one uses production mode. Maybe only @oleksandr-saienko ?

yura-ant:
> based off of these changes, what is the actual aim here? I see that [5fa4d6d](https://github.com/datarobot/datarobot-user-models/commit/5fa4d6dd25462230abdb56b5f289c500c88ad7e2) is pulling this in and copying it.
> 
> is the plan to then modify these specific pieces to change the eventual functionality? It looks like that's the plan.
> 
> ## requested change
> No matter then end use, this still needs to also pull in the unit tests for these components. Especially if you plan on changing them, we'll want those test in place.

The aim is to allow us modify in separate PR [uwsgi_broker.py](https://github.com/datarobot/mlpiper/blob/master/mlpiper-py/mlpiper/components/restful/uwsgi_broker.py) just to pass new threads option to uwsig.ini file via modify _generate_ini_file method and one constant [uwsgi_ini_template.py](https://github.com/datarobot/mlpiper/blob/master/mlpiper-py/mlpiper/components/restful/uwsgi_ini_template.py) that will make configuration of threads more transparent and in one place as today we can only set UWSGI_THREADS as environment variable when start drum to achieve the same functionality.

I was looking for such unit tests in mlpiper but unfortunately they are missed. I think they are missed as RESTfulComponent is abstract class where in drum we implement it and actually test it. The second file UwsgiBroker is kind of configuration of uwsgi server. I will add such unit test as I need to add here threads option in separate PR. 

Since I didn't change any functional logic in drum just change how we can modify desired files from mlpiper. As it  is prohibited to modify mlpiper repo.  So this PR looks like refactoring and we can merge it without new tests.

custom_model_runner/datarobot_drum/resource/components/Python/uwsgi_component/restful_component.py
yura-ant: wid is uwsgi.worker_id() 
wuuid  is str(uuid.uuid4()) just uniq id

yura-ant:
> @yura-ant We can have a meeting. Sorry, but I still can not agree with the approach.
> 
> 1. By copying those files we just introduce more imports from mlpiper while our goal is to stop using it.
It is true only in term of explicit vs implicit dependency. At the end we still have the same imports


> 2. The fact that you don't want to format added files and _keep them them close to the original_ introduces even more (mental?) binding between the drum and mlpiper, one always have to look back at the same files in mlpiper. It doesn't seem viable.
> 
My main aim to keep it close to mlpiper to allow reviewers of PR spend less time to review changes. And yes it should simplify look back at same files in mlpiper for example to understood why we copy them and why they exist. Imagine  if mlpiper back to life then we can simply move configuration of thread option back to mlpiper and remove those files.

> Another approach I could suggest. 0. Very very likely there will be a situation **_oh we need to copy one more file to DRUM_** 
You right but every such situation is different user story with different resources and owners. 
For example somebody may have decide to not use mlpiper together with uwsgi at all and switch to own version of server which may run python programs or threads option will be deprecated at some point.
As Result 
Unfortunately with my limited experience with mlpiper and drum (this is my first task actually) I would avoid such big changes  as it could lead to mistakes and deadlines. 

> 
> 1. You **_copy_** all the uwsgi/nginx logic into DRUM, considering it new WIP logic.
> 2. Duplicate in DRUM RestfulComponent and UwsgiComponent or merge them together (UwsgiComponent2 + RestfulComponent2 ). It will be new (but truly just duplicated into DRUM)  component.
> 3. Add additional env var (similar to PRODUCTION) which make server use new UwsgiComponent2.
> 4. You change/improve/test/etc.
> 5. When it is stable. We just switch RestfulComponent to this RestfulComponent2. Anyhow, as long as I'm aware, no one uses production mode. Maybe only @oleksandr-saienko ?

eric-s-s:
@yura-ant based on our discussions, should this PR be closed?

PR #1097, opened 2024-08-01
Yakov/mlpiper restrul refactoring

yakov-g:


yura-ant:
That PR has just one unit test which I've added when implement ability to configure threads option in mlpiper. This is definitely not enough for such big changes 

As reviewer I should be able to read unit tests to understood how proposed new functionality works and just after that review how it was implemented. 

Some existing functional tests (which exactly are not clear from PR) in DRUM cover how mlpiper was integrated before and they not test those all those classes which you plan to add. Those test are good as regression test to prove that expected integration with mlpiper continue works as before. 

Basic rule is when you add a new function you should add unit test. 
Once we follow it we have a lot benefits

Also this PR introduce a lot of duplicated classes one from mlpiper another one form drum which could lead to issues if user  chose wrong import.  Ideally if we remove completely dependency to mlpiper or remove those classes from mlpiper.

PR #1101, opened 2024-08-15
[CODEGEN-1918] Support concurrent requests for Custom Models

yura-ant:

## Summary
Allow enable flask process via command line option --max_workers or define CUSTOM_MODEL_WORKERS  numeric runtime parameter
Option 3 from
https://docs.google.com/document/d/1yq-T7-pmjW49RsvbslNIl3BF_GRbYLti2McVmZgd_To/edit#heading=h.8vln2nt42rgh

TODO
I would like to add more test. But in general I think it is ready for review

### DIFF
--- a/tests/functional/test_inference_per_framework.py
+++ b/tests/functional/test_inference_per_framework.py
@@ -313,6 +313,7 @@ class TestInference:
         ],
     )
     @pytest.mark.parametrize("pass_args_as_env_vars", [False])
+    @pytest.mark.parametrize("max_workers", [None, 2])
     def test_custom_models_with_drum_prediction_server(
         self,
         resources,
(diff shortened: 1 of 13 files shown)


yura-ant:
jarvis please test failed

yura-ant:
jenkins run test_functional_by_framework-check_by_framework

tests/functional/test_inference_per_framework.py
```diff
@@ -399,6 +406,7 @@ def test_custom_models_with_drum_prediction_server(
         ],
     )
     @pytest.mark.parametrize("pass_args_as_env_vars", [True])
+    @pytest.mark.parametrize("max_workers", [1])
```
baekdahl, line 409: Why parametrize when there is only one param?

baekdahl (approved):
LGTM.

tests/functional/test_inference_per_framework.py
yura-ant: To keep it consistent with @pytest.mark.parametrize("pass_args_as_env_vars", [True]). As we reuse same function in different tests. I think the test is too much parameterise but it works

scottp-ml (approved):
LGTM

eric-s-s (approved):

yura-ant:
Hi @yakov-g,  Have you had a chance to look at this?

klichukb (approved):
Good start and no nginx/uwsgi stack!

Overall for the future I think we might still consider using gunicorn for example, as I think app.run(processes) is still [not suggested ](https://flask.palletsprojects.com/en/3.0.x/deploying/) as a production-grade option for serving multiple clients

PR #1108, opened 2024-08-29
[APP-2364] [APPLS-2364] Add new custom app env 3.12 with streamlit 1.36.0

kideh88:

## Summary

We are advancing from the Experimental stage to a Production state. At the same time we can remove the dr_components package from environment 

## Rationale

- Similar environment to 3.9 python with streamlit however dr_components and other unnecessary packages removed

### DIFF
--- a/public_dropin_environments/python312_streamlit/Dockerfile
+++ b/public_dropin_environments/python312_streamlit/Dockerfile
@@ -2,12 +2,6 @@ FROM python:3.12-slim
 
 WORKDIR /app
 
-RUN apt-get update && apt-get install -y \
-    build-essential \
-    curl \
-    software-properties-common \
-    git \
-    && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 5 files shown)


rvorobii (approved):

public_dropin_environments/python312_streamlit/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM python:3.12-slim
+
+WORKDIR /app
+
+RUN apt-get update && apt-get install -y \
+    build-essential \
+    curl \
+    software-properties-common \
+    git \
+    && rm -rf /var/lib/apt/lists/*
```
ben-cutler-datarobot, line 10: All of this can be removed

ben-cutler-datarobot (approved):
Looks good, but we really don't need to add extra stuff to our packages.

PR #1110, opened 2024-09-02
[CODEGEN-1946] Chat monitoring

baekdahl:

## Summary
* Move `chat()` method up to `BaseLanguagePredictor` from `PythonPredictor`. This is done because we want to support chat for other predictors such as `BaseOpenAiGpuPredictor` which will be implemented in CODEGEN-1942.
* Merge different kinds of mlops initialization to `BaseLanguagePredictor._init_mlops()`. 
* Always initialize mlops when predictor supports chat.

## Rationale
Chat API requests are bypassing the prediction servers which is the place where mlops monitoring is normally happening. This PR implements monitoring in DRUM for chat requests.


custom_model_runner/datarobot_drum/drum/language_predictors/python_predictor/python_predictor.py
```diff
@@ -37,22 +38,21 @@ def __init__(self):
...
 
     def mlpiper_configure(self, params):
+        target_type = TargetType(params.get("target_type"))
+        code_dir = params["__custom_model_path__"]
+
+        self._model_adapter = PythonModelAdapter(model_dir=code_dir, target_type=target_type)
+
+        sys.path.append(code_dir)
+        self._model_adapter.load_custom_hooks()
+
```
baekdahl, line 48: Intialize model adapter before calling `super()` so that we can check if model adapter has chat hook when `_supports_chat` is called from super.

custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
```diff
@@ -174,6 +197,94 @@ def _predict(self, **kwargs) -> RawPredictResponse:
...
+
+        latest_message = completion_create_params["messages"][-1]["content"]
+        features_df = pd.DataFrame([{"prompt": latest_message}])
+
+        predictions = [message_content]
+        try:
+            self._mlops.report_predictions_data(
+                features_df,
+                predictions,
+                association_ids=[str(uuid.uuid4())],
```
klichukb, line 252: Is this by design to have a random uuid?

custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
```diff
@@ -174,6 +197,94 @@ def _predict(self, **kwargs) -> RawPredictResponse:
...
+            response_iter = iter(response)
+            first_chunk = next(response_iter)
+
+            if getattr(first_chunk, "object", None) == "chat.completion.chunk":
+                # Return a new iterable where the peeked object is included in the beginning
+                return itertools.chain([first_chunk], response_iter)
+        except StopIteration:
+            return iter(())
+        except Exception:
+            pass
```
klichukb, line 281: How about we log this somehow? who knows what kind of bug may be hiding here.

custom_model_runner/datarobot_drum/drum/language_predictors/python_predictor/python_predictor.py
```diff
@@ -79,17 +86,13 @@ def _init_mlops(self, params):
...
 
-        self._mlops = (
-            MLOps()
-            .set_model_id(params["model_id"])
-            .set_deployment_id(params["deployment_id"])
-            .set_channel_config(monitor_settings)
-            .agent(
-                mlops_service_url=params["external_webserver_url"],
-                mlops_api_token=params["api_token"],
+        self._mlops.set_channel_config(monitor_settings)
```
akshoop, line 89: Do we need to worry about the following from the base class?
```
if self._supports_chat():
    self._mlops.set_channel_config("spooler_type=API")
```
https://github.com/datarobot/datarobot-user-models/pull/1110/files#diff-a69efcf5368c853d546210b873e714a134fe082c7130cf5e37571b0c3313ba4cR127-R128

custom_model_runner/datarobot_drum/resource/chat_helpers.py
```diff
@@ -0,0 +1,2 @@
+def is_streaming_response(response):
+    return getattr(response, "object", None) != "chat.completion"
```
akshoop, line 2: Make sure to have Copyright block at top of this new file

akshoop (approved):
Overall LGTM from Core Modeling

custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
baekdahl: Yes. I need to pass some unique association id. This id should be passed to moderations framework when their chat integration is implemented at some point

custom_model_runner/datarobot_drum/drum/language_predictors/python_predictor/python_predictor.py
baekdahl: `_configure_mlops()` is only called when `_supports_chat()` is `False` so they won't conflict with each other. Would readability be improved if I rename `_configure_mlops()` to _configure_mlops_non_chat()` or something like that?

akshoop: Renaming would help with readability, maybe calling it `_configure_mlops_for_non_chat()`

custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
baekdahl: Good point. I have changed this so that the new exception will be raised from the caught one which should preserve the context

custom_model_runner/datarobot_drum/drum/language_predictors/python_predictor/python_predictor.py
baekdahl: Done 👍

custom_model_runner/datarobot_drum/resource/chat_helpers.py
baekdahl: Done

klichukb (approved):

PR #1113, opened 2024-09-03
[RAPTOR-11378] update Py support to Py>=3.8, bump version to 1.12.0

yakov-g:

## Summary
Update Py support to Py>=3.9

* LLM lazy loading support in DRUM  needs to use a package that depends on Py>=3.8.
* Newer versions of DRUM has never been tested and likely are not used in the older Py versions.
* Py3.7 is already end-of-life, Py3.8 is end-of-life on Oct 31, 2024.

Please share your opinion.

### DIFF
--- a/custom_model_runner/CHANGELOG.md
+++ b/custom_model_runner/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### [1.12.0] - 2024-09-03
 ##### Changed
-- Update support to Py>=3.8
+- Update support to Py>=3.9
 
 #### [1.11.6] - 2024-08-30
 ##### Changed
--- a/custom_model_runner/CHANGELOG.md
+++ b/custom_model_runner/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### [1.12.0] - 2024-09-03
 ##### Changed
-- Update support to Py>=3.9
+- Update support to Py>=3.8
 
 #### [1.11.6] - 2024-08-30
 ##### Changed
--- a/custom_model_runner/CHANGELOG.md
+++ b/custom_model_runner/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+#### [1.12.0] - 2024-09-03
+##### Changed
+- Update support to Py>=3.8
+
+#### [1.11.6] - 2024-08-30
+##### Changed
+- Don't cache vLLM dirs in DRUM
+
 #### [1.11.5] - 2024-06-18
 ##### Changed
 - Set numpy version to <2.0.0
--- a/custom_model_runner/setup.py
+++ b/custom_model_runner/setup.py
@@ -47,6 +47,7 @@ setup(
     license="Apache License, Version 2.0",
     classifiers=[
         "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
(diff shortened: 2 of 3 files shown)


oleksandr-saienko (approved):

custom_model_runner/setup.py
```diff
@@ -71,5 +67,5 @@
     scripts=["bin/drum"],
     install_requires=requirements,
     extras_require=extras_require,
-    python_requires=">=3.4,<3.12",
+    python_requires=">=3.8,<3.12",
```
elatt, line 70: Why do we have the strict max? Do we know for a fact that DRUM doesn't work with 3.12?

yakov-g: If I remember correct, at the time it was added, we did not have strict versions pinning for some dependencies (maybe we still do). So in this case, some dependency can pull the latest version and break things.

elatt: I think I've seen Poetry (incorrectly) complain about no upper bound pin on a Python Version (but if that's really true, I would suggest something like this)
```suggestion
    python_requires=">=3.8,<4.0",
```

Because PY 3.12 has been out now since end of 2023 so I don't think we should restrict folks from using it unless if we know for a fact DRUM doesn't support it.

elatt (approved):
LGTM. Consider loosening upper bound pin as Python versions tend to be a lot more backward compatible these days so having an upper pin (unless if we **know** something is actually broken) are counter productive.

custom_model_runner/setup.py
eric-s-s: so, @yakov-g , do you mean that because we don't pin versions of other libraries, that when someone uses 3.12, it can break things?  do we have tests for this anywhere?

yakov-g: For example we don't pin `requests` (put here any name). So if there is a requests version for 3.12, now it is not installed as DRUM doesn't support 3.12.
Once DRUM starts supporting 3.12 that `requests` may break things. 
Does it make sense? I don't remember the exact reason, why we limit Py<3.12, but likely the one i mentioned.

We also pin pandas pandas>=1.5.0,<=2.0.3, while 3.12 support starts with pandas==2.2.0...

And, "no" we don't test for different Py versions.

So need to resolve this... with `poetry` ;)

eric-s-s (approved):

custom_model_runner/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+#### [1.12.0] - 2024-09-03
+##### Changed
+- Update support to Py>=3.8
```
DrewParallel, line 9: I'd say >= 3.9

yakov-g: Doesn't work. 
Our R env is Py3.8.
Requires to invest a bit more time to update.

custom_model_runner/setup.py
elatt: I'm not following this logic 🤔 . The version of the `requests` package (or any package for that matter) that is installed along with `datarobot-drum` shouldn't matter with what our max `python_version` pin is.  The value of the python version pin should simply reflect the usage of Python stdlib functions and other syntax that requires or disallows certain versions of Python.

All the other dependencies or transitive deps that need to be specified when actually building a base execution env should not be the responsibility of the DRUM library to specify (it would actually be incorrect to have the drum package artificially restrict the version of say the `requests` package when it works with all known versions of the library). It should be the responsibility of building the envs to specific the correct mix of libraries and versions.

yakov-g: I agree with you, and this was my logic, that DRUM shouldn't pin the version as it usually uses very basic functionality. 

But failures already happened, for example with flask and werkzeug, just during some general import when a new version of those unpinned packages was released.

So probably this not about the upper Py version, but only about pinning deps.

Anyhow, need to add tests per py version. Probably in harness it could be a pipeline with several stages, per Py version,  that runs unit tests and a couple of integration/funcional tests. I mean DRUM integration/functional, not E2E with DR.

klichukb (approved):

PR #1118, opened 2024-09-12
[AGENT-6033 AGENT-6065] Bump version of vLLM env to v0.6.2

elatt:

## Summary
- Bump vLLM to v0.6.2
- Bump DRUM v1.13.0b1
- Set vLLM cache dir override env vars in Dockerfile
- Stop install `autoawq`

## Rationale
The new vLLM supports more models and also has performance improvements. The new version of drum allows us to be more nible to vLLM change by allowing us to have more changes in the `Dockerfile` rather than needing to rely on DRUM code changes to support new vLLM versions. We have stopped installing `autoawq` because it wasn't compatible with the newest version of vLLM and also isn't needed for all cases.

### DIFF
--- a/public_dropin_gpu_environments/vllm/Dockerfile
+++ b/public_dropin_gpu_environments/vllm/Dockerfile
@@ -7,17 +7,16 @@ RUN apt-get update && apt-get install -y \
 
 # Don't send any telemetry data (vLLM or HuggingFace libraries)
 ENV DO_NOT_TRACK=1
+ENV PIP_NO_CACHE_DIR=1
+ENV CODE_DIR=/opt/code ADDRESS=0.0.0.0:8080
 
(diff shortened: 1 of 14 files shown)


alexromanyuk (approved):

dulcardo (approved):

rvorobii (approved):

public_dropin_gpu_environments/vllm/Dockerfile
```diff
@@ -1,24 +1,22 @@
-FROM vllm/vllm-openai:v0.5.5
+FROM vllm/vllm-openai:v0.6.2
 USER root
 RUN apt-get update && apt-get install -y \
-    python3-pip \
-    python3-venv \
+    python3.9 \
```
elatt, line 4: vLLM is using PY 3.12 but DRUM doesn't support that version yet.

PR #1121, opened 2024-09-19
[CODEGEN-1949] Use custom model workers with and without runtime_params_file

yura-ant:

## Summary

Try to use CUSTOM_MODEL_WOKERS runtime parameter even if it was set without using runtime_params_file.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/main.py
+++ b/custom_model_runner/datarobot_drum/drum/main.py
@@ -92,11 +92,11 @@ def main():
             try:
                 loader = RuntimeParametersLoader(options.runtime_params_file, options.code_dir)
                 loader.setup_environment_variables()
-                if RuntimeParameters.has("CUSTOM_MODEL_WORKERS"):
-                    options.max_workers = RuntimeParameters.get("CUSTOM_MODEL_WORKERS")
             except RuntimeParameterException as exc:
                 print(str(exc))
                 exit(255)
+        if RuntimeParameters.has("CUSTOM_MODEL_WORKERS"):
+            options.max_workers = RuntimeParameters.get("CUSTOM_MODEL_WORKERS")
         runtime.options = options
 
         # mlpiper restful_component relies on SIGINT to shutdown nginx and uwsgi,
--- a/tests/unit/datarobot_drum/drum/test_main.py
+++ b/tests/unit/datarobot_drum/drum/test_main.py
@@ -23,7 +23,6 @@ def test_custom_model_workers(
     options = argparse.Namespace()
     options.max_workers = 0
     options.code_dir = "dir"
-    options.runtime_params_file = "runtime_params_file"
 
     arg_parser = Mock()
     arg_parser.parse_args.return_value = options


custom_model_runner/datarobot_drum/drum/main.py
```diff
@@ -92,11 +92,11 @@ def signal_handler(sig, frame):
             try:
                 loader = RuntimeParametersLoader(options.runtime_params_file, options.code_dir)
                 loader.setup_environment_variables()
-                if RuntimeParameters.has("CUSTOM_MODEL_WORKERS"):
-                    options.max_workers = RuntimeParameters.get("CUSTOM_MODEL_WORKERS")
             except RuntimeParameterException as exc:
                 print(str(exc))
                 exit(255)
+        if RuntimeParameters.has("CUSTOM_MODEL_WORKERS"):
```
eric-s-s, line 98: this seems _very_ iffy.  This tends to indicate that the workers is a special case and isn't actually a runtime parameter at all.  that kind of makes sense.  These runtime parameters are user-injected env vars and user-inject secret env vars.  they aren't about configuration of how the server is run and so this looks very much like the number of workers here is not really a runtime param, but a regular param that gets passed in to the `drum server` command using the args parser (these can also be passed in as env vars that get picked up by that args parser).  If you need some help understanding how the args parsing works, let me know.

yura-ant: Just want to make it consistent for end user. As user define runtime parameters CUSTOM_MODEL_WORKERS via UI in DataRobot  it can be not automatically used by drum  as it could be started without runtime_params_file option. There is a way to manually get CUSTOM_MODEL_WORKERS value from env vars and  use it when start drum with   --max-workers  option  but it looks to me like workaround.

eric-s-s: can you provide links to the DR code, please?  I'm missing the context here.

yura-ant: I've discovered it on staging  when try to define CUSTOM_MODEL_WORKERS  runtime parameter and my custom start_server.sh run  drum without  runtime_params_file option then CUSTOM_MODEL_WORKERS is not used. As I added it only for use case when user use runtime_params_file option when start drum

yura-ant: looks like it is here https://github.com/datarobot/DataRobot/blob/19d867ee7e6af21295b5d68d98ba0bce58e7a732/raptor_lake/custom_model/services/executable_service.py#L992

eric-s-s: but why is the custom model workers using launch config _at all_?  it's not a launch config.  it's a setting.

yura-ant: since options.max_workers was already in place to  controls a number of processes in the drum server. And at the same time we had a requirement that we would like to introduce runtime parameter CUSTOM_MODEL_WORKERS which could be used to controls a number of processes  and Users can override the runtime parameter at deployment time.

rvorobii:
@yakov-g please review this PR

custom_model_runner/datarobot_drum/drum/main.py
yakov-g: So as I understand, RuntimeParameters are intended to be only used within custom.py, while here we use it directly in DRUM.
If we want to pass some values to DRUM, the right way to do this, probably would be to add an api  drum.set_number_of_workers() and call it in the `init()` hook. Or smth like that. (If I remember correct, server is started after init() and load_model() hooks)

Does it sound about right?
@eric-s-s ?

tests/unit/datarobot_drum/drum/test_main.py
```diff
@@ -23,7 +23,6 @@ def test_custom_model_workers(
     options = argparse.Namespace()
     options.max_workers = 0
```
eric-s-s, line 24: this should be the string "0", correct?  if you're reading from env vars, then it won't be cast to an int.

custom_model_runner/datarobot_drum/drum/main.py
eric-s-s: This should probably be part of the resource allocations that are set as part of a custom model version like this 
image

and not set as runtime parameters.

eric-s-s: @yura-ant this is where we define env vars that are `drum server` settings https://github.com/datarobot/DataRobot/blob/19d867ee7e6af21295b5d68d98ba0bce58e7a732/raptor_lake/custom_model/services/executable_service.py#L919   .  The drum server is read from the env var like this test shows: https://github.com/datarobot/datarobot-user-models/pull/1126

yakov-g: > I am looking for the simplest option for end user to avoid update multiple files before start using CUSTOM_MODEL_WORKERS runtime parameters.

If you say that there is a requirement to make CUSTOM_MODEL_WORKERS user defined, we should brainstorm together on how to do this. Maybe we could define system RuntimeParams and reserve some names, and then properly set env vars in the: `setup_environment_variables`
I suggest to invite @elatt and @ahamino to the discussion

elatt: We don't really have a scheme for "system" defined runtime params. We had some brainstorming sessions in the past on adding "execution environment defined" runtime params that a custom model would inherit based on what execution environment was used. This is not a trivial thing to add though and the work has not been prioritized.

I don't know if I totally agree that this belongs with the replica and mem settings because those are k8s scheduling related stuff so the DR platform has control of them. This option is really specific to DRUM so it does seem more like it should be associated with the environment (I.e. envs with older versions of drum don't support it).

However in lieu of actually doing the big work to add exec env defined parameters, we have opted to just allow them to be defined lazily (only if a user wants to actually override them). Since this is becoming a common pattern (GPU envs, GenAI envs, and now this) we should improve the interface in drum for dealing with undefined/optional params.

baekdahl: Just want to add that implementing this using a runtime parameter was in [the spec](https://docs.google.com/document/d/1yq-T7-pmjW49RsvbslNIl3BF_GRbYLti2McVmZgd_To/edit#heading=h.dajgx8borbez), so it would be been great to have this conversation earlier.

I agree, that using runtime parameters for this kind of configuration is not optimal but it fits the requirements without having to do a lot of extra work.

yura-ant: > @yura-ant this is where we define env vars that are `drum server` settings https://github.com/datarobot/DataRobot/blob/19d867ee7e6af21295b5d68d98ba0bce58e7a732/raptor_lake/custom_model/services/executable_service.py#L919 . The drum server is read from the env var like this test shows: #1126

I can try to use MAX_WORKER  in  executable_service.py if CUSTOM_MODEL_WORKERS runtime parameter was defined. But I still think that current PR is better option today as it will allow use  CUSTOM_MODEL_WORKERS  runtime parameter without DataRobot app in case when it was defined without runtime_params_file. Anyway we need to sync CUSTOM_MODEL_WORKERS to MAX_WORKER  if  runtime_params_file option was used as it is currently  implemented without this PR.

tests/unit/datarobot_drum/drum/test_main.py
elatt: Runtime params are type aware and support numeric types and casts them (to floats actually I think).

elatt (approved):
Now that I have the context, this LGTM

1. It appears that someone has implemented a way to deal with _optional_ runtime params in DRUM (e.g. ` RuntimeParameters.has()`)
2. It is correct to move your code out of the `runtime_params_file` block. That path is only for local custom model development. When a model actually runs in the platform, params are injected as env vars, not via a file (currently) so you shouldn't depend on its presence.

eric-s-s (approved):
I appreciate the conversation.  For what we have now, this makes sense as the best way forward.

custom_model_runner/datarobot_drum/drum/main.py
```diff
@@ -92,11 +92,11 @@ def signal_handler(sig, frame):
...
+            options.max_workers = RuntimeParameters.get("CUSTOM_MODEL_WORKERS")
```
elatt, line 99: I presume this is meant to be a `numeric` type -- if so, shouldn't this be cast as an `int()` because in the platform all numerics are floats

yura-ant: It works even if user define float like "2.1"  https://staging.datarobot.com/registry/custom-model-workshop/66e9709bb28fc936864e52e2/test.  But ideally I am thinking about int runtime parameters to not allow user define float value at all.  I can update to options.max_workers = init(RuntimeParameters.get("CUSTOM_MODEL_WORKERS")) if it more safe/clear for us. Does it cause any issues today ?

elatt: I just tested in my own model and it looks like it works as is. I was just concerned because I ran into a case recently where something didn't support "1.0" so I needed to cast it to an int first 🤷‍♂️. Anyway, I agree what you need more is a constraint on the UI side to say only allow integer values.

elatt: We never added ints to runtime params because we are trying to keep things simple and there haven't been many cases we found where we really care about an int vs a float.

PR #1125, opened 2024-09-23
[CODEGEN-1946] Chat monitoring improvements

baekdahl:

## Summary
* Don't skip drift or accuracy tracking when reporting chat prediction.
* Don't validate streaming chat response using `object` property because this property is not set for first chunk returned by OpenAI.
* Get prompt column name from deployment instead of using fixed name `prompt`.
* Use `MLOPS_DEPLOYMENT_ID` and `MLOPS_MODEL_ID` environment variables to populate these DRUM settings instead the previous names without `MLOPS_` prefix. The names with `MLOPS_` prefix are used when DRUM is running inside DataRobot so it makes sense to align the names.

## Rationale
These fixes are required for chat monitoring to work properly.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/args_parser.py
+++ b/custom_model_runner/datarobot_drum/drum/args_parser.py
@@ -562,14 +562,16 @@ class CMRunnerArgsRegistry(object):
 
             parser.add_argument(
                 ArgumentsOptions.DEPLOYMENT_ID,
-                default=os.environ.get("MLOPS_DEPLOYMENT_ID", None),
-                help="Deployment id to use for monitoring model predictions (env: MLOPS_DEPLOYMENT_ID)",
+                default=os.environ.get(
+                    "DEPLOYMENT_ID", os.environ.get("MLOPS_DEPLOYMENT_ID", None)
+                ),
+                help="Deployment id to use for monitoring model predictions (env: DEPLOYMENT_ID)",
             )
 
             parser.add_argument(
                 ArgumentsOptions.MODEL_ID,
-                default=os.environ.get("MLOPS_MODEL_ID", None),
-                help="MLOps model id to use for monitoring predictions (env: MLOPS_MODEL_ID)",
+                default=os.environ.get("MODEL_ID", os.environ.get("MLOPS_MODEL_ID", None)),
+                help="MLOps model id to use for monitoring predictions (env: MODEL_ID)",
             )
 
             parser.add_argument(
--- a/custom_model_runner/datarobot_drum/drum/args_parser.py
+++ b/custom_model_runner/datarobot_drum/drum/args_parser.py
@@ -562,14 +562,14 @@ class CMRunnerArgsRegistry(object):
 
             parser.add_argument(
                 ArgumentsOptions.DEPLOYMENT_ID,
-                default=os.environ.get("DEPLOYMENT_ID", None),
-                help="Deployment id to use for monitoring model predictions (env: DEPLOYMENT_ID)",
+                default=os.environ.get("MLOPS_DEPLOYMENT_ID", None),
+                help="Deployment id to use for monitoring model predictions (env: MLOPS_DEPLOYMENT_ID)",
             )
 
             parser.add_argument(
                 ArgumentsOptions.MODEL_ID,
-                default=os.environ.get("MODEL_ID", None),
-                help="MLOps model id to use for monitoring predictions (env: MODEL_ID)",
+                default=os.environ.get("MLOPS_MODEL_ID", None),
+                help="MLOps model id to use for monitoring predictions (env: MLOPS_MODEL_ID)",
             )
 
             parser.add_argument(
--- a/custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
+++ b/custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
@@ -270,13 +270,12 @@ class BaseLanguagePredictor(DrumClassLabelAdapter, ABC):
             response_iter = iter(response)
             first_chunk = next(response_iter)
 
-            if getattr(first_chunk, "object", None) == "chat.completion.chunk":
+            if type(first_chunk).__name__ == "ChatCompletionChunk":
                 # Return a new iterable where the peeked object is included in the beginning
                 return itertools.chain([first_chunk], response_iter)
             else:
                 raise Exception(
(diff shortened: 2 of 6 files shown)


custom_model_runner/datarobot_drum/drum/args_parser.py
```diff
@@ -562,14 +562,14 @@ def _reg_args_monitoring(*parsers):
 
             parser.add_argument(
                 ArgumentsOptions.DEPLOYMENT_ID,
-                default=os.environ.get("DEPLOYMENT_ID", None),
-                help="Deployment id to use for monitoring model predictions (env: DEPLOYMENT_ID)",
+                default=os.environ.get("MLOPS_DEPLOYMENT_ID", None),
```
akshoop, line 565: Any backwards compatibility concerns with this?

custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
```diff
@@ -125,12 +130,31 @@ def _init_mlops(self):
...
 
+    def _configure_mlops_for_chat(self):
+        self._mlops.set_channel_config("spooler_type=API")
+
+        self._prompt_column_name = self._get_prompt_column_name()
+        logger.debug("Prompt column name: %s", self._prompt_column_name)
+
+    def _get_prompt_column_name(self):
+        if not self._params.get("deployment_id", None):
+            logger.error("No deployment id found while configuring mlops for chat.")
```
akshoop, line 147: Maybe also mention falling back to using default prompt column? So it could say something like:
```suggestion
            logger.error("No deployment ID found while configuring mlops for chat. Fallback to default prompt column name ('promptText')")
```

tests/unit/datarobot_drum/drum/language_predictors/test_base_language_predictor.py
```diff
@@ -113,11 +113,52 @@ def chat_hook(completion_request):
...
-            skip_drift_tracking=True,
-            skip_accuracy_tracking=True,
         )
         # Compare features dataframe separately as this doesn't play nice with assert_called
-        assert mock_mlops.report_predictions_data.call_args.args[0]["prompt"].values[0] == "Hello!"
+        assert (
+            mock_mlops.report_predictions_data.call_args.args[0]["promptText"].values[0] == "Hello!"
+        )
+
+    def test_prompt_column_name(self, chat_python_model_adapter, mock_mlops):
```
akshoop, line 122: Can we have a test for default prompt column name?

akshoop (approved):
Changes LGTM from Core Modeling with some questions/minor comments, CC @eric-s-s

yakov-g (changes requested):
Please improve DEPLOYMENT_ID / MODEL_ID

custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
```diff
@@ -125,12 +130,31 @@ def _init_mlops(self):
...
 
         if self._supports_chat():
-            self._mlops.set_channel_config("spooler_type=API")
+            self._configure_mlops_for_chat()
         else:
             self._configure_mlops_for_non_chat()
 
         self._mlops.init()
 
+    def _configure_mlops_for_chat(self):
```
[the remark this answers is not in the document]
baekdahl: I don't see chat as a separate predictor type. It's a method like predict and transform that can be used on all predictor types (if they support it).

Would it make sense to do something like a  `ChatMixin` and inherit that in `PythonPredictor` and `BaseOpenAiGpuPredictor`(I'm also adding chat support there) ?
In my opinion, the current approach is better.

custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
```diff
@@ -272,13 +294,12 @@ def _validate_chat_response(response):
             response_iter = iter(response)
             first_chunk = next(response_iter)
 
-            if getattr(first_chunk, "object", None) == "chat.completion.chunk":
+            if type(first_chunk).__name__ == "ChatCompletionChunk":
```
yakov-g, line 297: Why not `isinstance` ? 
In order not to import ChatCompletionChunk?

custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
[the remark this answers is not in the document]
yakov-g: I would probably create GenAiPredictor (with chat hook), to separate from BaseLanguagePredictor, and create a separate problem family, like we have Structured/Unstructured. 
And then on, on top of it, have OpenAi related predictors. Then Open UI + GPU.
But maybe I'm completely wrong wit this appoach.

For now  I'll leave it to you, but we can have a brainstorming session. Maybe with @alexromanyuk. And maybe get some input from Abdo on how to group the tasks/problems we are solving.

custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
baekdahl: Yes. We would need to depend on openai python package to have that.

custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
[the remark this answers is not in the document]
baekdahl: Yes, let's meet and talk about this, but any refactoring makes sense to put in a follow-up PR.

tests/unit/datarobot_drum/drum/language_predictors/test_base_language_predictor.py
baekdahl: [test_chat_with_mlops](https://github.com/datarobot/datarobot-user-models/blob/bacfedd9ac76b189df43b8e2f55ce74a7a3d1078/tests/unit/datarobot_drum/drum/language_predictors/test_base_language_predictor.py#L84) is testing this implicitly

custom_model_runner/datarobot_drum/drum/args_parser.py
baekdahl: Done

custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
baekdahl: Done

yakov-g (approved):

PR #1132, opened 2024-10-03
[AGENT-6315] Develop POC of latest NIM predictor support

elatt:

## Summary
Add a new GPU predictor that supports the new NIM images (1.x). Convert the existing NIM dropin environment to use the new predictor class. One caveat to note with the way the new NIMs work, is that
- They have the capability to download the engine files from HuggingFace or NGC itself (no need for `custom.py` code).
- The docker envs distributed by Nvidia are tied to a specific LLM. I haven't dug in enough to see exactly what differs between each docker env.

## Rationale
The prior code was buit to support the pre-release version of NIM and the 1.0 release changed the way DRUM needs to interact with NIM.


public_dropin_gpu_environments/nim_llm/dr_requirements.txt
```diff
@@ -43,7 +46,7 @@ datarobot==3.3.1
     # via
     #   -r dr_requirements.in
     #   datarobot-drum
-datarobot-drum==1.11.5
+datarobot-drum @ git+https://github.com/datarobot/datarobot-user-models.git@ec4587c0a09024deac1545f2460411be90a5794a#subdirectory=custom_model_runner&egg=datarobot-drum
```
elatt, line 49: Will remove this once this PR merges to master and I can release the version of DRUM

public_dropin_gpu_environments/nim_llm/Dockerfile
```diff
@@ -1,25 +1,24 @@
...
-
+FROM nvcr.io/nim/meta/llama-3.1-8b-instruct:1.2.2
 
 USER root
 RUN apt-get update && apt-get install -y \
     python3-pip \
     python3-venv \
-    zstd \
+    python3-wheel \
+    git \
```
elatt, line 8: Remove this after we release DRUM

rvorobii:
@yakov-g please review this one

tests/functional/test_inference_per_framework.py
```diff
@@ -1171,7 +1161,7 @@ def test_nemo_predictor(
             assert "predictions" in response_data, response_data
             assert len(response_data["predictions"]) == 1
             assert (
-                "Why don't scientists trust atoms?" in response_data["predictions"][0]
+                "What do you call a fake noodle?" in response_data["predictions"][0]
```
alexromanyuk, line 1164: It learned new jokes! :+1:

public_dropin_gpu_environments/nim_llm/Dockerfile
```diff
@@ -1,25 +1,23 @@
-FROM nvcr.io/ohlfw0olaadg/ea-participants/nim_llm:24.02.1
-
+FROM nvcr.io/nim/meta/llama-3.1-8b-instruct:1.2.2
```
alexromanyuk, line 1: Are we going to create more dropin environments per each model?

alexromanyuk (approved):
:+1: looks good to me

public_dropin_gpu_environments/nim_llm/Dockerfile
elatt: Not without much improved infrastructure (because they also release new versions of the envs when there are CVEs, etc) so we will be spending all our time maintaining these envs with the current system. This solution should be considered a POC used for demoing and driving the discussion to build an official integration with NIM. cc @ahamino

scottp-ml (approved):

PR #1134, opened 2024-10-04
[CODEGEN-1957] Make chat work without monitoring

baekdahl:

## Summary
If deployment id is not set, don't initialize mlops.

## Rationale
Running without monitoring is useful e.g. when testing or running locally.

### DIFF
--- a/tests/unit/datarobot_drum/drum/language_predictors/test_base_language_predictor.py
+++ b/tests/unit/datarobot_drum/drum/language_predictors/test_base_language_predictor.py
@@ -28,14 +28,7 @@ class TestLanguagePredictor(BaseLanguagePredictor):
 
 class TestChat:
     @pytest.fixture
-    def mock_mlops(self):
-        with patch("datarobot_drum.drum.language_predictors.base_language_predictor.MLOps") as mock:
-            mlops_instance = Mock()
-            mock.return_value = mlops_instance
-            yield mlops_instance
-
-    @pytest.fixture
-    def language_predictor(self, chat_python_model_adapter, mock_mlops):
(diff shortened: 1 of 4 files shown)


YuriyGuts:
Could you please bump the package version as well with this change? We'd need to install a new DRUM in Buzok's envs for the change to take effect.

tests/unit/datarobot_drum/drum/language_predictors/test_base_language_predictor.py
```diff
@@ -27,25 +27,72 @@ def _chat(self, completion_create_params):
...
                 "target_type": TargetType.TEXT_GENERATION,
                 "__custom_model_path__": "/non-existing-path-to-avoid-loading-unwanted-artifacts",
+                "deployment_id": "deployment_id",
             }
         )
 
         yield predictor
 
+    @pytest.mark.parametrize("stream", [False, True])
+    def test_chat_without_mlops(self, language_predictor, stream):
```
akshoop, line 68: I think it would be better for readability to split out the parametrized test. So you would have two tests (names are suggestions):
```
test_chat_yes_stream_without_mlops
test_chat_no_stream_without_mlops
```

akshoop (approved):
LGTM Core Modeling

tests/unit/datarobot_drum/drum/language_predictors/test_base_language_predictor.py
baekdahl: Thanks for the suggestion but I'm using the `stream` parameter in many tests in this file so I would prefer to keep it like this for consistency.

klichukb (approved):
sorry for the delay

PR #1135, opened 2024-10-11
Add watchdog to GPU predictors

elatt:

## Summary
- Bump the version of DRUM to latest version in the vLLM and NIM environments.
- Add watchdog to restart OpenAI server if it crashes randomly
- Refactor some code into base class
- Add support for `engine_config.json` for NIM server to support full set of NIM configuration
- Removed `trust_remote_code` runtime param

## Rationale
Changes to the base OpenAI predictor code was made recently. Also, this makes the NIM and vLLM envs more robust by adding a watchdog thread that will restart random crashes of the underlying service. Without this, the server may crash/OOM but DRUM may stay up so the user would have no recourse other than to deactivate their Deployment to restart the server.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/gpu_predictors/base.py
+++ b/custom_model_runner/datarobot_drum/drum/gpu_predictors/base.py
@@ -11,8 +11,10 @@ import logging
 import os
 import signal
 import sys
+import time
 import typing
 from pathlib import Path
+from threading import Event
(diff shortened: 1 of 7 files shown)


kijimoshib (approved):
LGTM

DrewParallel (approved):
I'd say this does a bit more than just update to latest version of DRUM

PR #1140, opened 2024-10-16
Test vLLM env in multi-process mode

elatt:

## Summary
- Pass runtime param in test to turning on multiple HTTP workers
- Hack uwsgi

## Rationale
The GPU predictors would benefit from more workers because the underlying vLLM server can support more concurrent requests but DRUM is blocking this.

### DIFF
--- a/public_dropin_gpu_environments/vllm/dr_requirements.txt
+++ b/public_dropin_gpu_environments/vllm/dr_requirements.txt
@@ -35,7 +35,7 @@ datarobot==3.3.2
     # via
     #   -r dr_requirements.in
     #   datarobot-drum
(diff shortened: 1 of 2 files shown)


public_dropin_gpu_environments/vllm/dr_requirements.txt
```diff
@@ -35,7 +35,7 @@ datarobot==3.3.2
     # via
     #   -r dr_requirements.in
     #   datarobot-drum
-datarobot-drum==1.13.0b2
+datarobot_drum-1.13.0b2-py3-none-any.whl
```
elatt, line 38:
```suggestion
datarobot-drum==1.13.0b2
```

PR #1154, opened 2024-10-27
[RAPTOR-11322] Add lazy loading processor to download the associated lazy loading remote files to the local volume

zohar-mizrahi:
## Summary
In the context of the lazy loading custom model support, it is required to process the related environment variable, if exists, and download the required files to the root folder of the executed custom model.

## Changes
* Add a new method to the `BaseLanguagePredictor` to handle lazy loading files and call it from the `PythonPredictor` within the configuration phase.
* Add a new handler `LazyLoadingHandler` to deal with processing the lazy loading structured data JSON from an environment variable and use `datarobot_storage` to download the lazy loading files in parallel.
* Small improvement to the `credential_type` attribute definition in `S3Credentials`.
* Minor change to `.gitconfig`.
* Unit, integration and functional tests - note that the functional test requires special environment variables in order not to be skipped. It assumes an accessible MinIO service that is set with specific files. (I run it locally in my environment).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,7 +86,7 @@ celerybeat-schedule
 *.sage.py
 
 # Environments
-.env
+.env*
 .venv
 env/
 venv/
--- a/custom_model_runner/datarobot_drum/drum/lazy_loading/exceptions.py
+++ b/custom_model_runner/datarobot_drum/drum/lazy_loading/exceptions.py
@@ -1,0 +1,2 @@
+class InvalidValue(Exception):
+    pass
--- a/custom_model_runner/datarobot_drum/drum/lazy_loading/exceptions.py
+++ b/custom_model_runner/datarobot_drum/drum/lazy_loading/exceptions.py
@@ -1,2 +1,0 @@
-class InvalidValue(Exception):
-    pass
--- a/custom_model_runner/requirements.txt
+++ b/custom_model_runner/requirements.txt
@@ -26,4 +26,4 @@ termcolor
 packaging
 markupsafe<=2.1.3
 pydantic==2.9.2
-datarobot-storage
+datarobot-storage>=0.0.0,<1.0.0
--- a/custom_model_runner/requirements.txt
+++ b/custom_model_runner/requirements.txt
@@ -26,3 +26,4 @@ termcolor
 packaging
 markupsafe<=2.1.3
 pydantic==2.9.2
+datarobot-storage
--- a/tests/functional/test_lazy_loading.py
+++ b/tests/functional/test_lazy_loading.py
@@ -1,0 +1,99 @@
+import json
+import os
+import time
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+from datarobot_storage import get_storage
+from datarobot_storage.amazon import S3Storage
+from datarobot_storage.enums import FileStorageBackend
+
+from datarobot_drum.drum.lazy_loading.constants import BackendType
+from datarobot_drum.drum.lazy_loading.constants import EnumEncoder
+from datarobot_drum.drum.lazy_loading.constants import LazyLoadingEnvVars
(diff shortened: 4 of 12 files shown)


rvorobii (approved):

.gitignore
```diff
@@ -86,7 +86,7 @@ celerybeat-schedule
 *.sage.py
 
 # Environments
-.env
+.env*
```
akshoop, line 89: Why this change?

custom_model_runner/datarobot_drum/drum/lazy_loading/exceptions.py
```diff
@@ -0,0 +1,2 @@
+class InvalidValue(Exception):
```
akshoop, line 1: This seems unused

custom_model_runner/requirements.txt
```diff
@@ -26,3 +26,4 @@ termcolor
 packaging
 markupsafe<=2.1.3
 pydantic==2.9.2
+datarobot-storage
```
akshoop, line 29: I think this is only available internally, are we sure we can add this dependency?

tests/functional/test_lazy_loading.py
```diff
@@ -0,0 +1,99 @@
...
+    def skip_minio_test_if_missing_env_vars(self):
+        if "AWS_ACCESS_KEY_ID" not in os.environ:
+            pytest.skip("AWS_ACCESS_KEY_ID not set in environment")
+        if "AWS_SECRET_ACCESS_KEY" not in os.environ:
+            pytest.skip("AWS_SECRET_ACCESS_KEY not set in environment")
+        if "MINIO_ENDPOINT_URL" not in os.environ:
+            pytest.skip("Minio endpoint URL not set in q")
+
+    @pytest.fixture
+    def repository(self, skip_minio_test_if_missing_env_vars):
```
akshoop, line 31: Do you need the `skip_minio_test_if_missing_env_vars` fixture here?

custom_model_runner/requirements.txt
akshoop: Would be good to pin to specific version too

.gitignore
zohar-mizrahi: Because I have different env files for different environments, that is: .env.s3 and .env.minio

custom_model_runner/requirements.txt
zohar-mizrahi: Yes, we can use it - there was a decision to make it public.

zohar-mizrahi: Ok, I'll allow any versions up to a major upgrade.

tests/functional/test_lazy_loading.py
zohar-mizrahi: Yes, because the given fixture tries to read expected environment variable: `os.environ[MINIO_ENDPOINT_URL]`

akshoop: But it looks like there's only 1 test in this module `test_download_from_minio_success` which already uses `skip_minio_test_if_missing_env_vars` fixture. So I'm saying that having it in `repository` fixture seems redundant

custom_model_runner/requirements.txt
```diff
@@ -26,3 +26,4 @@ termcolor
 packaging
 markupsafe<=2.1.3
 pydantic==2.9.2
+datarobot-storage>=0.0.0,<1.0.0
```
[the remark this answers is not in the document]
YuriyGuts: Curious about the reason to pin this to `0.*` versions specifically. Does it have something to do with artifactory releases vs. OSS releases? The reason I'm asking is Buzok is at `datarobot-storage==2.0.3` and we can't bump DRUM to the latest version because of this conflict.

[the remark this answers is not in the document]
YuriyGuts: Sorry, I should probably rather ask in #drum in case we need the input of related teams on this.

PR #1155, opened 2024-10-28
Prepare GPU envs for 10.2 release

elatt:

## Summary
- Add README to top level directory
- Strip annotations from requirements files
- Upgrade versions of libraries
- Make NIM env public
- Prepare envs for release

## Rationale
With 10.2 wrapped, I'm preparing a release candidate build of the MLOps GPU envs. I think it is ok to make NIM env public because it requires an NGC key to use anyway and making it public will make it more discoverable. I also opted to remove the annotations from the generated requirements files because they were ending up in the version description of the UI and cluttering things.


elatt:
Manually ran `jenkins/test_functional_per_framework.sh {vllm, nim, triton}`

kijimoshib (approved):

amarmudrankit (approved):
LGTM

PR #1156, opened 2024-10-28
[APP-3453] Add python unbuffered environment variable

ben-cutler-datarobot:
The way docker containers run python makes 
```
print("my-statement-here")
```
Causes the print statements to not show up in the DataRobot logging APIs (eg `/api/v2/customApplications/<id>/logs/` )
https://github.com/docker-library/python/issues/604

We could make customers run:
```
print('my-statement-here', file=sys.stderr)
```
but that would be a clunky solution. 

We can add the environment variable `PYTHONUNBUFFERED` to the containers.


shopin32 (approved):

kideh88 (approved):

akshoop (approved):

public_dropin_environments/python311_genai/Dockerfile
```diff
@@ -2,6 +2,9 @@
 # It contains a variety of common useful data-science packages and tools.
 FROM datarobot/dropin-env-base-jdk:ubi8.8-py3.11-jdk11.0.22-drum1.11.5-mlops9.2.8
 
+# This makes print statements show up in the logs API
+ENV PYTHONUNBUFFERED=1
```
klichukb, line 6: I was right about to do this for all environments. Can I ask you to add this to others as well?

klichukb: Thanks a lot!
I think we can introduce to the listed python-aware base dropins. @yakov-g can you please assist publishing new versions of the base image to our dockerhub?

ben-cutler-datarobot: Where is the base image? I'm looking in drum for:
```
FROM datarobot/dropin-env-base:debian11-py3.9-jre11.0.16-drum1.11.5-mlops9.2.8
```
and I can't find that. Should I just do it one-by-one 😕 ?

klichukb: They are in the directory that Erik linked above.
https://github.com/datarobot/datarobot-user-models/tree/master/docker

ben-cutler-datarobot: ok, is there one specific Dockerfile which should be using it? I assumed `dropin_env_base`, but LMK if that's incorrect.

ben-cutler-datarobot: Ah, I didn't realize the jdk one had python. Got it as well.

yakov-g: Added, can merge

yakov-g: Hold on again.
I'll update drum and image tag in the base envs

yakov-g:
Done.
I'll run piplelines to publish base images.
When I worked on those pipelines I did all the testing, but never published new ones. So I'll do this.

PR #1157, opened 2024-10-29
[CODEGEN-1974] Pass association id from chat to moderations

baekdahl:

## Summary
* Pass association id to moderations.
* Don't fail request if reporting deployment stats fails.

## Rationale
We need to use the same association for custom metrics in moderations and when reporting chat predictions to be able to join them in deployment data quality table.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
+++ b/custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
@@ -749,7 +749,7 @@ class PythonModelAdapter(AbstractModelAdapter):
         PythonModelAdapter._validate_unstructured_predictions(predictions)
         return predictions
 
-    def chat(self, completion_create_params, model):
+    def chat(self, completion_create_params, model, association_id):
         if (
             self._target_type == TargetType.TEXT_GENERATION
             and self._guard_pipeline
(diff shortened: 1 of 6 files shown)


custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
```diff
@@ -749,7 +749,7 @@ def predict_unstructured(self, model, data, **kwargs):
         PythonModelAdapter._validate_unstructured_predictions(predictions)
         return predictions
 
-    def chat(self, completion_create_params, model):
+    def chat(self, completion_create_params, model, association_id):
```
elatt, line 752: The API is a little confusing to me now. It's not clear to me readily what chat() function has the association ID as an arg and which doesn't. From this PR, the convention seems to be the _chat() **has** the extra arg and the public chat() does not, so should this be renamed to _chat()?

elatt (approved):

custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
baekdahl: Yeah it's a bit confusing with all these `chat` functions in different places.. My reasoning for not having `_` here in PythonModelAdapter is that this function is public because it is called from `PythonPredictor`.
So the call stack with association id looks like this:
* `PythonPredictor.chat(self, completion_create_params)`
  *  `PythonPredictor._chat(self, completion_create_params, association_id)`
      * `PythonModelAdapter.chat(self, completion_create_params, model, association_id)`
        * `# moderations library`

elatt: yeah that makes sense to have it be public then.

akshoop (approved):

PR #1171, opened 2024-11-01
[CODEGEN-1974] Don't fail request if mlops reporting fails

baekdahl:

## Summary
Catch exception if raised when trying to report that chat request failed.

## Rationale
We want to continue and return a meaningful response even if mlops fails.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
+++ b/custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
@@ -170,7 +170,7 @@ class BaseLanguagePredictor(DrumClassLabelAdapter, ABC):
             deployment = dr.Deployment.get(self._params["deployment_id"])
             return deployment.model["prompt"]
         except Exception:
-            logger.exception(
+            logger.info(
                 "Failed to get prompt column name from deployment. "
                 f"Fallback to default prompt column name ('{DEFAULT_PROMPT_COLUMN_NAME}')"
             )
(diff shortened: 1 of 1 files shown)


gkoundry (approved):
LGTM core modeling

custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
```diff
@@ -170,7 +170,7 @@ def get_prompt_column_name(self):
             deployment = dr.Deployment.get(self._params["deployment_id"])
             return deployment.model["prompt"]
         except Exception:
-            logger.exception(
+            logger.info(
```
klichukb, line 173: However this will hide the errors. Maybe not something user would want to see if this unexpected, but also will make troubleshooting harder.

baekdahl: How about I make it a warning? I have observed that prompt column name is not always set on deployments and things work out as expected anyway

klichukb: I dont mean leaving it an exception. I mean losing any stack trace completely. If its a stack trace with warning its still much better than nothing.

klichukb: So a logger.warning(exc_info=True) would show the failure

baekdahl: Done

klichukb (approved):

yakov-g (approved):

yakov-g:
git pull origin master

PR #1173, opened 2024-11-03
[RAPTOR-11626] Unpin 'datarobot-storage' package

zohar-mizrahi:
## Summary
Inspired by [this post](https://iscinumpy.dev/post/bound-version-constraints/), I included the `datarboto-storage` package without specifying version constraints.

### DIFF
--- a/custom_model_runner/requirements.txt
+++ b/custom_model_runner/requirements.txt
@@ -26,4 +26,4 @@ termcolor
 packaging
 markupsafe<=2.1.3
 pydantic==2.9.2
-datarobot-storage  # see: https://iscinumpy.dev/post/bound-version-constraints/
+datarobot-storage
--- a/custom_model_runner/requirements.txt
+++ b/custom_model_runner/requirements.txt
@@ -26,4 +26,4 @@ termcolor
 packaging
 markupsafe<=2.1.3
 pydantic==2.9.2
-datarobot-storage>=0.0.0,<1.0.0
+datarobot-storage  # see: https://iscinumpy.dev/post/bound-version-constraints/


zohar-mizrahi:
FYI - @YuriyGuts , @elatt

custom_model_runner/requirements.txt
```diff
@@ -26,4 +26,4 @@ termcolor
 packaging
 markupsafe<=2.1.3
 pydantic==2.9.2
-datarobot-storage>=0.0.0,<1.0.0
+datarobot-storage  # see: https://iscinumpy.dev/post/bound-version-constraints/
```
rvorobii, line 29: I see some of the requirements here are pinned, some are not. 
we had multiple issue with something suddenly braking because of unpinned dependencies.
I think we need to either pin it, or at least remove the comment with the link - there're several unpinned dependencies here, the link is probably related to all of them and is more about a general practice.

@yakov-g what do you think? ^

elatt: > we had multiple issue with something suddenly braking because of unpinned dependencies

Please read the post in the comment, this is exactly the fear that is discussed and why it isn't worth it to pin upper bounds.

tl;dr, upper bound pin is no guarantee  some _minor_ change won't break you (only guarantee is to explicitly pin to an exact version but I think we can all agree that is a bad idea for a library package). This false sense of security with the upper bound pin provides, has been shown to cause more trouble than it is worth, time and time again in the real world. As noted above, it is not full-proof solution, it causes more work on the package owner to release new versions of the library simply to bump upper bound pins, otherwise it eventually makes it impossible to install the package to address CVEs.

As for the pins that are already in there, I'd **love** if we started to remove them. I _think_ some are in the legitimately though as I've look at some `git blame` results and at least from the PR descriptions it sounds like there were some incompatibilities with newer libraries. That being said, we should schedule time to fix those incompatibilities so we can move forward as it makes addressing CVEs nearly impossible sometimes with these upper bound pins.

elatt (approved):
LGTM, though I agree that the link to the article in the file may be overkill as we should start to move to having all dependencies with as loose of a constraint as possible. And only comment in the places where we do have upper pins or exact pins to explain why we need to do that in each specific case.

akshoop (approved):

rvorobii (approved):
please remove the link to the article from the requirments.txt, the rest looks good

custom_model_runner/requirements.txt
yakov-g: Let's merge this. 
And I'll try to to unpin today or tomorrow

custom_model_runner/requirements.txt
```diff
@@ -26,4 +26,4 @@ termcolor
 packaging
 markupsafe<=2.1.3
 pydantic==2.9.2
```
yakov-g, line 28: @zohar-mizrahi do we have to pin `pydantic`

zohar-mizrahi: No. I've noticed that you have a follow up PR to test it.

PR #1175, opened 2024-11-05
[RAPTOR-11620] try to unpin DRUM dependencies

yakov-g:

## Summary
Kind of POC PR, let me know what you think.

* Unpin upper constraint for many DRUM deps based on the ticket request and https://github.com/datarobot/datarobot-user-models/pull/1173.
* Pin some of those deps in the env itself.
* Had to remove some deps pinning in python311_genai env.

## Rationale

1. In the Dockerfile for our env we consequently call pip install for `dr_reqirements.txt` and then for `requirements.txt`.
https://github.com/datarobot/datarobot-user-models/blob/master/public_dropin_environments/python311_genai/Dockerfile#L11-L17
This doesn't make sense as they may list conflicting package and it will never be enforced.
E.g. req1.txt: pkg==1.2.3; req2.txt: pkg==3.4.5. In the end, pkg from the file used the last will be installed.

I suggest to get rid of `dr_requirements` and use only `requirements.txt`. In this way packages in our env will be all cross compatible.
If user uses his own `requirements` - let it be, though we also could verify if there are any errors.

2. DRUM per framework functional tests work in the following way. 
There are images for every env in the dockerhub. If env is not changed in the PR, we start harness stage using that image, install there DRUM from source and run tests.(If env has been changed in this PR, we build tmp image first).
As env already has most of the packages installed, Installing DRUM may not update dependencies or may contradict packages already installed in the env look at ^1. Therefore, before running test, I need to install DRUM from source together with requirements related to that env.

### DIFF
--- a/public_dropin_environments/java_codegen/dr_requirements.txt
+++ b/public_dropin_environments/java_codegen/dr_requirements.txt
@@ -1,3 +1,4 @@
 pandas<=2.0.3
+numpy<2.0.0
 pyarrow>=0.14.1,<=14.0.1
 datarobot-drum==1.14.0
--- a/.harness/test_functional_by_framework.yaml
+++ b/.harness/test_functional_by_framework.yaml
@@ -131,6 +131,7 @@ pipeline:
                     shell: Bash
                     command: |-
                       echo "== Assuming running integration tests in framework container (inside Docker), for env: <+pipeline.variables.framework>"
+                      ROOT_DIR="$(pwd)"
 
                       echo "== Installing pytest =="
                       pip install pytest pytest-xdist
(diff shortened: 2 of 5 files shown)


baekdahl (approved):

yakov-g:
@baekdahl do you think it's ok to unpin deps in the `python311_genai` env?

public_dropin_environments/java_codegen/dr_requirements.txt
```diff
@@ -1,3 +1,4 @@
 pandas<=2.0.3
+numpy<2.0.0
```
elatt, line 2: I presume `numpy` is upper pinned here because `pandas` is upper pinned and I presume pandas is lacking the proper pins to be sure to pull in the right version of numpy?

If this is so, maybe it is worth adding some comments on why this stuff is pinned as it is?

elatt (approved):

elatt:
> I suggest to get rid of dr_requirements and use only requirements.txt. In this way packages in our env will be all cross compatible.
If user uses his own requirements - let it be, though we also could verify if there are any errors.

Yeah I never really understood the reason for the two requirements files? Do you know the history on why we had the two files?

yakov-g:
> > I suggest to get rid of dr_requirements and use only requirements.txt. In this way packages in our env will be all cross compatible.
> > If user uses his own requirements - let it be, though we also could verify if there are any errors.
> 
> Yeah I never really understood the reason for the two requirements files? Do you know the history on why we had the two files?

IIRC, it was done in order to tell user: "Hey dr_requirements.txt" is what DR needs to be installed in this env; and `requirements.txt` you can edit on your own - kind of separation to "system" and "user".

I'll create a PR with only one file.

public_dropin_environments/java_codegen/dr_requirements.txt
yakov-g: Likely you are right.
May I skip the comment for now, because we have pins in every env and we never came up with a policy how to properly manage deps in envs.

PR #1177, opened 2024-11-05
[RAPTOR-11637] make single requirements file for envs

yakov-g:

## Summary
In the Dockerfile for our env we consequently call pip install for dr_reqirements.txt and then for requirements.txt.
This doesn't make sense as they may list conflicting package and it will never be enforced.
E.g. req1.txt: pkg==1.2.3; req2.txt: pkg==3.4.5. In the end, pkg from the file used the last will be installed.

I suggest to get rid of dr_requirements and use only requirements.txt. In this way packages in our env will be all mutually compatible.


rvorobii (approved):

baekdahl (approved):

scottp-ml (approved):
LGTM

PR #1180, opened 2024-11-13
[AGENT-6505] Make BaseOpenAiGpuPredictor not care about model name

elatt:

## Summary
Update `BaseOpenAiGpuPredictor` to ignore the `model` ChatAPI param passed in and always override it to be the name of the model we configured when launching the underlying LLM inference server. Also rename `dr_requirements.txt` to `requirements.txt` since this was done in #1177 for other envs, we need to update ours so we can still work with the test harness. 

## Rationale
This is how the legacy predAPI works in that we don't rely on the user to pass care about the model name when making completion requests. In addition, being pedantic on the model name seems to provide no value for BYO LLMs and just adds extra friction to using the ChatAPI to have to know what model is actually running behind the deployment but we already have DR APIs to glean this data by asking for the CustomModelVersion or model package that is currently deployed.


ScottOglesby (approved):

amatanasov (approved):
LGTM Core Modeling

yakov-g (approved):

iamfeniak:
jarvis cherrypick this release/10.2

PR #1190, opened 2024-12-03
Add s5cmd and boto3

elatt:

## Summary
Adds `s5cmd` and `boto3` to vLLM and NIM envs.

## Rationale
Our air-gapped install requires use of minio so we need boto3 in the GPU envs. Also, s5cmd is faster so may be useful to include as well.

### DIFF
--- a/public_dropin_gpu_environments/nim_llm/Dockerfile
+++ b/public_dropin_gpu_environments/nim_llm/Dockerfile
@@ -2,10 +2,15 @@ FROM nvcr.io/nim/meta/llama-3.1-8b-instruct:1.2.2
 
 USER root
 RUN apt-get update && apt-get install -y \
+    curl \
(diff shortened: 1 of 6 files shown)


elatt:
I have not tested this but supposedly vLLM has had S3 loading built in as documented [here](https://docs.vllm.ai/en/latest/models/extensions/runai_model_streamer.html) and confirmed that the base Docker image should have these [libraries installed](https://github.com/vllm-project/vllm/blob/566f10a9293425483ac47892e2a6cd4960177a59/docker/Dockerfile#L300C121-L300C126).

Also, NVIDIA NIM has supposedly [added S3 capability](https://docs.nvidia.com/nim/large-language-models/latest/nim-repository-override.html) too (though we haven't tested it yet)

PR #1198, opened 2024-12-18
[RAPTOR-11783] Deprecate the production mode and replace its functionality with Flask running in multi-process mode

zohar-mizrahi:
## Summary
Deprecate the '--production' option in DRUM and replace its functionality with Flask running in multi-process mode. This functionality was originally introduced in this [commit](https://github.com/datarobot/datarobot-user-models/commit/692825d2d31da86d13f8f41683f7313d028a9ea4). The replacement also involved removing the dependency on Nginx and uWSGI servers.


elatt (approved):
GPU env changes look good

yakov-g (approved):
Great!

akshoop (approved):
Overall LGTM Core Modeling, CC @eric-s-s @scottp-ml for awareness

akshoop:
CC @yura-ant for awareness

akshoop:
Would suggest communicating to team channels about this change once merged

yura-ant:
Hi @zohar-mizrahi  Why we keep --production option as we still use the same dev Flask server?  If we keep  --production option for back ward compatibility but as I know nobody use it. If end user use  --production he probably expect it will be not dev server.

zohar-mizrahi:
> Hi @zohar-mizrahi Why we keep --production option as we still use the same dev Flask server? If we keep --production option for back ward compatibility but as I know nobody use it. If end user use --production he probably expect it will be not dev server.

Because the `production` mode was in the documentation, I think it worth to have a deprecation period. Even if it is only being used internally, I think it still worth having a graceful deprecation time.

public_dropin_environments/python3_keras/Dockerfile
```diff
@@ -17,7 +17,7 @@ WORKDIR ${CODE_DIR}
 COPY ./*.sh ${CODE_DIR}/
 
 ENV WITH_ERROR_SERVER=1
-# Uncomment the following line to switch from Flask to uwsgi server
-#ENV PRODUCTION=1 MAX_WORKERS=1 SHOW_STACKTRACE=1
+# [DEPRECATED] Uncomment the following line to switch to produciton mode
```
tsdaemon, line 20:
```suggestion
# [DEPRECATED] Uncomment the following line to switch to production mode
```

public_dropin_environments/python3_onnx/Dockerfile
```diff
@@ -17,7 +17,7 @@ WORKDIR ${CODE_DIR}
 COPY ./*.sh ${CODE_DIR}/
 
 ENV WITH_ERROR_SERVER=1
-# Uncomment the following line to switch from Flask to uwsgi server
-#ENV PRODUCTION=1 MAX_WORKERS=1 SHOW_STACKTRACE=1
+# [DEPRECATED] Uncomment the following line to switch to produciton mode
```
tsdaemon, line 20: If its deprecated, why keep this line?

tsdaemon (approved):
LGTM for Buzok

public_dropin_environments/python3_onnx/Dockerfile
zohar-mizrahi: Um :thinking: good question. I'll remove it from all related Dockerfiles

PR #1200, opened 2024-12-19
[RAPTOR-11788] Remove mlpiper from the DRUM scoring mode execution path

zohar-mizrahi:
## Summary
In the context of removing mlpiper from DRUM (https://datarobot.atlassian.net/browse/PBMP-6966), in this PR we avoid using mlpiper in the DRUM scoring command execution path, by implementing related functionality directly in DRUM. 

Please note the following:
* There was no attempt to improve the design nor architecture.
* The new implementation simply skip the mlpiper layer, which previously executed an artificial DAG pipeline.
* Eventually, the ultimate goal at this time is to get rid of CVEs (Common Vulnerabilities and Exposures) in DRUM.

Tests are already in-place to validate the `score` command.


custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -780,25 +784,80 @@ def _prepare_prediction_server_or_batch_pipeline(self, run_language):
...
-            infra_pipeline_str = self._prepare_prediction_server_or_batch_pipeline(run_language)
-        elif self.run_mode == RunMode.SCORE:
-            tmp_output_filename = None
-            # if output is not provided, output into tmp file and print
-            if not self.options.output:
-                # keep object reference so it will be destroyed only in the end of the process
-                __tmp_output_file = tempfile.NamedTemporaryFile(mode="w")
-                self.options.output = tmp_output_filename = __tmp_output_file.name
-            # in batch prediction mode infra pipeline == predictor pipeline
+        with self._setup_output_if_not_exists():
```
akshoop, line 791: Can you explain why you're doing options output file setup as a contextmanager instead of directly calling `self._setup_output_if_not_exists()` without yields?

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -780,25 +784,80 @@ def _prepare_prediction_server_or_batch_pipeline(self, run_language):
...
+            run_language = self._check_artifacts_and_get_run_language()
             infra_pipeline_str = self._prepare_prediction_server_or_batch_pipeline(run_language)
+
+            pipeline = json.loads(infra_pipeline_str)
+            if "pipe" not in pipeline or not pipeline["pipe"]:
+                raise DrumCommonException("Pipeline is empty")
+            if "arguments" not in pipeline["pipe"][0]:
```
akshoop, line 798: Should `"pipe"` and `"arguments"` be enums?

akshoop (approved):
Given that functionality should be same, LGTM Core Modeling with comments

akshoop:
CC @eric-s-s for awareness

custom_model_runner/datarobot_drum/drum/drum.py
zohar-mizrahi: Because the printout should take place only after the custom model was executed. This is a more elegant implementation of the previous code.

custom_model_runner/datarobot_drum/drum/drum.py
zohar-mizrahi: I would do it, however, a follow up project is supposed to cleanup all of these code. This `pipeline` is not longer needed and will be removed completely.

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -780,25 +784,80 @@ def _prepare_prediction_server_or_batch_pipeline(self, run_language):
...
+            sc.register_report(
+                "Init time (incl model loading)", "init", StatsOperation.SUB, "start"
+            )
+            sc.register_report("Run time (incl reading CSV)", "run", StatsOperation.SUB, "init")
+            sc.enable()
+
+            try:
+                sc.mark("start")
+                predictor = GenericPredictorComponent()
+                predictor.configure(pipeline["pipe"][0]["arguments"])
```
alexromanyuk, line 822: nit: if caller does not invoke the `configure`, the `GenericPredictorComponent` will not be properly initialized. I.e. call to `self._predictor` evaluates to NPE.

alexromanyuk (approved):

custom_model_runner/datarobot_drum/drum/drum.py
zohar-mizrahi: You're right. Originally, I wanted to only make some technical refactoring without improvements. However, it totally makes sense to make small changes to align with basic and correct SW engineering practice. 
Will fix it.

PR #1202, opened 2024-12-22
[RAPTOR-11789] Remove mlpiper from the DRUM server mode execution path

zohar-mizrahi:
## Summary
In the context of removing mlpiper from DRUM (https://datarobot.atlassian.net/browse/PBMP-6966), in this PR the mlpiper was completely removed, without any re-design nor architecture improvements.

It includes the following:
* Avoid using mlpiper in the DRUM scoring command execution path, by implementing related functionality directly in DRUM. The main change involve in using the `PredictionServer` directly, instead of via mlpiper pipelines.
* Remove imports from mlpiper and copy the related imported object into DRUM (.e.g `AwsHelper`, `ByteConv`) 
* The `resource` directory now contains resources only, without any code.
* Some renaming and refactoring for better readability

Please note the following:
* There was no attempt to improve the design nor architecture.
* The new implementation simply skip the mlpiper layer, which previously executed an artificial DAG pipeline.
* Eventually, the ultimate goal at this time is to get rid of CVEs (Common Vulnerabilities and Exposures) in DRUM.
* Tests are already in-place to validate the score command.

I also run DRUM tests with/without docker, making sure the `mlpiper` is not installed in the related environments. All the tests pass with success.


tests/functional/test_inference_per_framework.py
```diff
@@ -78,7 +78,7 @@
     R_FAIL_CLASSIFICATION_VALIDATION_HOOKS,
     R_PREDICT_SPARSE,
     R_TRANSFORM_SPARSE_INPUT,
-    R_TRANSFORM_SPARSE_OUTPUT,
+    # R_TRANSFORM_SPARSE_OUTPUT,
```
scottp-ml, line 81: Is this intentional?  If so why?

scottp-ml (commented):
This looks fine except for the commented out R test case.

tests/functional/test_inference_per_framework.py
zohar-mizrahi: To clearly indicate that the usage of the given constant was commented later in this file. When I run `black` - without the comment the constant is being removed, although the other comments remain in the file.

scottp-ml (approved):

DrewParallel (approved):

custom_model_runner/datarobot_drum/drum/gpu_predictors/utils.py
```diff
@@ -107,7 +107,7 @@ def _configure_ngc_client(self):
 class S3Client:
     def __init__(self, s3_url, credential):
         import boto3
-        from mlpiper.extra.aws_helper import AwsHelper
+        from aws_helper import AwsHelper
```
elatt, line 110: I think this import is wrong. It looks like you moved it into GPU predictor folder so you can use a relative import here.

custom_model_runner/datarobot_drum/drum/gpu_predictors/utils.py
```diff
@@ -107,7 +107,7 @@ def _configure_ngc_client(self):
 class S3Client:
     def __init__(self, s3_url, credential):
         import boto3
-        from mlpiper.extra.aws_helper import AwsHelper
+        from .aws_helper import AwsHelper
```
piratovskyi, line 110: Why do we need this import to happen inside of the class? Would there be any conflicts otherwise?

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -784,57 +801,45 @@ def _prepare_prediction_server_or_batch_pipeline(self, run_language):
...
-                sc.mark("run")
-            finally:
-                predictor.terminate()
-                sc.mark("end")
-                sc.disable()
+        predictor = (
+            PredictionServer() if self.run_mode == RunMode.SERVER else GenericPredictorComponent()
+        )
+        try:
+            if stats_collector:
```
piratovskyi, line 825: I am nitpicking here: have you considered splitting this `_run_predictions` into two separate functions for RunMode.SCORE  & RunMode.SERVER just to omit these if statements?

custom_model_runner/datarobot_drum/drum/gpu_predictors/utils.py
elatt: My only guess is that this class requires `boto3` installed and I'm not sure if that is a required dependency of `DRUM`? If it is required then we can move this import to the top level, if it is not required then I'd prefer leaving this import to be _just in time_

elatt (approved):

custom_model_runner/datarobot_drum/drum/drum.py
zohar-mizrahi: Yeah, I had some thoughts about it and wondered if we ever want to run statistics for server mode.
Most of the workflow in `_run_predictions` is the same for scoring and server mode. 
So, how about leave it as a follow up enhancement?

custom_model_runner/datarobot_drum/drum/gpu_predictors/utils.py
zohar-mizrahi: Just to clarify - the goal for this project is not to improve the exiting architecture, design or functionality. The goal is to make the necessary changes so that `mlpiper` will not be required. Therefore, many of the changes are only technical steps to achieve that goal.
Specifically, for the comment, yeah, second to @elatt, the `boto3` was probably not a mandatory requirement for DRUM. Even more, I can't find any usage for the `S3Client` in DRUM :shrug: .

elatt: Yeah, I think that code was intended to be used in custom.py scripts so customers could write less code. However, with remote download functionality, this may be superseded?

zohar-mizrahi: I guess so. This can be done as part of the project to cleanup and improve DRUM performance (.e.g DRUM 2.0).
cc: @yakov-g

piratovskyi: I didn't mean to request any change - I was rather curious about this. In any case, thanks for the explanation!

custom_model_runner/datarobot_drum/drum/drum.py
piratovskyi: Sounds good to me

piratovskyi (approved):

PR #1212, opened 2025-01-07
[MODEL-14740]Fix how sparse matrix is handled for comparison.

scottp-ml:

## Summary
Generating the output file for a sparse matrix when the prediction check fails does not work.  This fixes creating that file, and adds testing around it.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/perf_testing.py
+++ b/custom_model_runner/datarobot_drum/drum/perf_testing.py
@@ -808,9 +808,10 @@ class CMRunTests:
 
             message = """
                         Warning: Your predictions were different when we tried to predict twice.
-                        The last 10 predictions from the main predict run were: {}
+                        The last {} predictions from the main predict run were: {}
                         However when we reran predictions on the same data, we got: {}.
                         The sample used to calculate prediction reruns can be found in this file: {}""".format(
+                rows_to_display,
                 preds_full_subset[~matches][:rows_to_display].to_string(index=False),
                 preds_sample[~matches][:rows_to_display].to_string(index=False),
                 __tempfile_sample,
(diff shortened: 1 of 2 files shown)


custom_model_runner/datarobot_drum/drum/perf_testing.py
```diff
@@ -781,23 +779,40 @@ def check_prediction_side_effects(self):
...
+            if is_sparse:
+                _, __tempfile_sample = mkstemp(suffix=InputFormatExtension.MTX)
+                mmwrite(__tempfile_sample, data_subset.sparse.to_coo())
+            else:
+                _, __tempfile_sample = mkstemp(suffix=".csv")
+                data_subset.to_csv(__tempfile_sample, index=False)
+
+            message = """
+                        Warning: Your predictions were different when we tried to predict twice.
+                        The last 10 predictions from the main predict run were: {}
```
akshoop, line 811: nit: you could use formatting for this part too:
```suggestion
                        The last {} predictions from the main predict run were: {}
```

custom_model_runner/datarobot_drum/drum/perf_testing.py
```diff
@@ -781,23 +779,40 @@ def check_prediction_side_effects(self):
 
             self._schema_validator.validate_outputs(preds_sample)
 
-            matches = np.isclose(preds_full_subset, preds_sample, rtol=rtol, atol=atol)
-            if not np.all(matches):
-                if is_sparse:
-                    _, __tempfile_sample = mkstemp(suffix=InputFormatExtension.MTX)
-                    sparse_mat = vstack(x[0] for x in data_subset.values)
```
akshoop, line 788: Looks like the fix was by removing this `vstack`and just using `data_subset.sparse` directly?

akshoop (approved):
LGTM Core Modeling with minor comments

zohar-mizrahi (approved):

custom_model_runner/datarobot_drum/drum/perf_testing.py
scottp-ml: That is correct.  That statement gets the first value instead of the row, which can't be stacked.  It doesn't make sense to do it this way, the built in method does what we want.

PR #1215, opened 2025-01-12
[RAPTOR-11836] Remove support for Apache Arrow

zohar-mizrahi:
## Summary
It appears that Apache Arrow support in DataRobot had remained in alpha maturity and was never integrated into the public API. Including it in DRUM introduces significant complexity when building a drop-in environment that supports it. For instance, starting from a minimal base image like `python:3.11.11-alpine3.20` requires extensive system installations and compiling Arrow from source. Since this feature is not really used, and following a discussion with @Bogdan Klichuk, we have decided to remove its support.


elatt (approved):
GPU envs look fine

yakov-g (approved):

custom_model_runner/requirements.txt
```diff
@@ -14,8 +14,6 @@ strictyaml==1.4.2
 PyYAML
 texttable
 py4j~=0.10.9.0
-# only constrained by other packages, not DRUM
-pyarrow
```
akshoop, line 18: Less dependencies, less tech debt, nice 👍

akshoop (approved):
Since it sounds like Arrow support in DataRobot was only in alpha maturity, removal without going through deprecation process (including communications) makes sense to me. LGTM Core Modeling

akshoop (approved):
Also LGTM from Buzok, CC @mxpoliakov @tsdaemon @YuriyGuts @amatanasov about removal of `pyarrow` from GenAI dropin envs

PR #1219, opened 2025-01-16
[AGENT-6667]  No monitoring if drift tracking is disabled

amarmudrankit:
If the API access is enabled and deployment has drift disabled, no need to report monitoring data

Ensured existing tests are running successfully

Added test cases for chat interface

Added test cases for score interface too

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/description.py
+++ b/custom_model_runner/datarobot_drum/drum/description.py
@@ -4,6 +4,6 @@ All rights reserved.
 This is proprietary source code of DataRobot, Inc. and its affiliates.
 Released under the terms of DataRobot Tool and Utility Agreement.
 """
-version = "1.16.0"
+version = "1.16.1"
 __version__ = version
 project_name = "datarobot-drum"
--- a/custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
+++ b/custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
@@ -8,6 +8,7 @@ import itertools
 import logging
 import os
 import time
+import traceback
 import uuid
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
@@ -85,6 +86,10 @@ class BaseLanguagePredictor(DrumClassLabelAdapter, ABC):
             negative_class_label=negative_class_label,
             class_labels=class_labels,
         )
+        self._deployment = None
+        self._feature_drift_enabled = True
+        self._target_drift_enabled = True
+        self._predictions_data_collections_enabled = True
         self._model = None
         self._code_dir = None
         self._params = None
@@ -127,6 +132,24 @@ class BaseLanguagePredictor(DrumClassLabelAdapter, ABC):
         if not mlops_loaded:
             raise Exception("MLOps module was not imported: {}".format(mlops_import_error))
 
+        if to_bool(self._params.get("allow_dr_api_access")):
+            try:
+                self._deployment = dr.Deployment.get(deployment_id)
+                drift_settings = self._deployment.get_drift_tracking_settings()
+                self._feature_drift_enabled = drift_settings.get("feature_drift", True).get(
+                    "enabled", True
+                )
+                self._target_drift_enabled = drift_settings.get("target_drift", True).get(
+                    "enabled", True
+                )
+                self._predictions_data_collections_enabled = (
+                    self._deployment.get_predictions_data_collection_settings().get('enabled', True)
+                )
+            except Exception as e:
+                logger.warning(f"Failed to get deployment / settings: {e}")
+                traceback.format_exc()
+                return
+
         self._mlops = MLOps()
 
         self._mlops.set_deployment_id(deployment_id)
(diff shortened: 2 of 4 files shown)


custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
```diff
@@ -127,6 +132,24 @@ def _init_mlops(self):
...
+                )
+                self._target_drift_enabled = drift_settings.get("target_drift", True).get(
+                    "enabled", True
+                )
+                self._predictions_data_collections_enabled = (
+                    self._deployment.get_predictions_data_collection_settings().get("enabled", True)
+                )
+            except Exception as e:
+                logger.warning(f"Failed to get deployment / settings: {e}")
+                traceback.format_exc()
```
elatt, line 150: In the interest in keeping all the info together, why not just
```suggestion
                logger.warning(f"Failed to get deployment / settings: {e}", exc_info=True)
```

custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
```diff
@@ -127,6 +132,24 @@ def _init_mlops(self):
...
 
+        if to_bool(self._params.get("allow_dr_api_access")):
+            try:
+                self._deployment = dr.Deployment.get(deployment_id)
+                drift_settings = self._deployment.get_drift_tracking_settings()
+                self._feature_drift_enabled = drift_settings.get("feature_drift", True).get(
+                    "enabled", True
+                )
+                self._target_drift_enabled = drift_settings.get("target_drift", True).get(
+                    "enabled", True
```
elatt, line 143: I forget, can this stuff be toggled on and off dynamically in a deployment? If so, is this code going to pick up such changes?

custom_model_runner/datarobot_drum/drum/description.py
```diff
@@ -4,6 +4,6 @@
 This is proprietary source code of DataRobot, Inc. and its affiliates.
 Released under the terms of DataRobot Tool and Utility Agreement.
 """
-version = "1.16.0"
+version = "1.16.1"
```
akshoop, line 7: Will you update `custom_model_runner/CHANGELOG.md` for this new minor version changes?

akshoop (approved):
Overall LGTM Core Modeling

custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
amarmudrankit: Good point ! I am checking with #mmm about it .

amarmudrankit: You are right !  The flags can be toggled on / off dynamically and not possible to pick those changes unless you query deployment every time.  But, another way is to check if the training data is assigned or not.  For custom models - assigning training data => new model version => model replacement.  So, we can easily verify during the startup if the training data is assigned to this model or not and then decide to report the predictions.

I will close this PR and open a new one later. I think we also need to handle other aspects of reporting with it.

elatt: Conversely, you can
- have a background thread that periodically checks the state of the upstream setting
- not needing threads but similar, perform the check on demand but cache the answer from some amount of time (can add some unexpected latency to requests periodically).

amarmudrankit: Either of the solution will be a big architectural change - ideally would like to have bigger motivation than this optimization to drive it.  We do have some other unrelated changes planned and I will see if any of those could be benefitted by a thread or cache.

elatt: Understood. My thoughts are though that with no training data, it is **always** true that drift tracking will be disabled; however, with training data assigned, it is not always true that drift tracking is enabled (because it can still be turned off at the deployment level for whatever reason) 🤷‍♂️ I'll don't want the perfect to be the enemy of the good though so just wanted to raise these concerns but will defer to you on the ultimate choice.

amarmudrankit: Right ! without training data - drift is definitely off. With training data it can be on or off - but typically will be on.  The worst case where training data is there and drift is off - thats when the API call will be a waste - but that will be a non-typical case.  It will solve more than 80% of the problem.  Currently it always posts - and most of the times for text gen the training data is missing

custom_model_runner/datarobot_drum/drum/language_predictors/python_predictor/python_predictor.py
```diff
@@ -47,13 +47,13 @@ def configure(self, params):
...
         self._model_adapter.load_custom_hooks()
 
-        super(PythonPredictor, self).configure(params)
-
         if to_bool(params.get("allow_dr_api_access")):
             logger.info("Initializing DataRobot Python client.")
             dr_api_endpoint = self._dr_api_url(endpoint=params["external_webserver_url"])
             dr.Client(token=params["api_token"], endpoint=dr_api_endpoint)
 
+        super(PythonPredictor, self).configure(params)
```
zohar-mizrahi, line 55: Would it make sense, instead, to move the code block that initialises the client to the base language predictor?

amarmudrankit: Sure, I think thats a good idea too.  I will make the change in the new PR.

PR #1205, opened 2024-12-24
[RAPTOR-11809] Avoid using the mlops tracking agent to report statistics to DataRobot

zohar-mizrahi:
## Summary
As of today, the `mlops-agent-9.2.8.jar` has a critical CVE. This JAR file is included and used by DRUM for mlops reporting. It is suggested to change the approach and avoid using the agent inside the custom model. Instead, use different MLOps APIs for reporting, which do not require the agent.

# Changes
* Provide the URL and API token to the MLOps API spooler. 
* Set numpy requirement to `numpy<2.0` to comply with latest `pandas` in PyPI
* Set requirement for `datarobot-mlops>=10.2.0`
* Small fix when running DRUM inside a docker
* Add new harness script to check and update Python version to comply with requirements.
* Update mlops reporting functional test to comply with the new spooler change.
* Fix various tests and make some tests as well as harness scripts more robust

PR #1221, opened 2025-01-20
[RAPTOR-11859] Implement a reliable way to inject lines into a Docker file during testing

zohar-mizrahi:
## Summary
Implement a reliable method to programmatically inject lines into a Dockerfile during MLOps monitoring functional test. Instead of relaying on specific instructions inside the Dockerfile, use a dedicated comment for it.

### DIFF
--- a/public_dropin_environments/python3_sklearn/Dockerfile
+++ b/public_dropin_environments/python3_sklearn/Dockerfile
@@ -11,6 +11,8 @@ RUN pip3 install -U pip && \
     pip3 install -U --upgrade-strategy eager --no-cache-dir -r requirements.txt  && \
     rm -rf requirements.txt
 
+# MARK: ADD-HERE. DO NOT DELETE OR MOVE THIS LINE. This line is used by the functional tests.
+
 # Copy the drop-in environment code into the correct directory
 # Code from the custom model tarball can overwrite the code here
 ENV HOME=/opt CODE_DIR=/opt/code ADDRESS=0.0.0.0:8080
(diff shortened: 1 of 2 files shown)


elatt (approved):

akshoop:
These refactor changes seem related to previous changes from: https://github.com/datarobot/datarobot-user-models/pull/1205

public_dropin_environments/python3_sklearn/Dockerfile
```diff
@@ -11,6 +11,8 @@ RUN pip3 install -U pip && \
     pip3 install -U --upgrade-strategy eager --no-cache-dir -r requirements.txt  && \
     rm -rf requirements.txt
 
+# MARK: ADD-HERE. DO NOT DELETE OR MOVE THIS LINE. This line is used by the functional tests.
```
akshoop, line 14: I have concerns about this because it's updating public-facing sklearn drop-in env Dockerfile, which would mean end-users will be able to see this comment. Do we want to allow this? Who exactly is the audience when you say "do not delete or move this line" ? Is it to a DRUM developer or to an end-user?

If you want to go with this approach, then consider updating the wording to be something like:
```suggestion
# MARK: FUNCTIONAL-TEST-ADD-HERE. (This line is used by DRUM functional test automation and can be safely ignored.)
```

akshoop (changes requested):
See question/comment

public_dropin_environments/python3_sklearn/Dockerfile
zohar-mizrahi: Your comment is fair and clearly reflect the its usage. The comment is intended for a contributor, but not for users who use this code and en example for building their own drop-in environment.

akshoop:
Therefore it sounds like we'll expect the updated changes for sklearn drop-in env to be deployed to prod once you update [env_info.json](https://github.com/datarobot/datarobot-user-models/blob/bd6ba88541e00c2373607765b692aa944b3518c2/public_dropin_environments/python3_sklearn/env_info.json) right? That would happen for the next version release of DRUM correct?

akshoop (approved):
Thanks for updating in-line comment in Dockerfile, LGTM Core Modeling

PR #1230, opened 2025-01-23
[AGENT-6684] Update NIM predictor for newer start server script

elatt:

## Summary
Support both `start-server.sh` and `start_server.sh` startup scripts

## Rationale
It appears that in newer NIMs NVIDIA has renamed their start server script.


elatt:
fyi @amarmudrankit I'm trying to push out a change in DRUM as well. I can push to pypi after this merges to master

yakov-g (approved):

kijimoshib (approved):

PR #1232, opened 2025-01-27
[AGENT-6675] Add support for CUSTOM_MODEL_WORKERS param in GPU predictors

elatt:

## Summary
- Bump DRUM version v1.16.3
- Bump DRUM pin in vLLM and NIM env image builds
- Bump base image of vLLM to v0.6.6.post1
- Update watchdog thread logic to work in a forked HTTP worker model
- Update health check to not rely on thread state and look directly at OS process state of exec'd NIM or vLLM process.

## Rationale
To take advantage of the NIM and vLLM dynamic batching capabilities and saturate the GPU, we need to be able to support concurrent requests. DRUM was single threaded until the introduction of `CUSTOM_MODEL_WORKERS` runtime param (https://docs.datarobot.com/en/docs/release/archive-release-notes/v10.2/v10.2.0-mlops.html#custom-model-workers-runtime-parameter) but the implementation did not support GPU workers.

### DIFF
--- a/custom_model_runner/CHANGELOG.md
+++ b/custom_model_runner/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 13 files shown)


kijimoshib (approved):
LGTM

yakov-g (approved):

PR #1234, opened 2025-01-28
[BUZZOK-24833] Fix issue with openai

tsdaemon:

There is an issue about unpinned httpx in which causes openai to fail with an error `Client.__init__() got an unexpected keyword argument 'proxies'`


baekdahl (approved):

sir-Gollum (approved):

public_dropin_gpu_environments/python311_genai/env_info.json
```diff
@@ -3,6 +3,6 @@
   "name": "[DataRobot][NVIDIA] Python 3.11 GenAI",
   "description": "This template environment can be used to create GenAI-powered custom models and includes common dependencies for workflows using OpenAI, Langchain, vector DBs, or transformers in PyTorch. This environment is created to run on machines with GPU support. Similar to other drop-in environments, you can either include a `.pth` artifact or any other code needed to deserialize your model, and optionally, a `custom.py` file.",
   "programmingLanguage": "python",
-  "environmentVersionId": "666b39b6b606304d5d11200f",
+  "environmentVersionId": "6798e070763b883530f848ce",
```
elatt, line 6: Have you tried building this image? When I try to build it I get this error:
```
292.8 Collecting langsmith<0.3,>=0.1.17 (from langchain==0.3.12->-r requirements.txt (line 5))
292.9   Downloading langsmith-0.2.11-py3-none-any.whl.metadata (14 kB)
292.9 INFO: pip is looking at multiple versions of langchain to determine which version is compatible with other requirements. This could take a while.
293.0 ERROR: Cannot install -r requirements.txt (line 4), -r requirements.txt (line 5) and pydantic==2.4.0 because these package versions have conflicting dependencies.
293.0 
293.0 The conflict is caused by:
293.0     The user requested pydantic==2.4.0
293.0     openai 1.30.2 depends on pydantic<3 and >=1.9.0
293.0     langchain 0.3.12 depends on pydantic<3.0.0 and >=2.7.4
293.0 
293.0 To fix this you could try to:
293.0 1. loosen the range of package versions you've specified
293.0 2. remove package versions to allow pip to attempt to solve the dependency conflict
293.0 
293.8 ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
------
Dockerfile:17
--------------------
  16 |     COPY requirements.txt requirements.txt
  17 | >>> RUN pip3 install -r requirements.txt --no-cache-dir --extra-index-url https://download.pytorch.org/whl/cu118 && \
  18 | >>>     rm -rf requirements.txt
  19 |     
```

I don't think this image is actually used or maintained? There are no usable execution envs on app.datarobot.com
image

tsdaemon: Actually I did not, I just ported the same fix. I can fix it and tested, but lets first clarify.
@iamfeniak I would like to understand what is the plan for public execution environments for which we are listed as owners. Do we need to support the GPU version of GenAI environment? Seems like the only version of it in prod is broken, so we had no customers using it. I am happy to add a fix for it, but generally supporting those envs is rather tedious as there is no automated tests for them.

tsdaemon: @iamfeniak @irenekovalchuk I went ahead and fixed this environment, and deployed an upgraded package to all environments.

elatt: Thanks for that. I still ask the question though: is this even env worth maintaining and is there someone that is actively maintaining it (i.e. updating package versions and publishing those out to SaaS).

PR #1235, opened 2025-01-29
[RAPTOR-11892] Fix base image tag generation in the base image builder script

zohar-mizrahi:
## Summary
Misspelling of a certain variable resulted in an invalid image tag.

PR #1229, opened 2025-01-22
[RAPTOR-11834] DR base python image is derived from an official Python image

zohar-mizrahi:
## Summary
After the cleanup of `datarobot-drum` (1.16.1) from the MLOps agent and mlpiper, JRE support is not longer needed for Python based images.

## Changes
- Build the base image from `python:3.11.11-slim-bookworm`.
- Avoid copying MLOps tracking agent jar files into the image.
- Avoid installing any dependencies.
- Apply a fix for `CVE-2024-6345`.
- Improve the `build_image.sh` and use different tagging of the form: `1.0.0-<base-image-tag>`
- Improve harness scripts.

PR #1236, opened 2025-01-29
[RAPTOR-11835] Build the Python based drop-in environments from DR base python image

zohar-mizrahi:
## Summary
Update the Python drop-in environments with the base image that was created after these PRs: https://github.com/datarobot/datarobot-user-models/pull/1229,  https://github.com/datarobot/datarobot-user-models/pull/1235.

The new base image is: `datarobot/dropin-env-base:1.0.0-python-3.11.11-slim-bookworm`

It is an intermediate change that is part of the FedRamp compliance initiative.

Additionally, an improvement has been made to `harness_scripts/functional_by_framework/check_by_framework_run1_step_entrypoint.sh`, ensuring that the requirements.txt from the tested environment is copied to a temporary file, leaving the original file unmodified.

### DIFF
--- a/harness_scripts/functional_by_framework/check_by_framework_run1_step_entrypoint.sh
+++ b/harness_scripts/functional_by_framework/check_by_framework_run1_step_entrypoint.sh
@@ -24,7 +24,7 @@ pip install pytest pytest-xdist
 title "Uninstalling datarobot-drum"
 pip uninstall datarobot-drum datarobot-mlops -y
 
-title "Installing datarobot-drum from source"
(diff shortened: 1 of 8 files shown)


elatt (approved):
LGTM, however looking at the base image now, it looks pretty empty? What are thoughts on just removing it as an intermediary and just referencing the upstream image directly in each of these Dockerfiles?

I ask because in the case of needing to push out CVE fixes, it would be less PRs to not have this intermediary container.

shopin32 (approved):

akshoop (approved):

PR #1240, opened 2025-02-01
[AGENT-6608][AGENT-6612][AGENT-6613] Add use_datarobot_predict ang time series args to server and score command parsers.

kijimoshib:

## Rationale
The changes ensure that time series parameters:
- Only work with the score command
- Require --use-datarobot-predict flag
- Support proper ISO-8601 timestamp validation
- Have proper mutually exclusive options validation
- Are properly passed through the entire pipeline

## Summary
Here's a summary of the main changes to implement time series parameters:

1. **args_parser.py**:
   - Added time series validation to only allow in score command
   - Modified _reg_arg_time_series to check command type
   - Updated register_args to only attach time series args to score_parser
   - Added command validation in verify_options

2. **drum_score_adapter.py**:
   - Added new time series parameters:
     - forecast_point
     - predictions_start_date
     - predictions_end_date
   - Added documentation for new parameters
   - Added instance variables for time series parameters
     
5. **predict_mixin.py**:
   - Added `request.args` as param to predict

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/args_parser.py
+++ b/custom_model_runner/datarobot_drum/drum/args_parser.py
@@ -9,6 +9,7 @@ import os
 import subprocess
 import sys
 from urllib.parse import urlparse
+from datetime import datetime
 
 import trafaret as t
 from datarobot_drum.drum.description import version
@@ -406,6 +407,19 @@ class CMRunnerArgsRegistry(object):
                     "[DEPRECATED] Run prediction server in production mode, which means Flask running in multi-process. "
                     f"The argument can also be provided by setting {ArgumentOptionsEnvVars.PRODUCTION} env var. "
                     "(It requires --max-workers option)."
+                ),
+            )
+
+    @staticmethod
+    def _reg_arg_use_datarobot_predict(*parsers):
+        for parser in parsers:
+            parser.add_argument(
+                ArgumentsOptions.USE_DATAROBOT_PREDICT,
+                action="store_true",
+                default=False,
+                help="Use datarobot-predict as a default predictor for scoring code models, including time-series. "
+                "The argument can also be provided by setting {} env var.".format(
+                    ArgumentOptionsEnvVars.USE_DATAROBOT_PREDICT
                 ),
             )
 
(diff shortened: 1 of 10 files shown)


akshoop (approved):
Overall LGTM Core Modeling

akshoop:
CC @elatt for awareness

yakov-g (approved):

custom_model_runner/datarobot_drum/drum/args_parser.py
```diff
@@ -863,6 +877,84 @@ def get_arg_option(options: argparse.Namespace, arg_option: str):
...
+            datetime.fromisoformat(arg.replace("Z", "+00:00"))
+            return arg
+        except ValueError:
+            raise argparse.ArgumentTypeError(f"The value {arg} is not a valid ISO-8601 timestamp!")
+
+    @staticmethod
+    def _reg_arg_time_series(*parsers):
+        def require_both_dates(arg):
+            # First check if this is score command
+            if ArgumentsOptions.SCORE not in sys.argv[1]:
```
elatt, line 892: Is this commonly how we do this type of argument validation in DRUM, e.g. we manually parse through `sys.argv` looking for other options? I see you doing this below but this will break down if there is a mixture of things set as env vars and things set and CLI arguments right?

If this is the existing pattern then so-be-it but if this is the first time we have such interdependencies may I suggest instead to let the args all get parsed (and only include validation that is local to the specific flag, such as parsing ISO-8601 format) and then `verify_options()` is the place to make sure the set of options as a whole are logically consistent?

custom_model_runner/datarobot_drum/drum/args_parser.py
```diff
@@ -1186,6 +1282,18 @@ def verify_options(options):
...
+        time_series_args = ["forecast_point", "predictions_start_date", "predictions_end_date"]
+        if any(getattr(options, arg, None) is not None for arg in time_series_args):
+            if options.subparser_name != ArgumentsOptions.SCORE:
+                print("\nError: Time series arguments can only be used with the score command")
+                exit(1)
+            if not getattr(options, "use_datarobot_predict", False):
+                print(
+                    "\nError: Time series arguments require --use-datarobot-predict to be enabled"
+                )
+                exit(1)
```
elatt, line 1295: Can this get moved into a helper function: `CMRunnerArgsRegistry.verify_timeseries_args(...)

elatt (approved):
LGTM, just some minor note about CLI arg verification style

custom_model_runner/datarobot_drum/drum/args_parser.py
elatt: Also, I think it breaks down if users do stuff like `--myflag=12356` vs `--myflag 123456` where I think Python's argparse will accept both styles equally? So yeah, I think it is always best to do pure argument parsing in one phase and then logical validation in a second phase.

PR #1241, opened 2025-02-02
[RAPTOR-11899] Add new scikit-learn FedRamp compliant drop-in environment

zohar-mizrahi:
## Summary
Add a new FedRAMP-compliant scikit-learn environment, built from the chain-guard image: `datarobotdev/mirror_chainguard_datarobot.com_python-fips:3.11`. The environment's image is created through a two-step build process. In the first step, the development chain-guard image (`datarobotdev/mirror_chainguard_datarobot.com_python-fips:3.11`) is used for various packages installation. In the second step, the artifacts from the first step are copied into the final image.

#### NOTE
The new FedRAMP-compliant scikit-learn environment is not yet supported by the DataRobot system. Support is in progress and will be available soon. In the meantime, it should not be installed in DataRobot.

## Changes
* Add a new public drop-in environment: `python3_sklearn_fedramp_compliant` (Building the image requires access to the DR private Docker Hub).
* Use latest scikit-learn, scipy, numpy and pandas releases in related sklearn environments.
* Regenerate `sklearn` model artifacts for newer scikit-learn release.
* Enable reading a security context from a drop-in environment (if exists) during a functional test.
* Use an optional security context when preparing and building a Docker image with DRUM.
* Small improvements.
* The MLOps reporting functional test requires pulling a Chainguard image from the DR private repository. Therefore, log in to Docker Hub during harness pipeline execution.
* Add functional test for the FedRAMP-compliant image.


tests/fixtures/drop_in_model_artifacts/SKLearn.md
zohar-mizrahi, line 1: This is an auto-generated file produced during the creation of new scikit-learn model artifacts.
The related command is [here](https://github.com/datarobot/datarobot-user-models/blob/a8380dfddbbb4b9bafea21cec13b9e638e43517c/tools/generate-model-artifacts.sh#L44).

tests/fixtures/python3_sklearn_binary_hyperparameters/custom.py
```diff
@@ -68,7 +68,7 @@ def fit(
     estimator.set_params(
         **{
             "model__penalty": parameters["penalty"],
-            "model__dual": parameters["dual"],
+            "model__dual": bool(parameters["dual"]),
```
zohar-mizrahi, line 71: Newer scikit-learn releases required the `dual` to be boolean.

tests/functional/test_validation_check.py
```diff
@@ -136,10 +135,7 @@ def test_validation_check(
             assert_if_fail=False,
         )
 
-        if language == NO_CUSTOM:
-            assert re.search(r"Null value imputation\s+FAILED", stdo)
-        else:
-            assert re.search(r"Null value imputation\s+PASSED", stdo)
+        assert re.search(r"Null value imputation\s+PASSED", stdo)
```
zohar-mizrahi, line 138: The null value imputation test functions properly with scikit-learn models generated by the latest releases.

public_dropin_environments/python3_sklearn_fedramp_compliant/Dockerfile
```diff
@@ -0,0 +1,73 @@
+# This is a private chain-guard development image that is stored in DataRobot's private registry.
+# Replace it with your own development chain-gaurd image if you build your own.
+ARG BASE_ROOT_IMAGE=datarobotdev/mirror_chainguard_datarobot.com_python-fips:3.11-dev
+FROM ${BASE_ROOT_IMAGE} AS build
+
+USER root
+
+# Fix these CVEs: CVE-2025-0395, CVE-2024-13176, CVE-2025-0395
+RUN apk update && apk upgrade --available && \
```
elatt, line 9: This looks a little weird to me. This is the build image, why do we care what libs are installed into it? Are we statically linking against these when we build pip wheels (which then get copied to prod image)?

public_dropin_environments/python3_sklearn_fedramp_compliant/Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+    apk add --no-cache glibc=2.40-r6 glibc-dev=2.40-r6 glibc-locale-posix=2.40-r6 \
+    ld-linux=2.40-r6 libcrypt1=2.40-r6 nss-db=2.40-r6 nss-hesiod=2.40-r6 \
+    openssl=3.4.0-r6 libcrypto3=3.4.0-r6 libssl3=3.4.0-r6 && \
+    rm -rf /var/cache/apk/* && \
+    mkdir -p /opt/code  # Required for custom-models
```
elatt, line 14: This is in the build image, is it really necessary?

public_dropin_environments/python3_sklearn_fedramp_compliant/Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+RUN pip install -U pip && \
+    pip install --no-cache-dir -r requirements.txt  && \
+    rm -rf requirements.txt
+
+# MARK: FUNCTIONAL-TEST-ADD-HERE. (This line is used by DRUM functional test automation and can be safely ignored.)
+
+# This is a private production chain-guard image that is stored in DataRobot's private registry.
+# Replace it with your own production chain-gaurd image if you build your own.
+FROM datarobotdev/mirror_chainguard_datarobot.com_python-fips:3.11
+
```
elatt, line 31: I think you want to set the Python unbuffered env var in this image

public_dropin_environments/python3_sklearn_fedramp_compliant/Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+    rm -rf /var/cache/apk/* && \
+    mkdir -p /opt/code  # Required for custom-models
+
+USER nonroot
+
+ENV HOME=/home/nonroot
+WORKDIR ${HOME}
+
+COPY requirements.txt requirements.txt
+RUN pip install -U pip && \
```
elatt, line 22: Why aren't you using a venv?

public_dropin_environments/python3_sklearn_fedramp_compliant/Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+
+# This is a private production chain-guard image that is stored in DataRobot's private registry.
+# Replace it with your own production chain-gaurd image if you build your own.
+FROM datarobotdev/mirror_chainguard_datarobot.com_python-fips:3.11
+
+USER root
+
+# Copy updated libraries free of CVEs
+COPY --from=build /lib /lib
+COPY --from=build /usr/lib /usr/lib
```
elatt, line 36: This feels like a pretty big hammer. Is the issue is that our CIIT team hasn't pulled the latest mirrored image. I would just wait for upstream to refresh the image because this could have unexpected side effects I think.

public_dropin_environments/python3_sklearn_fedramp_compliant/env_info.json
```diff
@@ -0,0 +1,15 @@
+{
+  "id": "679b441e9c77601e2fadd45e",
+  "name": "[DataRobot] Python 3.9 Scikit-Learn Drop-In [FedRAMP-compliant]",
```
elatt, line 3:
```suggestion
  "name": "[DataRobot][FedRAMP-compliant] Python 3.9 Scikit-Learn Drop-In",
```

I feel like federal customers will want to only use compliant images so better to have that information first?

elatt (approved):

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -976,16 +990,16 @@ def _prepare_docker_command(self, options, run_mode, raw_arguments):
...
+                        f' -v "{fit_row_weights_filename}":{in_docker_fit_row_weights_filename}'
                     )
                     DrumUtils.replace_cmd_argument_value(
                         in_docker_cmd_list,
                         ArgumentsOptions.WEIGHTS_CSV,
                         in_docker_fit_row_weights_filename,
                     )
 
-        docker_cmd += " {} {}".format(docker_cmd_args, options.docker)
+        docker_cmd += f" {options.docker}"
```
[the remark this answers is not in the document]
zohar-mizrahi: Yes, it seemed redundant and even more confusing since some of the arguments were already passed directly to `docker_cmd`.

harness_scripts/functional_general/mlops_reporting_entrypoint.sh
```diff
@@ -4,6 +4,13 @@ script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 . ${script_dir}/../common/common.sh
 . ${script_dir}/../../tools/create-and-source-venv.sh
 
+DOCKER_HUB_SECRET=$1
```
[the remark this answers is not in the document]
zohar-mizrahi: I actually tried that without much success.
Harness is supposed to support a secured mechanism for setting up environment variables and be able to reference them from a bash script, however I did not managed to make work - maybe because of the Harness release that we use.
How about working on it in a separate ticket?

public_dropin_environments/python3_sklearn_fedramp_compliant/Dockerfile
zohar-mizrahi: It was done to be able to install certain libraries to fix existing CVEs (see the comment above it).
But maybe, it should not be our concern?

public_dropin_environments/python3_sklearn_fedramp_compliant/Dockerfile
zohar-mizrahi: The `mkdir` tool is not included by default in the production image. As a temporary workaround, I copy it to the production image because it is used by the DR backend. The goal is to remove this dependency once the related functionality in the DR backend is fixed.

public_dropin_environments/python3_sklearn_fedramp_compliant/Dockerfile
zohar-mizrahi: Is it anyway a temporary stage, where stuff are being installed by the `non-root` user. 
The products are installed under the `non-root` user home directory and later being copied to the production image.
Do you think we should do it in `venv`?

public_dropin_environments/python3_sklearn_fedramp_compliant/Dockerfile
zohar-mizrahi: Maybe you're right. We can wait. Having said that, the libraries are supposed to be comparable with those in the production, except for the upgraded ones.
Do you know what is the timeline for the CIIT team to refresh the images?

harness_scripts/functional_general/mlops_reporting_entrypoint.sh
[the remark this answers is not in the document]
elatt: Absolutely. I have you the overall ✅ for the PR. All my comments are simply suggestions.

public_dropin_environments/python3_sklearn_fedramp_compliant/Dockerfile
elatt: No, I think it is fine to copy the `mkdir` binary to the prod image, I just don't understand the point of making the `/opt/code` directory in the build image? The essential thing is for it to be in the prod image (and have the right permissions, as you are already doing); this just looks redundant?

public_dropin_environments/python3_sklearn_fedramp_compliant/Dockerfile
elatt: Yeah I think artifacts inherited from the base image should be the responsibility of the base image to update (our internal tooling is getting better at detecting the base image and not reporting duplicate CVEs for all downstream images). While your code here will make the image CVE free, I think it adds extra maintenance for us (i.e. to remove this code when the base has updated their deps or update them when there is a future CVE found in one of these libraries).

public_dropin_environments/python3_sklearn_fedramp_compliant/Dockerfile
elatt: Ah, I _think_ I understand.  I think that `pip` automatically detects when it doesn't have permission to install packages into the system location, then it falls back to installing into users's homedir. You could make that a little more obvious here by just explicitly setting the `--user` flag to `pip install --user`. Then I'm fine with not using a venv.

public_dropin_environments/python3_sklearn_fedramp_compliant/Dockerfile
elatt: Looks like they have a pipeline that runs daily (https://app.harness.io/ng/account/oP3BKzKwSDe_4hCFYw_UWA/module/ci/orgs/SDTK/projects/chainguardmirrors/pipelines/mirror_chainguard_datarobotcom_pythonfips/triggers/?storeType=INLINE)

I can ask in Slack if the team is aware that they are shipping images with fixable CVEs

elatt: 😕 I just now pulled down the `datarobotdev/mirror_chainguard_datarobot.com_python-fips:3.11{,-dev}` images and did a trivy scan of both images and found zero CVEs.

elatt: Ok so I pulled down a 10 day old image from our mirror and it had the CVEs in question
```
datarobotdev/mirror_chainguard_datarobot.com_python-fips:3.11-drbuild-291 (chainguard 20230214)

Total: 4 (UNKNOWN: 0, LOW: 0, MEDIUM: 4, HIGH: 0, CRITICAL: 0)

┌────────────────────┬───────────────┬──────────┬────────┬───────────────────┬───────────────┬────────────────────────────────────────────────────────┐
│      Library       │ Vulnerability │ Severity │ Status │ Installed Version │ Fixed Version │                         Title                          │
├────────────────────┼───────────────┼──────────┼────────┼───────────────────┼───────────────┼────────────────────────────────────────────────────────┤
│ glibc              │ CVE-2025-0395 │ MEDIUM   │ fixed  │ 2.40-r5           │ 2.40-r6       │ glibc: buffer overflow in the GNU C Library's assert() │
│                    │               │          │        │                   │               │ https://avd.aquasec.com/nvd/cve-2025-0395              │
├────────────────────┤               │          │        │                   │               │                                                        │
│ glibc-locale-posix │               │          │        │                   │               │                                                        │
│                    │               │          │        │                   │               │                                                        │
├────────────────────┤               │          │        │                   │               │                                                        │
│ ld-linux           │               │          │        │                   │               │                                                        │
│                    │               │          │        │                   │               │                                                        │
├────────────────────┤               │          │        │                   │               │                                                        │
│ libcrypt1          │               │          │        │                   │               │                                                        │
│                    │               │          │        │                   │               │                                                        │
└────────────────────┴───────────────┴──────────┴────────┴───────────────────┴───────────────┴────────────────────────────────────────────────────────┘
```

However, any newer image since then has everything fixed so I'm not sure why you are still seeing the issue?

yakov-g (approved):

public_dropin_environments/python3_sklearn/requirements.txt
```diff
@@ -1,6 +1,6 @@
...
+datarobot-drum==1.16.3
 
-scikit-learn==1.3.2
-scipy>=1.1,<1.11
-numpy==1.23.5
-pandas<=2.0.3
+scikit-learn==1.6.1
+scipy==1.15.1
+numpy==2.2.2
+pandas==2.2.3
```
akshoop, line 6: Don't forget to update env_info.json for the regular py3_sklearn drop-in env

akshoop (approved):
LGTM Core Modeling

public_dropin_environments/python3_sklearn_fedramp_compliant/Dockerfile
zohar-mizrahi: The point is that if we don't copy the `mkdir` tool, we won't be able to create the directory in the production image. Therefore, we need to create the required path in the build image and copy it to the production image.
In any case, if we decide to always copy the mkdir tool, we can simply move the path creation to the production image.

public_dropin_environments/python3_sklearn_fedramp_compliant/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+ENV PATH=${HOME}/.local/bin:${PATH}
+ENV CODE_DIR=/opt/code
+ENV ADDRESS=0.0.0.0:8080
+
+# This makes print statements show up in the logs API
+ENV WITH_ERROR_SERVER=1 \
+    PYTHONUNBUFFERED=1
+
+COPY --from=build ${HOME} ${HOME}
+COPY --chown=nonroot:nonroot ./*.sh ${CODE_DIR}/
```
elatt, line 71: Won't this create /opt/code for you so you don't need to copy the empty directory from the build image?

public_dropin_environments/python3_sklearn_fedramp_compliant/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+COPY --from=build /bin/chmod /bin/chmod
+
+# Required for DR backend to create /opt/code (Maybe required by applications other than custom-models)
+COPY --from=build /bin/mkdir /bin/mkdir
+
+# Required for custom-models to install dependencies
+COPY --from=build /usr/bin/pip /usr/bin/pip
+
+# Just for convenience
+COPY --from=build /bin/ls /bin/ls
```
elatt, line 28: FWIW, I posted a question to ChainGuard about this approach of copying these binaries from dev image to production image to make sure this isn't an anti-pattern or anything (https://datarobot.slack.com/archives/C07LR41TMV2/p1738692838032889). I'm only curious because I think it is a clever approach, yet I didn't see them recommending it in any of their documentation I have read. Instead I simply see them talk about using the `-dev` image in production if you need to customize the image 🤷‍♂️

elatt (approved):

public_dropin_environments/python3_sklearn_fedramp_compliant/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+COPY --from=build /bin/chown /bin/chown
+
+# Required to change the permissions of the 'start_server.sh' that is copied into the managed-image
+COPY --from=build /bin/chmod /bin/chmod
+
+# Required for DR backend to create /opt/code (Maybe required by applications other than custom-models)
+COPY --from=build /bin/mkdir /bin/mkdir
+
+# Required for custom-models to install dependencies
+COPY --from=build /usr/bin/pip /usr/bin/pip
```
elatt, line 25: Since you are doing `RUN python -m ensurepip --default-pip &&` below do you still need to have this line?

public_dropin_environments/python3_sklearn_fedramp_compliant/Dockerfile
elatt: tl;dr: this is one of the two approved approaches and IMO, for this usecase, the better of the two but I can tag you in the thread so you are familiar with the other approach as well.

PR #1244, opened 2025-02-05
[RAPTOR-11931] Add a FedRamp-compliant flag to the environment metadata

zohar-mizrahi:
## Summary
As part of the FedRAMP initiative, we are introducing a new execution environment that is FedRAMP-compliant. It would be beneficial to include a specific indicator in its metadata file, env_info.json, to denote compliance. This will be used later during installation in DR.


zohar-mizrahi:
For now, it looks like it is not really needed, because we assume that no one will ever change the base images for non-fips compliant images

PR #1245, opened 2025-02-05
[RAPTOR-11936] Add virtual environment to the FeRamp-compliant scikit-learn environment

zohar-mizrahi:
## Summary
Instead of installing Python dependencies at the user level, create a virtual environment and install all required packages in it. This should be done by the root user. The container's entrypoint should activate the virtual environment before running drum. Additionally, update start_server.sh to activate the virtual environment in case it is called directly, such as when the entrypoint is overridden (as done by the DR backend).

## Notes
* All installations are done by the root user.
* The virtual environment is placed under `/opt/venv`
* Script files are copied to `/opt/code` by the root - in any case the ownership of this path is changed by DR backend.


zohar-mizrahi:
https://github.com/datarobot/datarobot-user-models/pull/1246

PR #1246, opened 2025-02-05
[RAPTOR-11936] Add virtual environment to the FIPS-compliant scikit-learn environment

zohar-mizrahi:
## Summary
Instead of installing Python dependencies at the user level, create a virtual environment and install all required packages in it. This is done by the root user. The container's entrypoint activates the virtual environment before running drum. Additionally, update start_server.sh to activate the virtual environment in case it is called directly, such as when the entrypoint is overridden (as done by the DR backend).

## Notes
* All installations are done by the root user.
* The virtual environment is placed under /opt/venv
* Script files are copied to /opt/code by the root - in any case the ownership of this path is changed by DR backend.
* The changes include rolling back the use of the security context in the execution environment.

## Tests
* One functional tests (MLOps reporting).
* Manually tested end-to-end by installing it in DR, creating a custom model with dependencies, and running custom model testing.
* (Harness pipeline test is still in progress - will come later).


akshoop:
is the tarball `public_dropin_environments/python3_sklearn_fedramp_compliant.tar.gz` intentional?

public_fips_dropin_environments/python3_sklearn/README.md
```diff
@@ -20,7 +20,7 @@ For specific version information, see [requirements](requirements.txt).
 
 ## Instructions
 
-1. From the terminal, run `tar -czvf py_dropin.tar.gz -C /path/to/public_dropin_environments/python3_sklearn_fedramp_compliant/ .`
+1. From the terminal, run `tar -czvf py_dropin.tar.gz -C /path/to/public_dropin_environments/python3_sklearn_fips_compliant/ .`
```
akshoop, line 23:
```suggestion
1. From the terminal, run `tar -czvf py_dropin.tar.gz -C /path/to/public_fips_dropin_environments/python3_sklearn/ .`
```

tests/functional/test_mlops_monitoring.py
```diff
@@ -16,7 +16,7 @@
 from subprocess import TimeoutExpired
 import sys
 import time
-
+from typing import Tuple, Optional
```
akshoop, line 19: Unnecessary imports

akshoop (approved):
LGTM Core Modeling

akshoop:
CC @elatt for awareness

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -893,12 +875,9 @@ def _prepare_docker_command(
...
         in_docker_runtime_parameters_file = "/opt/runtime_parameters.yaml"
 
-        security_context = security_context or SecurityContext(
-            user_id=os.getuid(), group_id=os.getgid()
-        )
         docker_cmd = (
-            "docker run --rm --init --entrypoint '' --interactive"
-            f" --user {security_context.user_id}:{security_context.group_id}"
+            f"docker run --rm --init --entrypoint '' --interactive"
+            f" --user {os.getuid()}:{os.getgid()}"
```
klichukb, line 880: Are these changes made for running locally? Wondering why is this in scope of the PR

public_fips_dropin_environments/python3_sklearn/Dockerfile
```diff
@@ -27,28 +27,29 @@ COPY --from=build /usr/bin/pip /usr/bin/pip
...
+    python -m ensurepip --default-pip && \
+    python -m pip install --upgrade pip && \
+    python -m pip install --no-cache-dir -r requirements.txt"
 
-# MARK: FUNCTIONAL-TEST-ADD-HERE. (This line is used by DRUM functional test automation and can be safely ignored.)
-
-ENV HOME=/home/nonroot
-ENV PATH=${HOME}/.local/bin:${PATH}
+ENV PATH=${VIRTUAL_ENV}/bin:${PATH}
+ENV HOME=/opt
```
klichukb, line 41: Surprised that having $HOME as /opt does no cause issues (we use this even now in production)

Overall I would like us to make sure that in a custom model you can write to /opt/code from the custom.py (to verify needed permissions).

custom_model_runner/datarobot_drum/drum/drum.py
zohar-mizrahi: Just rolled it back to the original implementation. This is somehow imitates what we do in DR backend - an arbitrary user.

public_fips_dropin_environments/python3_sklearn/Dockerfile
zohar-mizrahi: It is supposed to work, because the backend changes ownership recursively. But as you said - we better test it.

elatt: Yeah, I've run up against the functional tests in this repo vs the way _things actually work in DR Platform_ related to this. Basically the `Dockerfile` (and thus image) you build from this repo isn't exactly what will be built and run in DataRobot because in DR, we layer on more commands that chmod/chown recursively the contents of `/opt/code` and also in DR we always run as user `1000:1000` where in tests in this repo, I think we run as the current UID of the user running the tests?

elatt (approved):
LGTM

klichukb (approved):

public_fips_dropin_environments/python3_sklearn/Dockerfile
klichukb: Yeah right we just override all the relevant paths to the user we want in DR( 1000:1000 )

PR #1250, opened 2025-02-06
[AGENT-6757] Add NIM EmbedQA

elatt:

## Summary
Adds new env for NIM embedding model and adds ability of NIM predictor to support unstructured prediction requests.

## Rationale
There are other NIMs from just LLMs so any expident way to suppor them is to allow `/predictUnstructured/` to be called. However, we do not provide a default implementation so users will need to provide a `custom.py` to appropriate hook to implement the desired behavior specific to the non-LLM NIM they have launched.


alexromanyuk (approved):

yakov-g (approved):

public_dropin_gpu_environments/nim_embedqa/start_server.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
+# Copyright 2024 DataRobot, Inc. and its affiliates.
```
akshoop, line 2:
```suggestion
# Copyright 2025 DataRobot, Inc. and its affiliates.
```

akshoop (approved):
LGTM Core Modeling

PR #1252, opened 2025-02-07
Add script to generate fully pinned dependencies lists of existing images

G-Despacito:
## Summary
This PR introduces a Bash script to generate fully pinned dependency lists for all existing images with `requirements.in`. The script ensures that dependencies are locked using `pip-compile`, maintaining consistency and reproducibility across environments.  

## Rationale
- By pinning dependencies, we prevent unexpected version changes that could break environments.  
- Uses `pip-tools` and a specific `pip` version to ensure consistent behavior across different environments.


tools/reconcile-dependencies.sh
```diff
@@ -0,0 +1,41 @@
...
+env_dirs=("../public_dropin_gpu_environments" "../public_dropin_environments")
+
+# Iterate over each environment directory
+for dir in "${env_dirs[@]}"; do
+    if [[ -d "$dir" ]]; then  # Check if the directory exists
+        for env in "$dir"/*; do
+            if [[ -d "$env" && -f "$env/requirements.in" ]]; then  # Check for a valid environment with requirements.in
+                echo "Processing $env"
+                # Generate a fully pinned requirements.txt file from requirements.in
+                pip-compile --index-url=https://pypi.org/simple \
```
NolanJMcCafferty, line 26: just to confirm, even without specifying `--output-file` here it creates the requirements.txt in the correct subdirectory?

NolanJMcCafferty (approved):

PR #1259, opened 2025-02-11
[RAPTOR-11954] Fix the default image tag in the environment build and publish Harness pipeline

zohar-mizrahi:
## Summary
The default image tag value should be constructed from the drop-in environments folder and the drop-in environment name.

### DIFF
--- a/.harness/env_image_publish.yaml
+++ b/.harness/env_image_publish.yaml
@@ -75,4 +75,4 @@ pipeline:
       type: String
       description: ""
       required: true
-      value: <+input>.default(<+pipeline.variables.env_name>_latest)
+      value: <+input>.default(<+pipeline.variables.env_folder>_<+pipeline.variables.env_name>_latest)


.harness/env_image_publish.yaml
```diff
@@ -75,4 +75,4 @@ pipeline:
       type: String
       description: ""
       required: true
-      value: <+input>.default(<+pipeline.variables.env_name>_latest)
+      value: <+input>.default(<+pipeline.variables.env_folder>_<+pipeline.variables.env_name>_latest)
```
elatt, line 78: Why do we encode so much in the tag and not have this part of the image name?

elatt (approved):

.harness/env_image_publish.yaml
yakov-g: Because every name == repository which is managed by CI team.
So, assuming, all these images have one purpose - testing - at the time of repo creation  we (I and CI) decided that it can be kept in a single repository. 

Now question/memo?
Should we have a single repository for all the envs images we discuss in the Envs Lifecycle project...

yakov-g (approved):

PR #1273, opened 2025-02-13
Cleanup vllm multi-gpu handling

elatt:

## Summary
Be sure to check for other CLI options that are aliases for the parallelism modes. Also add runtime param to switch default mode.

## Rationale
For quantized models, you can't use `-tp` and need to use `-pp` so it is good to provide this flexibility.


kijimoshib (approved):
🎉

PR #1276, opened 2025-02-14
Setup harness pipeline to reconcile dependencies for each environment

G-Despacito:
## Summary
This pull request introduces several new CI pipelines for various public drop-in environments. These pipelines are designed to reconcile Python dependencies automatically.

## Rationale
- When a PR is opened or updated, if there are changes in an env folder (e.g., public_dropin_environments/python3_keras), the pipeline for that env will be triggered. 
- It will generate a new requirements.txt based on the latest commit. If there are any changes, the pipeline will commit the update; otherwise, no changes will be made.
- If the pipeline creates a new commit, it will trigger another round of GitHub checks, ensuring the updates are validated before merging into master.


shopin32 (approved):

tsdaemon (commented):
Never had a chance to discuss this, but maybe its not too late. Would it make sense to make this a PR check instead of a post PR action? Its easier to error in PR when it need a reconcilement, so a PR author can fix it immediately

public_dropin_gpu_environments/triton_server/requirements.txt
```diff
@@ -35,14 +35,14 @@ google-resumable-media==2.7.2
...
 idna==3.10
-importlib-metadata==8.5.0
 isodate==0.7.2
 itsdangerous==2.2.0
 jinja2==3.1.5
 jmespath==1.0.1
 julia==0.5.7
 markupsafe==2.1.3
 memory-profiler==0.61.0
+mlpiper==2.6.2
```
klichukb, line 45: This definitely looks wrong. Lets discuss this.

klichukb (changes requested):
Let's not yet embed any worklow changes into existing environments.
They are considered to be part of a critical flow.
Lets define a test environment, if that's possible even in a different directory and have this be the sandbox.

We should work on a single happy path for a single environment rather than moving by  work-in-progress layers.

2. Are version bumps in this PR done as part of the resulting automation? I certainly see some dependencies that must not be here.

PR #1280, opened 2025-02-17
[MMM-18250] Fill in the VDB custom-metrics

rickwporter-datarobot:

## Summary

During VDB deployment creation, the DataRobot application creates 4 "custom-metrics". These changes update DRUM with processing for populating those custom-metrics.

## Rationale

All VDB deployments should have a common set of custom-metrics.

## Design

When initializing the adapter, this code pulls the `/api/v2/deployments/{id}/customMetrics/` to find the overlap (by name) of metrics with the VDB specific custom-metrics. If overlap is found, then a `CustomMetricsProcessor` is added to the adapter. The `CustomMetricsProcessor` is initialized with processors for the VDB metrics -- this is NOT something that should need to be done on each prediction.

The `CustomMetricsProcessor` has a list of metric processors that look at the resulting dataframe and calculate a score. The score result is uploaded as part of a bulk-upload that is done once per prediction. There are currently four different metric processors that potentially contribute to the metrics:
* `CitationTokenCount` - counts the total number of tokens in the citation columns
* `CitationTokenAverage` - averages the number of tokens in citation columns for a given response
* `DocumentCount` - counts the number of non-blank citation columns
* `DocumentAverage` - counts the number of non-blank citation columns

The design is such that new metrics could be added to the `CustomMetricsProcessor._aggregated_metrics` list. If/when we have per-prediction metrics, we will need to add infrastructure for getting the association-id column name from the deployment and iterating over the rows.

TODO: determine what the DocumentCount and DocumentAverage should be counting.

### DIFF
--- a/tests/unit/datarobot_drum/drum/adapters/model_adapters/test_python_model_adapter.py
+++ b/tests/unit/datarobot_drum/drum/adapters/model_adapters/test_python_model_adapter.py
@@ -705,6 +705,7 @@ class TestPythonModelAdapterWithGuards:
 
 
 def fake_metric_body(name: str) -> dict[str, Any]:
+    """Create a fake metric body with a random ID and the given name."""
     return {
         "id": str(random.randint(5000, 100000)),
         "name": name,
@@ -712,7 +713,7 @@ def fake_metric_body(name: str) -> dict[str, Any]:
 
 
 def fake_metric_response(names: list[str]) -> dict[str, Any]:
-    """Create a fake response -- mapping name to a fake body"""
+    """Create a fake response mapping names to a fake metric bodies."""
     return {name: fake_metric_body(name) for name in names}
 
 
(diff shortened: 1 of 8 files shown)


tests/unit/datarobot_drum/drum/adapters/model_adapters/test_python_model_adapter.py
```diff
@@ -701,3 +702,94 @@ def test_guard_chat_wrapper_not_invoked_if_not_defined(self, tmp_path):
             # Even if guard pipeline exists - moderation chat wrapper does not exist, so invoke
             # only user chat method
             assert response.choices[0].message.content == "Hello there"
+
+
+def fake_metric_body(name: str) -> dict[str, Any]:
```
ahmdelemam, line 707:
```suggestion
def fake_metric_body(name: str) -> dict[str, Any]:
    """Create a fake metric body with a random ID and the given name."""
```

ahmdelemam (changes requested):
Looks good, just small improvements.

ahmdelemam (approved):
Thanks

PR #1285, opened 2025-02-19
[AGENT-6788] Prompt could be in the form of list

amarmudrankit:
For chat api, the prompt could be in the form of list.  Make sure to serialize it before reporting for monitoring

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
+++ b/custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
@@ -297,7 +297,25 @@ class BaseLanguagePredictor(DrumClassLabelAdapter, ABC):
         except DRCommonException:
             logger.exception("Failed to report deployment stats")
 
-        latest_message = completion_create_params["messages"][-1]["content"]
+        prompt_content = completion_create_params["messages"][-1]["content"]
+        if isinstance(prompt_content, str):
+            latest_message = completion_create_params["messages"][-1]["content"]
+        elif isinstance(prompt_content, list):
+            concatenated_prompt = []
+            for content in prompt_content:
+                if content["type"] == "text":
+                    message = content["text"]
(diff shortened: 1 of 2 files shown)


custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
```diff
@@ -297,7 +297,25 @@ def _mlops_report_chat_prediction(
...
+                elif content["type"] == "image_url":
+                    message = f"Image URL: {content['image_url']['url']}"
+                elif content["type"] == "input_audio":
+                    message = f"Audio Input, Format: {content['input_audio']['format']}"
+                else:
+                    message = f"Unhandled content type: {content['type']}"
+                concatenated_prompt.append(message)
+            latest_message = "\n".join(concatenated_prompt)
+        else:
+            logger.warning(f"Unhandled prompt type: {type(prompt_content)}")
```
elatt, line 317: Should we also consider sending an MLOps event because this message will only be visible in the model log. This probably requires more thought (as would we spam our platform with events for a poorly configured model?)

elatt (approved):

custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
amarmudrankit: Yeah, it needs more thought.

I think sending event will be little too expensive - because in case of series of wrong prompts - it will fill up events table.  This is model monitoring not able to understand the type and not the model.  So, if model is misconfigured, user would know the error right away.  In case of monitoring - we can try to deduplicate the events and send only 1 event for 1 unhandled type.  I will create an improvement ticket but not sure if we will be able to work on it by 11.0

orgie77 (approved):
LGTM

yakov-g (approved):

PR #1294, opened 2025-02-25
[AGENT-6541 AGENT-6865] Allow user to override parts of nim predictor logic

elatt:

## Summary
- Prepare 1.16.6 changelog
- Adds `served_model_name` runtime param
- No longer always override `model` param from chat completion request
- Add support for customer `score()` hook

## Rationale
For LoRA models and multi-container models we need to be more flexible with our served model name. This allows users to set it via runtime param and also we defer to chat completion to set it and only inject the default to maintain backward compatibility.


elatt:
@alexromanyuk does this help the situation with multi-containers? I understand that if this is the only thing we do, the UX will still be poor but at least it unblocks things at the lowest level right? I still need to update/add tests but wanted to get your opinion on this approach.

alexromanyuk (approved):
👍

PR #1297, opened 2025-02-26
Setup compile dependencies pipeline for fips python3_keras environment in sandbox

G-Despacito:


public_dropin_environments_sandbox/.harness/fips_python3_keras.yaml
NolanJMcCafferty, line 1: please add the pipeline name into this inputSet file name so that we can distinguish between inputSets for different pipelines

NolanJMcCafferty (approved):

public_dropin_environments_sandbox/.harness/fips_python3_keras.yaml
G-Despacito: Discussed offline with @NolanJMcCafferty , we are reusing the same inputSet for different pipelines. So it's fine

klichukb (approved):

PR #1302, opened 2025-02-27
[AGENT-6869 AGENT-6882] Major refactoring of GPU functional tests

elatt:

## Summary
- Bump DRUM 1.16.7
- Fix bug in how we handle when use does not pass `model` param (though it is a required field so this shouldn't ever happen when using OpenAI client).
- Moved GPU tests into `tests/functional/test_inference_gpu_predictors.py`, however this is still part of the main `test_functional_per_framework.sh` script.
- Removed all the temp NIM envs now that we have sidecar
- Update `embedqa` example code to support passing `model` field
- Make sure tests don't run GPU tests in parallel since we typically test on machines with a single GPU
- Large refactoring to add test coverage of NeMoGuard, Reranking and LLM families
- Rebase `nim_sidecar` image to use ChainGuard

## Rationale
We need to scale our functional tests to make it easier to test various NIMs using our sidecar environment. Also, stuffing all the GPU stuff in the one test file was needless so I split it out (appologies for making it harder to review but timelines are tight).

Additionally, with the testing I found a subtle bug with our `model` handling so I added more test cases and fixed the code and prepared things for another release.


custom_model_runner/datarobot_drum/drum/gpu_predictors/base.py
```diff
@@ -116,7 +116,7 @@ def supports_chat(self):
     def _chat(self, completion_create_params, association_id):
         # Force the incoming model name to to match the expected model name because the
         # name isn't very applicable to BYO LLMs.
-        model_name = completion_create_params.pop("model") or self.served_model_name
+        model_name = completion_create_params.pop("model", None) or self.served_model_name
```
elatt, line 119:
```suggestion
        model_name = completion_create_params.get("model") or self.served_model_name
```

custom_model_runner/datarobot_drum/drum/gpu_predictors/base.py
```diff
@@ -116,7 +116,7 @@ def supports_chat(self):
     def _chat(self, completion_create_params, association_id):
         # Force the incoming model name to to match the expected model name because the
         # name isn't very applicable to BYO LLMs.
```
elatt, line 118:
```suggestion
        # Defer to the caller for the `model` name but to maintain our old behavior, also allow the field
        # to be optional and fallback to the configured default name.
```

zohar-mizrahi (approved):

alexromanyuk (approved):

PR #1346, opened 2025-03-12
[AGENT-6917] return error if model has no chat support

ScottOglesby:

## Note
Earlier commits involved a try-catch for a TypeError (where a chat hook was called but its value was None). This was the wrong approach. `supports_chat` is a robust way of determining whether the hook is there, and the model's target type tells us whether it's reasonable to expect it would be there (textgen). (The only way `supports_chat` didn't work well was in a unit test where mocks created an unrealistic condition.)

Based on review comments, the original HTTP 422 response (your request is bad) is replaced with HTTP 404 (chat not available).

## Summary
Return HTTP 404 if `chat/completions` is requested but the PredictionServer encounters an error. Return a specific message if the model does not define a `chat` hook.

## Rationale
When you run a chat verification custom model check, and there's any error at all, you see only:
```
-------- Chat Verification Check --------
Failed
```

It can be useful to let the user know explicitly if the model does not define `chat()`.

## Changes
* `do_chat` now logs an error and returns HTTP 404 if `chat/completions` is requested but there is no chat hook.
* - if the model is textgen, the error mentions `chat` not being implemented (missing code in custom.py, or gap in expectations between model author and user)
* - otherwise, the error mentions `chat` not being supported (wrong model type)
* added fixtures and unit tests for both cases

## Testing
* reran unit tests
* made heavy use of `drum server` locally (with `datarobot-drum` installed in editable mode); with various requests using `curl`

## Notes
* I did not make similar changes in the case of `score()` being unimplemented. Reason: drum does this well at model load time:

```
datarobot_drum.drum.exceptions.DrumCommonException:

Running environment language: Python.
Could not find any framework to handle loaded model and a score hook is not provided
```


custom_model_runner/datarobot_drum/drum/root_predictors/predict_mixin.py
```diff
@@ -368,19 +368,42 @@ def do_predict_unstructured(self, logger=None):
...
+            "This model's chat interface was called, but chat() is not implemented."
+        )
+        # _predictor is a BaseLanguagePredictor attribute of PredictionServer;
+        # see PredictionServer.__init__()
+        if not self._predictor.supports_chat():
+            if logger is not None:
+                logger.error(unsupported_chat_message)
+            return (
+                {"message": "ERROR: " + unsupported_chat_message},
+                HTTP_422_UNPROCESSABLE_ENTITY,
```
amarmudrankit, line 384: I think 404 is more appropriate than 422, because its about interface "not found" than any issue in the input request.

custom_model_runner/datarobot_drum/drum/root_predictors/predict_mixin.py
```diff
@@ -368,19 +368,42 @@ def do_predict_unstructured(self, logger=None):
...
-        result = self._predictor.chat(completion_create_params)
-        if not is_streaming_response(result):
-            response = result.to_dict()
-        else:
-            response = Response(
-                stream_with_context(PredictMixin._stream_openai_chunks(result)),
-                mimetype="text/event-stream",
+            return response, HTTP_200_OK
+        except TypeError:  # if chat() is not defined, hook evaluates to None
+            if logger is not None:
```
amarmudrankit, line 400: nit: can we move this into a method? So many `if`s just to check if logger is defined or not makes code unreadable.  Something like `def log_if_possible()` :)

amarmudrankit (commented):
I think we should return 404, but will wait to see what others have to say.

Regarding testing it - you can do following TEMPORARY steps
1. From this branch, you can build your own DRUM wheel file
2. Choose [buzok Exec env](https://github.com/datarobot/buzok/tree/main/worker/execution_environments/python312_genai) - copy wheel in the image.
3. Modify dockerfile to install this DRUM wheel in the environment
4. Upload this env on staging and then create custom model using this env and test it

mjnitz02 (approved):
The mixin itself is ok for Predictive AI. I think Amar makes some good points that refactoring could improve readability. It could also probably allow for more targetted unit tests of the `do_chat` function.

custom_model_runner/datarobot_drum/drum/root_predictors/predict_mixin.py
```diff
@@ -368,19 +368,42 @@ def do_predict_unstructured(self, logger=None):
...
 
-        result = self._predictor.chat(completion_create_params)
-        if not is_streaming_response(result):
-            response = result.to_dict()
-        else:
-            response = Response(
-                stream_with_context(PredictMixin._stream_openai_chunks(result)),
-                mimetype="text/event-stream",
+            return response, HTTP_200_OK
+        except TypeError:  # if chat() is not defined, hook evaluates to None
```
elatt, line 399: Type error feels like a petty generic exception to be catching to indicate that there is no chat implantation? Especially for user defined chat hooks, I could see them raising TypeErrors for other code issues and us then incorrectly assuming this means there is no implementation and hiding the real error. Aren't there other ways to determine if a specific hook had been implemented

custom_model_runner/datarobot_drum/drum/root_predictors/predict_mixin.py
klichukb: I agree, 400/422 is kind of actionable (go fix your request since its a bad request), but this one is not actionable for the caller.

custom_model_runner/datarobot_drum/drum/root_predictors/predict_mixin.py
klichukb: If we need to define an event that something is not implemented, we should use NotImplementedError and then catch that

klichukb: To Erik's point I would not use TypeError for this, its too generic

ScottOglesby: The new commit has a better understanding of drum internals and bypasses the need to read the tea leaves in a `TypeError`.

custom_model_runner/datarobot_drum/drum/root_predictors/predict_mixin.py
```diff
@@ -368,6 +375,25 @@ def do_predict_unstructured(self, logger=None):
         return response, response_status
 
     def do_chat(self, logger=None):
+        unsupported_chat_message = (
+            "This model's chat interface was called, but chat is not supported."
```
klichukb, line 379: Do you know how we react to this on the DataRobot side? Ultimately this ticket should mean handling things e2e.
I think wed'd need to make sure we actually handle such 404 and propagate back to the user.

The thing is that this message is English and is built here specifically for DRUM, we won't be able to propagate it back to the user as is as it is not translated. This means on the DR side we will need to compose such message again based on 404 response and return a gettext() version of it.

klichukb: I think for now this comment doesnt change whats delivered in this PR but I'm emphasizing that we need to handle the DR side properly

custom_model_runner/datarobot_drum/drum/root_predictors/predict_mixin.py
```diff
@@ -368,6 +375,25 @@ def do_predict_unstructured(self, logger=None):
...
+        )
+        undefined_chat_message = (
+            "This model's chat interface was called, but chat() is not implemented."
+        )
+        # _predictor is a BaseLanguagePredictor attribute of PredictionServer;
+        # see PredictionServer.__init__()
+        if not self._predictor.supports_chat():
+            if self._target_type == TargetType.TEXT_GENERATION:
```
klichukb, line 387: I see this has already 2 different messages. Which already makes DR side a bit problematic. It might be that DR side will have only 1 version of this message at least for 11.0

I also don't fully understand the difference between the two.

custom_model_runner/datarobot_drum/drum/root_predictors/predict_mixin.py
ScottOglesby: The chat prediction handler in custom model logs the error and returns an empty completions list, which is not ideal: https://github.com/datarobot/DataRobot/blob/ddce7d224b8b1f503e5d168ace7333c3f732230e/raptor_lake/custom_model/secure_helpers/prediction_handlers.py#L940

amarmudrankit: /chat/completions is served through prediction gateway - so in case of a deployment prediction, the response of 404 will be relayed back to user.

klichukb: Probably then need to check with Docs team if we can return an untranslated string.

ScottOglesby: I asked Nick A. about this. Drum and moderations library do not translate any strings; we might need to standardize and then translate in `predictions_gateway`.

klichukb: Okay, if you communicate this - fine.

custom_model_runner/datarobot_drum/drum/root_predictors/predict_mixin.py
klichukb: Since we just pass through right now and we have communicated this to Docs team, I guess its fine, but long term needs a solution. @yakov-g  used to suggest error codes

ScottOglesby:
@klichukb Sorry, I mistakenly clicked re-review for you. Was looking for final comments from Erik and Amar before merging. No changes since yesterday.

elatt (approved):

amarmudrankit (approved):

PR #1347, opened 2025-03-14
[RAPTOR-12084] Use 'exec' in Dockerfile ENTRYPOINT for better signal handling and efficiency

zohar-mizrahi:
## Summary
Currently, the `start_server.sh` starts as a child process of `/bin/sh`, meaning:
- The shell /bin/sh remains in memory.
- Signals (e.g., SIGTERM, SIGINT) may not reach start_server.sh properly.

It is desired to replace the parent shell process, using the `exec`, which means:
- The parent shell will not stay in memory.
- The start_server.sh and later drum becomes PID 1 inside the container.
- Signals (e.g., SIGTERM, SIGINT) will be handled properly.

### DIFF
--- a/public_dropin_environments/java_codegen/Dockerfile
+++ b/public_dropin_environments/java_codegen/Dockerfile
@@ -61,4 +61,4 @@ ENV WITH_ERROR_SERVER=1 \
 COPY ./*.sh ${CODE_DIR}/
 WORKDIR ${CODE_DIR}
 
-ENTRYPOINT ["sh", "-c", "${CODE_DIR}/start_server.sh \"$@\"", "--"]
+ENTRYPOINT ["sh", "-c", "exec ${CODE_DIR}/start_server.sh \"$@\"", "--"]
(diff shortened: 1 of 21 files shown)


yakov-g (approved):

klichukb (approved):
I guess the change is fine but did this become necessary with the move to chainguard?

shopin32 (approved):

tsdaemon (approved):
I wonder if we need to do the same with the exec env we are using for Buzok. We have `ENTRYPOINT ["/opt/code/start_server.sh"]`

yakavaliou (approved):

zohar-mizrahi:
> I guess the change is fine but did this become necessary with the move to chainguard?

No, it has nothing to do with chainguard. 
The only reason for using this pattern is to interpret the environment variable `${CODE_DIR}`. It cannot be interpreted during the docker building. Alternatively, I could have written the path hard-coded, instead of relaying on the `${CODE_DIR}`. In which case, I would use the `/opt/code/start_server.sh` directly in the `ENTRYPOINT`.

PR #1349, opened 2025-03-17
[BUZZOK-25116] added new env

mkavf:


public_dropin_gpu_environments/python312_genai/install_dependencies.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+set -e
+
+ARTIFACTORY_MAVEN_URL="https://artifactory.devinfra.drdev.io/artifactory/maven-central/com/datarobot"
```
yakov-g, line 4: May we don't have artifactory references in this env?
1. Environments in this repo are considered to be public and buildable outside of datarobot.
2. We have branch cut release job that triggers HHI based pipeline to build a release tarball. It doesn't have access to the artifactory and will fail.

mkavf: Ok, I will look into that, thanks

PR #1351, opened 2025-03-18
AGENT-6982 Fix DRUM served_model_name

alexromanyuk:

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/gpu_predictors/base.py
+++ b/custom_model_runner/datarobot_drum/drum/gpu_predictors/base.py
@@ -110,42 +110,21 @@ class BaseOpenAiGpuPredictor(BaseLanguagePredictor):
 
     @cached_property
     def served_model_name(self):
-        """
-        There are two ways to specify the model name:
-        - Explicitly set in the request: Required by the /chat/completions API.
-        - Not set in the request: In this case, we ensure backward compatibility with the /predictions API.
-
-        Next, there are two ways to run the model:
-        - Old way: Single-container deployment, where both DRUM and the OpenAI server run in the same container.
-        - New way: Multi-container deployment, where DRUM runs as a sidecar alongside the OpenAI server.
-
-        In the new approach, the OpenAI server starts separately. Therefore, we must either: set NIM_SERVED_MODEL_NAME
-        in the second container, or retrieve the model name from the API.
-        """
-        served_model_name = self.get_optional_parameter(
-            "served_model_name", self.DEFAULT_MODEL_NAME
-        )
+        """Return model name from runtime parameters or fetch from OpenAI server."""
+        default_name = self.get_optional_parameter("served_model_name", self.DEFAULT_MODEL_NAME)
 
         if not self.is_sidecar:
-            return served_model_name
+            return default_name
 
-        model_ids = self._get_deployed_model_ids()
-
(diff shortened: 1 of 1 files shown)


custom_model_runner/datarobot_drum/drum/gpu_predictors/base.py
```diff
@@ -392,18 +371,24 @@ def health_check(self) -> typing.Tuple[dict, int]:
...
                 "message": f"{self.NAME} server is not ready: {str(err)}"
             }, http_codes.SERVICE_UNAVAILABLE
 
+    def _reset_cached_model_name(self):
+        """
+        Resets the cached model name to ensure it's correctly re-fetched from the OpenAI server.
+        This prevents caching incorrect values if the model name was requested before the server was ready.
+        """
+        if hasattr(self, "served_model_name"):
+            del self.served_model_name
```
elatt, line 390: I don't think this is the right logic? When I looked at the source for `cached_property` it is basically the following (at least in my version of Python)
```python
class cached_property:
    def __init__(self, func):
        self.func = func
        self.attrname = None
        self.__doc__ = func.__doc__
        self.lock = RLock()

    def __set_name__(self, owner, name):
        if self.attrname is None:
            self.attrname = name
        elif name != self.attrname:
            raise TypeError(
                "Cannot assign the same cached_property to two different names "
                f"({self.attrname!r} and {name!r})."
            )

    def __get__(self, instance, owner=None):
        if instance is None:
            return self
        if self.attrname is None:
            raise TypeError(
                "Cannot use cached_property instance without calling __set_name__ on it.")
        try:
            cache = instance.__dict__
        except AttributeError:  # not all objects have __dict__ (e.g. class defines slots)
            msg = (
                f"No '__dict__' attribute on {type(instance).__name__!r} "
                f"instance to cache {self.attrname!r} property."
            )
            raise TypeError(msg) from None
        val = cache.get(self.attrname, _NOT_FOUND)
```

So I think the thing you want to delete is actually
```suggestion
    def _reset_cached_model_name(self):
        """
        Resets the cached model name to ensure it's correctly re-fetched from the OpenAI server.
        This prevents caching incorrect values if the model name was requested before the server was ready.
        """
        if "served_model_name" in self.__dict__:
            del self.__dict__["served_model_name"]
```

elatt (approved):

public_dropin_nim_environments/nim_sidecar/env_info.json
```diff
@@ -1,10 +1,10 @@
 {
   "id": "67a6236b8206bdd08ef065cc",
-  "name": "[GenAI][NVIDIA] Generic image used to proxy requests to NIM",
+  "name": "DRUM NIM Sidecar",
   "description": "",
   "programmingLanguage": "python",
-  "environmentVersionId": "67bf9d31b9eb89421a6b665a",
-  "environmentVersionDescription": "Fix handling of 'model' parameter in Chat Completion API",
+  "environmentVersionId": "67d94ed3da51e13ed7d1e8a2",
+  "environmentVersionDescription": "",
```
elatt, line 7:
```suggestion
  "environmentVersionDescription": "Fix spurious served_model_name error during DRUM startup",
```

klichukb (approved):

zohar-mizrahi (approved):

PR #1352, opened 2025-03-18
[RAPTOR-12062] Add Harness pipeline support for testing local development drop-in environment Dockerfiles

zohar-mizrahi:
## Summary
Set up Harness input sets and triggers to test the local development images for all drop-in environments. These images are built using the `Dockerfile.local` file located in each drop-in environment folder.


yakov-g (approved):

klichukb (approved):
I guess its OK to start with but I'm not sure we will continue with this approach

PR #1360, opened 2025-03-21
[AGENT-7023] Default to 10 HTTP workers for nim sidecar container

elatt:

## Summary
Set MAX_WORKERS to 10 via env var set in Dockerfile.

## Rationale
This instance of DRUM is primarily an HTTP proxy and we have found 10 HTTP works to be a good default for good performance with NIMs. This method also allows users to still override this default with a runtime param.

### DIFF
--- a/public_dropin_nim_environments/nim_sidecar/Dockerfile
+++ b/public_dropin_nim_environments/nim_sidecar/Dockerfile
@@ -47,4 +47,7 @@ ENV PYTHONUNBUFFERED=1
 WORKDIR ${CODE_DIR}
 COPY . ${CODE_DIR}/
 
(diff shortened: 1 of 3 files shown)


yakov-g (approved):

elatt:
@kijimoshib or @rickwporter-datarobot how does the migrator pickup new changes to this container (and generate a new pre-built image for on-prem distribution)? Especially now that we have branched for 11.0. Do I need to cherry-pick this to the `release/11.0` branch in this repo? Is there any other bumps I need to make in other repos to _refresh_ the migrator all-in-one artifact?

alexromanyuk (approved):

PR #1368, opened 2025-03-28
[AGENT-6674] Add support for extra_body argument in vLLM

elatt:

## Summary
- Prepare release of 1.16.11 DRUM
- Prepare new vLLM env to use new DRUM version
- Add testcase for `extra_body` handling
- Update logic in default `chat()` hook to correctly pass `extra_body` args to the underlying OpenAI server (e.g. vLLM)

## Rationale
With the old code, users would have to nest `extra_body` args to have them correctly passed to vLLM server. This change makes DRUM more transparent and makes the user feel as though they are communicating directly with the underlying OpenAI server.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/gpu_predictors/base.py
+++ b/custom_model_runner/datarobot_drum/drum/gpu_predictors/base.py
@@ -5,6 +5,7 @@ This is proprietary source code of DataRobot, Inc. and its affiliates.
 Released under the terms of DataRobot Tool and Utility Agreement.
 """
 import csv
+import inspect
 import io
 import json
 import logging
@@ -45,6 +46,16 @@ from datarobot_drum.drum.language_predictors.base_language_predictor import (
 )
 from datarobot_drum.drum.root_predictors.drum_server_utils import DrumServerProcess
 from datarobot_drum.drum.server import HTTP_513_DRUM_PIPELINE_ERROR
+
+# OpenAI client isn't a required dependency for DRUM, so we need to check if it's available
+try:
+    from openai import OpenAI
+    from openai.resources.chat.completions import Completions
+
+    COMPLETIONS_CREATE_SIGNATURE = inspect.signature(Completions.create)
+    _HAS_OPENAI = True
+except ImportError:
+    _HAS_OPENAI = False
 
 
 class ChatRoles:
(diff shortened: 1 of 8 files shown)


amarmudrankit:
Just wondering - Is this change required for NIM 11.0 release? I think we would like to plumb this `extra_body` / `extra_headers` parameters for overall `/chat/completions` API (Bolt on governance).  For example, currently there is no way for the user to send `association_id` with chat prompt - with `extra_body` user can do it.

custom_model_runner/datarobot_drum/drum/gpu_predictors/base.py
```diff
@@ -139,7 +147,11 @@ def _chat(self, completion_create_params, association_id):
         if not model_name_from_request or model_name_from_request == self.DEFAULT_MODEL_NAME:
             completion_create_params["model"] = self.served_model_name
 
-        return self.ai_client.chat.completions.create(**completion_create_params)
+        # Separate valid kwargs from extra kwargs
+        valid_params = set(COMPLETIONS_CREATE_SIGNATURE.parameters.keys())
```
peterzdeb, line 151: `COMPLETIONS_CREATE_SIGNATURE` is defined only when _HAS_OPENAI is "True". Does this method need a check for this flag?

peterzdeb (approved):

custom_model_runner/datarobot_drum/drum/gpu_predictors/base.py
elatt: So it's not the most pretty thing but in the object constructor, we have this check so I feel like it is ok in any method to then assume that the library is installed.

elatt:
> Just wondering - Is this change required for NIM 11.0 release?
> 
> I think we would like to plumb this `extra_body` / `extra_headers` parameters for overall `/chat/completions` API (Bolt on governance). For example, currently there is no way for the user to send `association_id` with chat prompt - with `extra_body` user can do it. So, would need rework for the broader use.
> 
> If its not urgent - we probably can hold off - but if its blocker - probably its ok to go ahead with it

@amarmudrankit I think this is still useful on its own and could provide a base to build the functionality that you are speaking of.

elatt:
Jarvis cherry pick 11.0

PR #1371, opened 2025-04-05
[RAPTOR-12205] Fixed vllm environment

klichukb:
Issue 1. Venv was not activated, hence DRUM used wrong python to start vllm.
error: vllm module not found.
Issue 2. Torch used by vllm tries to resolve the current user. We don't create the user 1000. 
Issue 3. Make sure .cache in the HOME dir has user permissions (1000)

Tested e2e on staging with Nemotron:

```
2025-04-05 09:24:01,827 INFO drum.datarobot_drum.drum.gpu_predictors.base:  INFO:     Application startup complete.
2025-04-05 09:24:04,689 INFO drum.datarobot_drum.drum.gpu_predictors.base:  INFO:     127.0.0.1:56634 - "GET /health HTTP/1.1" 200 OK
2025-04-05 09:24:04,689 INFO werkzeug:  10.190.123.104 - - [05/Apr/2025 09:24:04] "GET /67f105064e643d70cfdcfa0a/67f1556812252410c86fb038/ HTTP/1.1" 200 -
2025-04-05 09:24:04,712 INFO werkzeug:  10.190.96.175 - - [05/Apr/2025 09:24:04] "GET /67f105064e643d70cfdcfa0a/67f1556812252410c86fb038/capabilities/ HTTP/1.1" 200 -
2025-04-05 09:24:07,621 INFO drum.datarobot_drum.drum.gpu_predictors.base:  OpenAI Server is ready; switching watchdog to active mode...
```

Prediction snippet:

```
2025-04-05 19:36:28,628 run.py:36 INFO Trying Simple prompt first: "What would it take to colonize Mars?"
2025-04-05 19:36:41,183 _client.py:1025 INFO HTTP Request: POST https://staging.datarobot.com/api/v2/deployments/67f1556212252410c86fb013/chat/completions "HTTP/1.1 200 OK"
Colonizing Mars would require a significant amount of planning, resources, and technological advancement. Here are some key factors that would be essential to establish a sustainable human presence on Mars:
....
```

PR #1377, opened 2025-04-06
PR followups

klichukb:

### DIFF
--- a/public_dropin_gpu_environments/vllm/Dockerfile
+++ b/public_dropin_gpu_environments/vllm/Dockerfile
@@ -17,21 +17,24 @@ ENV HOME=/opt/.home
 RUN groupadd -g 1000 datarobot && \
     useradd -u 1000 -g datarobot -m -d $HOME -s /bin/bash datarobot
 
-USER datarobot
+USER 1000:0
 
-COPY --chown=1000:1000 .gitkeep ${HOME}/.placeholder
-COPY --chown=1000:1000 .gitkeep ${CODE_DIR}/.placeholder
-COPY --chown=1000:1000 .gitkeep ${DATAROBOT_VENV_PATH}/.placeholder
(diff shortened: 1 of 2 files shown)


elatt (approved):

public_dropin_gpu_environments/vllm/Dockerfile
```diff
@@ -17,21 +17,24 @@ ENV HOME=/opt/.home
...
 
-USER datarobot
+USER 1000:0
 
-COPY --chown=1000:1000 .gitkeep ${HOME}/.placeholder
-COPY --chown=1000:1000 .gitkeep ${CODE_DIR}/.placeholder
-COPY --chown=1000:1000 .gitkeep ${DATAROBOT_VENV_PATH}/.placeholder
+COPY --chown=1000:0 .gitkeep ${HOME}/.placeholder
+COPY --chown=1000:0 .gitkeep ${CODE_DIR}/.placeholder
+COPY --chown=1000:0 .gitkeep ${DATAROBOT_VENV_PATH}/.placeholder
```
peterzdeb, line 24: `datarobot:root` should work as well?

peterzdeb (approved):

public_dropin_gpu_environments/vllm/Dockerfile
klichukb: Perhaps, but as Erik mentioned (I think I google this as well) that openshift prefers numeric uids/gids instead of names (at least for the USER command). I'm not sure however if this is relevant for COPY --chown however, so decided to stay on the safe side..
https://github.com/datarobot/datarobot-user-models/pull/1371/files

peterzdeb: Ok, makes sense then

PR #1388, opened 2025-04-16
[MMM-19130] Update moderations initialization to enable VDB pipelines

rickwporter-datarobot:

## Summary

This uses the `moderations` initialization updates from https://github.com/datarobot/moderations/pull/287 to enable running the new moderations VDB pipelines.

## Rationale

The `moderations` library has new initialization functions to help streamline the DRUM initialization such that DRUM does NOT need to know so much about the `moderations` library. This helps streamline the process of including processing for VDB deployments, too.

When the new initialization functions are not present, we initialize the function references and pipeline using the old functions.

## Changes

For VDB models, the `moderations` library adds metrics (and does not [currently] provide guards), so several of the names change from `guard` to `moderation`.

Instead of creating a `PythonModelAdapter._moderations_hooks` map that indexes by a function name, we just store the functions in `_moderation_score_fn` and `_moderation_chat_fn`. The new variables are populated using `moderations.drum_integration.get_moderations_fn()` which returns a callable reference to the function. This eliminates the need for DRUM to know the function names for VDB vs LLM deployments. 

The remainder of the changes are effectively just using the renamed `_moderation_pipeline`, and new `_moderation_score_fn` and `_moderation_chat_fn` variables.


custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
```diff
@@ -107,43 +104,35 @@ def __init__(self, model_dir, target_type=None):
...
-                GUARD_CHAT_WRAPPER_NAME: getattr(guard_module, GUARD_CHAT_WRAPPER_NAME, None),
-            }
-            if self._guard_moderation_hooks[GUARD_INIT_HOOK_NAME] and (
-                self._guard_moderation_hooks[GUARD_SCORE_WRAPPER_NAME]
-                or self._guard_moderation_hooks[GUARD_CHAT_WRAPPER_NAME]
-            ):
-                self._guard_pipeline = self._guard_moderation_hooks[GUARD_INIT_HOOK_NAME]()
-            else:
-                self._logger.debug("No guards defined")
+            self._moderation_score_fn = mod_module.get_moderations_fn(self._target_type, "score")
```
amarmudrankit, line 127: "score" -> CustomHooks.SCORE
"chat" -> CustomHooks.CHAT

amarmudrankit: I will also call them `moderation_*_hook` instead of `fn` - they are really hooks into moderation library

amarmudrankit: You also need to check existence of `create_pipeline`.  It is possible that moderation library is not updated and will fail.  Ensures backward compatibility

amarmudrankit (commented):
Probably missing test for VDB target type as well

amarmudrankit (approved):
LGTM

custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
```diff
@@ -107,43 +104,48 @@ def __init__(self, model_dir, target_type=None):
...
-                self._load_guard_hooks_for_drum()
+            self._load_moderation_hooks()
         else:
             self._target_name = None
 
-    def _load_guard_hooks_for_drum(self):
+    def _load_moderation_hooks(self):
         try:
-            guard_module = __import__(GUARD_HOOK_MODULE, fromlist=[MODERATIONS_LIBRARY_PACKAGE])
+            mod_module = __import__(MODERATIONS_HOOK_MODULE, fromlist=[MODERATIONS_LIBRARY_PACKAGE])
```
yakov-g, line 123: We should start moving to OOP to have additional functionality to be imported as classes.
I suggest start adding(or even completely moving to OOP) such support into moderations.

Probably moderations would be a good candidate for POC as currently it is still relatively small.

yakov-g (approved):

PR #1392, opened 2025-04-18
[MODEL-18996] update custom task select param to allow longer strings

eric-s-s:

This PR is a companion to: https://github.com/datarobot/DataRobot/pull/143873  and will not merge until after that one.

The original limit seems to have been arbitrary, and we got a request for longer limits, so i'm making it 1024 len. 

some context on where to find advanced tuning code is here: https://datarobot.atlassian.net/wiki/spaces/MODEL/pages/1989771284/Advanced+Tuning+Parameters

searching in DR code you should grep for advanced_tuning

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/model_metadata.py
+++ b/custom_model_runner/datarobot_drum/drum/model_metadata.py
@@ -43,7 +43,7 @@ from datarobot_drum.drum.typeschema_validation import (
 PARAM_NAME_MAX_LENGTH = 64
 
 # Max length of a select value
-PARAM_SELECT_VALUE_MAX_LENGTH = 32
+PARAM_SELECT_VALUE_MAX_LENGTH = 1024
 
 # Max number of possible select values
 PARAM_SELECT_NUM_VALUES_MAX_LENGTH = 24
--- a/tests/unit/datarobot_drum/model_metadata/test_model_metadata_hyperparameters.py
+++ b/tests/unit/datarobot_drum/model_metadata/test_model_metadata_hyperparameters.py
@@ -707,7 +707,7 @@ def test_yaml_metadata__select_hyper_param_invalid_value_size(
 
     with open(os.path.join(tmp_path, MODEL_CONFIG_FILENAME), mode="w") as f:
         yaml.dump(model_metadata, f)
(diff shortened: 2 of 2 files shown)


alinawishnivezki (approved):
Lgtm for PredAI

custom_model_runner/datarobot_drum/drum/model_metadata.py
```diff
@@ -43,7 +43,7 @@
 PARAM_NAME_MAX_LENGTH = 64
 
 # Max length of a select value
-PARAM_SELECT_VALUE_MAX_LENGTH = 32
+PARAM_SELECT_VALUE_MAX_LENGTH = 1024
```
eric-s-s, line 46: @klichukb @yakov-g we use this in: `datarobot_drum.drum.model_metadata.read_model_metadata_yaml`  where we validate this in a trafaret in: `datarobot_drum.drum.model_metadata.validate_model_metadata_hyperparameter`  why do we do this?  Isn't this the responsibility of the custom task version ingest?

also, if I want to apply this change, do I then need to bump the drum version and update all of the public drop-in envs?   if so, how do I do this?

yakov-g: Hey, I'm almost completely not familiar with model metadata, its validation and hyper parameters.
It all was developed by Jesse,   (4dhalibut - I forgot the name) and i think you contributed a lot into this.

Regarding release:
1. Create a pr bumping DRUM, merge, release DRUM - there is a pipeline.
2. Create a PR bumping DRUM in the envs. Env versions IDs will be bumped automatically.
Maybe you could do 1 and 2 in a single PR, not sure how tests are configured.
3. If you think envs should be published: build installer and then run it to install envs

https://datarobot.atlassian.net/wiki/spaces/RAPTOR/pages/1416691967/Install+drop-in+environments+on+staging+production+sts

yakov-g (approved):

custom_model_runner/datarobot_drum/drum/model_metadata.py
eric-s-s: @scottp-ml do you know why we have this here?

PR #1393, opened 2025-04-21
[AGENT-7138] Scott/agent 7138 models api in drum

ScottOglesby:

## Summary
Implements `models` and `v1/models` OpenAI API routes for text generation custom models. (ref: https://platform.openai.com/docs/api-reference/models/list)

If the model is not textgen, the response is HTTP 404.
If the model's `custom.py` implements `get_supported_llm_models`, the list from that function is returned.
If `custom.py` does not implement it, but the `LLM_ID` runtime parameter is defined, a 1-item list using that model ID is returned. Playground sets the LLM_ID parameter when sending to custom model workshop.
If neither the `custom.py` function nor the runtime parameter exists, an empty list is returned.

Based on Amar's PR #1389.

## Rationale
Implementing those routes supports OpenAI clients working with custom models, and helps DataRobot workflows using these models.

## Details
The 1-item list using the `LLM_ID` runtime parameter has the following details:

- "id": llm_id,
- "object": "model",
-  "created": 1735689600,  # Jan. 1, 2025; this information is not easily available
- "owned_by": "DataRobot",  # by convention; can be changed

## Changes
* added connections to `get_supported_llm_models` custom.py function
* added `LLM_ID` runtime parameter logic
*  added tests for prediction server and base language predictor
* added fixtures to inject runtime parameter environment variables. These are very convenient for working with RTPs, but they must be cleaned up after each test (since os.environ persists for the test session)
* added a model template with custom.py supporting model list and chat


ScottOglesby:
jarvis run base_checks-test_unit

custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
```diff
@@ -763,6 +766,50 @@ def chat(self, completion_create_params, model, association_id):
...
+        If that hook is not defined, return the model defined in LLM_ID runtime parameter.
+        If neither is present, return an empty list.
+
+        Parameters
+        ----------
+        model
+
+        Returns
+        -------
+
```
peterzdeb, line 784: No return? 
Actually it returns some structure, maybe even worth to prepare a structure type (datatype?) to make it self descriptive

peterzdeb (approved):

amarmudrankit (approved):

PR #1396, opened 2025-04-24
[MMM-19199] Allow building with moderations library 

rickwporter-datarobot:

## Summary

Allow specifying the moderations wheel file in `create-drum-dev-image.sh`. 

## Rationale

For some DRUM developers, we want to use a container with the moderations library included. This allows the user to specify the moderations wheel file, and the `Dockerfile` and `requirements.txt` will get updated automatically.

### DIFF
--- a/tools/create-drum-dev-image.sh
+++ b/tools/create-drum-dev-image.sh
@@ -20,6 +20,13 @@ if [ -z "$1" ]; then
   exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


yakov-g:
Never used this helper script.
I believe you tried and it works for you. :)
Worth adding a confluence page about testing drum with moderations if it is not there yet.

peterzdeb (approved):

PR #1398, opened 2025-04-25
[WIP][PBMP-7171] Offloading of monitoring data in Chat API 

yura-ant:

## Summary
Since mlops support async reporting. I would like to test if we can use it by default to speed response time. 
Any reason it is not by default today?

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
+++ b/custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
@@ -179,6 +179,7 @@ class BaseLanguagePredictor(DrumClassLabelAdapter, ABC):
                 mlops_service_url=self._params["external_webserver_url"],
                 mlops_api_token=self._params["api_token"],
             )
+            self._mlops.set_async_reporting()
 
     def get_prompt_column_name(self):
         if not self._params.get("deployment_id", None):


custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
```diff
@@ -179,6 +179,7 @@ def _configure_mlops(self):
                 mlops_service_url=self._params["external_webserver_url"],
                 mlops_api_token=self._params["api_token"],
             )
+            self._mlops.set_async_reporting()
```
amarmudrankit, line 182: Interesting ! I guess its worth trying.  I haven't tested async reporting with API Spooler.  @DrewParallel have you tried it?

Overall I feel, if it works with API Spooler, it does provide slight improvement over existing state of affairs.  But, there are bigger issues with API spooler:

1. Rate limited: API Spooler does not have retry logic - agent has it
2. Its the code in the container - so probably limited with resources - cpu and memory.  
     a. The underlying worker process is going to contend for the resource with the actual DRUM.  I think the async 
          reporting right away needs 512 MB of memory by default (so you would have to configure that too)
     b. API Spooler is sync - Typical latency to report stats (1+ second) and report predictions (2+ seconds - typically - depends on data) - if there are 10 requests/sec load - it will easily hog all CPUs soon.  
3. No batching - We will quickly get rate limited on number of API calls

DrewParallel: I have not tried it

ahamino: How are we testing this?

yura-ant: @ahamino I've created this PR with enabled async reporting  to check if it brake any test. All tests passes. But it doesn't mean that it is properly 100% covered in those existing tests. Also currently I am trying  in production create custom model and via override start_server.sh  enable async reporting when start drum server as looks like such option is already available. And then run cicada test to check if there will be speed up in response time and no errors. @ahamino @amarmudrankit Do we have any real custom model which require offloading of monitoring data to try enable there asynchronous reporting to check if it speed up response time and no cause any errors. If it works as expected then potentially  no changes are required.

yura-ant: There is  some documentation about --monitoring-settings in https://docs.datarobot.com/en/docs/predictions/port-pred/pps/custom-pps.html#drum-based-environment-variables

yura-ant: @amarmudrankit I've use RAG temple and modify custom model to use async api spooler via env variables bellow which deployed to  https://app.datarobot.com/console-nextgen/deployments/681b7591a69e08ca95f9ece8/activity-log/runtime-logs
 `export MLOPS_ASYNC_REPORTING=True
export MLOPS_SPOOLER_TYPE=“API"
export MLOPS_SERVICE_URL="https://app.datarobot.com"
export MLOPS_API_TOKEN=“Njc0NDY3NDMwMmIyODc3MTRhOTQ4ODYwOkF2cHNFTDcyN0FvUTlxWnpXeVdSaHlGQThaZ0pVVUNqUzgzSTFFeXJ0Zjg9"`.  To me it looks like it works as designed.  Could  also test if it meet your expectations?

amarmudrankit: Glad it worked, but as I said it is a slight improvement over existing infrastructure, because API spooler does not handle 429s - requirement for memory and CPU resource etc.  I think it will be in @ahamino 's purview if it meets the expectation or not.

yura-ant: @amarmudrankit  Do we have a tests which show that we meet all requirements if we turn off completely  monitoring data?  I think it mandatory requirement to have such tests to complete main story  https://datarobot.atlassian.net/browse/PBMP-7021  or not ?

ahamino: What's the latency like after this change? We say it is "improved" but by how much? 
Can I get a before and after? Does this actually resolve the problem or is it a bandaid?

I'm ok with shifting this to be an improvement in the spooler, but someone has to do it before we call 7021 done.

yura-ant: There is my test result  (which show that response is returned and after that we do report if async report is used )
reporting took ~ 1 second 
 
**no async reporting**
```
2025-05-08 13:54:37,277 DEBUG urllib3.connectionpool:  Starting new HTTPS connection (1): app.datarobot.com:443
2025-05-08 13:54:37,439 DEBUG urllib3.connectionpool:  https://app.datarobot.com:443 "POST /api/v2/deployments/681cb3cf8fca0898b2214dda/predictionRequests/fromJSON/ HTTP/1.1" 202 17
2025-05-08 13:54:37,439 DEBUG datarobot_mlops.common.api_sync:  Successfully sent request to https://app.datarobot.com/api/v2/deployments/681cb3cf8fca0898b2214dda/predictionRequests/fromJSON/
2025-05-08 13:54:37,441 DEBUG urllib3.connectionpool:  Starting new HTTPS connection (1): app.datarobot.com:443
2025-05-08 13:54:38,132 DEBUG urllib3.connectionpool:  https://app.datarobot.com:443 "POST /api/v2/deployments/681cb3cf8fca0898b2214dda/predictionInputs/fromJSON/ HTTP/1.1" 202 17
2025-05-08 13:54:38,132 DEBUG datarobot_mlops.common.api_sync:  Successfully sent request to https://app.datarobot.com/api/v2/deployments/681cb3cf8fca0898b2214dda/predictionInputs/fromJSON/

**2025-05-08 13:54:38,133 INFO werkzeug:  10.189.106.234 - - [08/May/2025 13:54:38] "POST /6819e0637aa266d04ab2cbc8/681cb3d38fca0898b2214e07/chat/completions HTTP/1.1" 200 -**
```

vs

**async reporting**

```
**2025-05-08 14:01:29,927 INFO werkzeug:  10.189.106.234 - - [08/May/2025 14:01:29] "POST /6819e0637aa266d04ab2cbc8/681b7594a69e08ca95f9ed11/chat/completions HTTP/1.1" 200 -**

2025-05-08 14:01:29,929 DEBUG urllib3.connectionpool:  Starting new HTTPS connection (1): app.datarobot.com:443
2025-05-08 14:01:30,756 DEBUG urllib3.connectionpool:  https://app.datarobot.com:443 "POST /api/v2/deployments/681b7591a69e08ca95f9ece8/predictionRequests/fromJSON/ HTTP/1.1" 202 17
2025-05-08 14:01:30,757 DEBUG datarobot_mlops.common.api_sync:  Successfully sent request to https://app.datarobot.com/api/v2/deployments/681b7591a69e08ca95f9ece8/predictionRequests/fromJSON/
2025-05-08 14:01:30,758 DEBUG urllib3.connectionpool:  Starting new HTTPS connection (1): app.datarobot.com:443
2025-05-08 14:01:31,823 DEBUG urllib3.connectionpool:  https://app.datarobot.com:443 "POST /api/v2/deployments/681b7591a69e08ca95f9ece8/predictionInputs/fromJSON/ HTTP/1.1" 202 17
2025-05-08 14:01:31,823 DEBUG datarobot_mlops.common.api_sync:  Successfully sent request to https://app.datarobot.com/api/v2/deployments/681b7591a69e08ca95f9ece8/predictionInputs/fromJSON/
```

ahamino: Am I reading this correctly, latency:

async: 827 ms 
async: 162 ms

aka 80% improvement in reduction of latency between?  Love it! cc @dr-nate-daly-pm @CatherineBo 

Can we do more than 1 request test .. Does the latency change under load? calculate p50, p95? 

What's did @amarmudrankit  mean by "API spooler does not handle 429s  - requirement for memory and CPU resource etc." in this context?

amarmudrankit: > What's did @amarmudrankit mean by "API spooler does not handle 429s - requirement for memory and CPU resource etc." in this context?

So, the API request to report monitoring data can fail with 429 (Rate limiting) - in which case - monitoring agent retries the request after specified number of seconds.  In the case of API Spooler, there is no retry mechanism.  If the request fails - for any reason - monitoring data for that request will not be reported (there is no retry mechanism in API Spooler)

Async reporting internally uses `multithreading` library - so a separate process running in the same container (as drum) and needs at least 512 MB memory.  So, it is going to compete for resources with DRUM.  Bigger problem if the request /sec starts increasing.

In steady state and happy path - this works beautifully, but I suspect there could be problems down the road.

yura-ant:
There is my test results for /chat/completions  with test payload  `--data '{"model": "datarobot-deployed-llm", "messages": [{"role": "system", "content": "You are a helpful assistant"}, {"role": "user", "content": "Hi!"}]} ` for to Guarded RAG assistant TextGeneration custom model which was created by our https://github.com/datarobot-community/guarded-rag-assistant.  One deployment use sync reporting (it is default one today). Another one use async reporting (enable it via env values). Based on results for 1 user or 5 users async reporting is faster.
Screenshot 2025-05-09 at 14 04 05
Screenshot 2025-05-09 at 13 57 20

custom_model_runner/datarobot_drum/drum/language_predictors/base_language_predictor.py
ahamino: Lets make sure the new min memory requirements are very visible to everyone

amarmudrankit:
Reviewing again with @ahamino perspective:
1. This needs to be worker safe .. like works with custom_model_workers - I think it will be worker safe - if we fork the processes as workers ( I think we do fork process and not threads).  However, each worker => new MLOps instance => 512 MB additional memory.  So, for example, 4 workers => 512 MB * 4 = 2GB memory requirement
2. Handling exceptions: Unfortunately - the API spooler [will raise exception up ](https://github.com/datarobot/tracking-agent/blob/master/mlops-py/datarobot_mlops/common/api_sync.py#L86) [to the async channel](https://github.com/datarobot/tracking-agent/blob/master/mlops-py/datarobot_mlops/channel/output_channel_queue.py#L517) and will likely [to kill the worker](https://github.com/datarobot/tracking-agent/blob/master/mlops-py/datarobot_mlops/channel/output_channel_queue.py#L399).  **_I feel this sounds more risky - because then there is no worker to process the async records. A single 429 or 503 could lead to whole monitoring process to halt_**. So, we will need changes in the MLOps library to handle the exception.  Note that exception cannot be handled in DRUM because of async reporting nature.

ahamino:
> 2. Handling exceptions: Unfortunately - the API spooler [will raise exception up ](https://github.com/datarobot/tracking-agent/blob/master/mlops-py/datarobot_mlops/common/api_sync.py#L86) [to the async channel](https://github.com/datarobot/tracking-agent/blob/master/mlops-py/datarobot_mlops/channel/output_channel_queue.py#L517) and will likely [to kill the worker](https://github.com/datarobot/tracking-agent/blob/master/mlops-py/datarobot_mlops/channel/output_channel_queue.py#L399).  **_I feel this sounds more risky - because then there is no worker to process the async records. A single 429 or 503 could lead to whole monitoring process to halt_**. So, we will need changes in the MLOps library to handle the exception.  Note that exception cannot be handled in DRUM because of async reporting nature.

Yea, please handle the exception .. the code should put error in the logs, but shouldn't kill the monitoring forever.

yura-ant:
@amarmudrankit thanks for those concerns. 
1. I will test if it works with custom_model_workers and asyncspooler on production via cicada test
2. Sounds like  we would like to improve existing asynchronous MLops to behave differently when it may stop working. This looks like a separate story to me. But I would like to share my thoughts  any approach may fail in some cases. For example if drum stop works cause sync spooler raise out of memory it can't fix yourself. Do you know what we do today for such use case ? I think if reporting stop works it could be restarted together with drum if such mechanism exist
Async spooler was developed by DataRobot in 2020 https://datarobot.atlassian.net/browse/AGENT-868  and was improved few times sounds like it used on production and supported please correct me if I am wrong. I can see you contribute a lot to tracking-agent project and it is huge benefits we can work on it together .@amarmudrankit @ahamino  Can we create a story with requirements for owners of tracking-agent to improve  asyncspooler in MLopls if current functionality not satisfy our needs but looks right way to continue. I can also try to work on it if we do not find any other candidate with right skills as I never contribute to this project before.

yura-ant:
It looks like it works with  async spooler and  5 workers  https://app.datarobot.com/console-nextgen/deployments/682dea1bfceb315f4c18242e/activity-log/runtime-logs
There is cicada result
Screenshot 2025-05-21 at 18 29 23

I've seen situation when monitoring stop works due to error bellow. Once I enabled it and restart it works well then.

```
raise DRMLOpsConnectedException(f"Request error for {url}: {e} {emsg}")
datarobot_mlops.common.connected_exception.DRMLOpsConnectedException: (('Request error for https://app.datarobot.com/api/v2/deployments/682dea1bfceb315f4c18242e/predictionInputs/fromJSON/: 422 Client Error: UNPROCESSABLE ENTITY for url: https://app.datarobot.com/api/v2/deployments/682dea1bfceb315f4c18242e/predictionInputs/fromJSON/ {"message": "Index 0: Feature Drift tracking and predictions data collection are disabled. Enable feature drift tracking or predictions data collection from deployment settings first, to post features"}',), {})

```

yakov-g:
> set_async_reporting

Unfortunately I can not.

Was async spooling already there in January in the published mlops version?
I think there were no specific intention to use sync and not to use async.

amarmudrankit:
Just curious - how much RAM did you allocate for model when running with 5 workers.

amarmudrankit:
Regarding improving MLOps library:
1. This is where I initially suggested - we need to run monitoring agent for the prediction gateway - It solves all the problems - lesser memory requirements - graceful handling of exception - no monitoring data loss etc 
2. Use of API spooler was a hack - All the changes - async reporting -> handling exceptions will be hacks - it will not get you to the production grade code.  API spooler was intended to be like demo spooler where CFDS could demo the monitoring functionality without running the Java agent process.  Not only we have to find resource to harden the API spooler but also maintain it later.  It will be re-inventing wheel that was already done with monitoring agent.
3.  Instead - setting up monitoring agent is an infrastructure work - not trivial - but less complicated to achieve and maintain later.

cc: @ahamino @chrj

chrj:
> * This is where I initially suggested - we need to run monitoring agent for the prediction gateway - It solves all the problems - lesser memory requirements - graceful handling of exception - no monitoring data loss etc

It's not obvious to me that this is the best option. That would require us to redundantly parse requests and responses adding extra overhead.

yura-ant:
> Just curious - how much RAM did you allocate for model when running with 5 workers.

I didn't allocate more memory it is default today 1GB RAM

yura-ant:
> Regarding improving MLOps library:
> 
> 1. This is where I initially suggested - we need to run monitoring agent for the prediction gateway - It solves all the problems - lesser memory requirements - graceful handling of exception - no monitoring data loss etc
> 2. Use of API spooler was a hack - All the changes - async reporting -> handling exceptions will be hacks - it will not get you to the production grade code.  API spooler was intended to be like demo spooler where CFDS could demo the monitoring functionality without running the Java agent process.  Not only we have to find resource to harden the API spooler but also maintain it later.  It will be re-inventing wheel that was already done with monitoring agent.
> 3. Instead - setting up monitoring agent is an infrastructure work - not trivial - but less complicated to achieve and maintain later.

Today API spooler is on Production as part of FedRAMP Compliance work https://datarobot.atlassian.net/wiki/spaces/ERP/pages/6496944848/DONE+PBMP-6973+Custom+Models+Drop-In+Environments+CVE+Remediation+for+FedRAMP+Compliance+Sign-off+document

Also If API spooler is a hack I see no reason to keep it (5 years we advertise to use this spooler). It should be removed from lib and docs or as first step marked as deprecated at least to avoid situation when people spend their time on it

In my example I showed that customer could also enable asynchronous functionality of API spooler on production

If we feel it is not production grade then I would advise  to create a critical live issue to fix it as soon as possible and not hide this problem unit real customers start complains

ikalnytskyi (approved):
I agree that API spooler usage has its own concerns, but it's something that's already used in DRUM. Switching it into an async mode (running actual API submission from a background process) shouldn't make things much worse. Memory consumption may grow a bit, but it's not like existing models are going to be implicitly affected by this.

I suggest we merge it, if @yura-ant has tested this and is confident that it works.

amarmudrankit:
> That would require us to redundantly parse requests and responses adding extra overhead.
Not sure which parsing you are referring to?  The library will write to a spool (say RabbitMQ) and the agent will read and parse it and post it.  Though all the agent work is happening in the background.

yura-ant:
> @yura-ant API spooler did not exist for 5 years. It was built within last 2 years. I think you meant async spooler.

looks like from 2020

`
Screenshot 2025-05-22 at 21 49 09
`

amarmudrankit:
> > @yura-ant API spooler did not exist for 5 years. It was built within last 2 years. I think you meant async spooler.
> 
> looks like from 2020 maybe it to was released later Screenshot 2025-05-22 at 21 49 09

**_AGENT-4850 Create API Spooler Skeleton  25th Aug 2023._** Thats when it really started.  I think the other history is not relevant to API spooler.

yura-ant:
> When dogfooding on our own systems (/chat/completions using monitoring library) we should not make this choice, especially when we know the details of it. If the customer does make this choice - we can educate them to not to do it and help them make the right choice. But, we cannot justify the choice just because it is already released. I thought that was the purpose of this PBMP, API spooler was a unfortunately a bad choice and we should rectify it.

@chrj @ahamino  I have no plane to fix "a bad choice api spooler" in scope of this story. Story has one requirement offloading monitoring data in Chat API. And this is possible without any changes today. Also if we want  make it by default then we can merge this PR  it will not effect existing model only new models (All new models require testing to be published to prod. If they had any issues they could switch to sync api spooler the same way as they can switch to async api spooler today). I personally prefer do not change default behaviour right now. As first step I recommend to satisfy needs at least one, two customers with existing solution potentially it may prove it was not so bad choice api spooler.

yura-ant:
> > Switching it into an async mode (running actual API submission from a background process) shouldn't make things much worse
> 
> I think it could - we should test how it behaves with 429s and errors etc especially with async reporting.

@amarmudrankit  @ikalnytskyi  Unfortunately there is no way to satisfy all requirements with one approach. (that is why we provide multiple variants to report stats isn't it?) The final acceptance testing should be made by person who deploy custom model to production. Switch to something new by default will lead to existing limitations of this new default solution.

amarmudrankit:
Can we test how async reporting behaves in case of 429 or 5XX? If the async reporting could not recover from those errors, then we rather should not merge this change.  The current sync way of reporting data does will also fail of course - but say for 429 case - it will recover and starts posting again successfully after 1 hour.

yura-ant:
@amarmudrankit Unfortunately I wasn't able to test  it locally as async option looks like not supported on Mac OS M1 processor when run test (
Screenshot 2025-05-26 at 18 57 52

). 
I think you are right asynchronous option not recover from those errors today and require a full restart (I can reproduce it when disable Enable prediction row storage on production). But potentially it could be fixed if we stop raising those errors and just logs them for async option. As first step I try to disable them via patch ApiSync class in https://app.datarobot.com/registry/custom-model-workshop/6819e0637aa266d04ab2cbc8/assemble via drum-custom file. Currently I can't figure out why `Traceback (most recent call last):
  File "/opt/code/./drum-custom", line 7, in <module>
    import requests
ModuleNotFoundError: No module named 'requests'` I know we did such testing in the past. Any ideas are welcome.
This is just to validate that  api spooler with async option could be resistance to them if we stop raise them. Proper fix will require more changes I think.


if we make async api spooler continue works after response with code 429 or 5XX it will satisfy our needs? cause any way that separate report process could be stopped due to other unexpected errors and then it will stop reporting and will require restart model.

yura-ant:
There is https://github.com/datarobot/tracking-agent/pull/3228 which make async api spooler continue works after failed  responses like 429 or 5XX

PR #1401, opened 2025-04-25
[MMM-19325] Use new object-oriented moderations pipeline

rickwporter-datarobot:

## Summary

This consumes the changes in https://github.com/datarobot/moderations/pull/295 that make interfacing with the moderations library easier.

## Rationale

Use object-oriented `ModerationsPipeline` that has `score()` and `chat()` functions instead of tracking a different pipeline class and functions separately. This should make maintenance of the moderations library easier, particularly after we can remove the interfaces to the old moderations libraries.

## Notes

Much of the change in the testing was due to issues encountered during test. The "last local module" was still in memory, so it was causing problems in other tests. The context is done so that the files are removed and the "local module" is unloaded following the test execution -- this is done to avoid bleeding that content between tests.


amarmudrankit (approved):
LGTM

mbromanowsky (approved):
LGTM for PredAI

PR #1404, opened 2025-04-28
[MMM-19225] DRUM changes for agentic workflow target type

automactic:


tests/functional/test_inference_per_framework.py
```diff
@@ -314,6 +317,7 @@ def test_custom_models_with_drum_capitalize_artifact_extensions(
             (PYTHON_TEXT_GENERATION, TEXT_GENERATION, PYTHON_TEXT_GENERATION, None),
             (PYTHON_GEO_POINT, GEO_POINT, PYTHON_GEO_POINT, None),
             (PYTHON_VECTOR_DATABASE, VECTOR_DATABASE, PYTHON_VECTOR_DATABASE, None),
+            (PYTHON_AGENTIC_WORKFLOW, AGENTIC_WORKFLOW, PYTHON_AGENTIC_WORKFLOW, None),
```
yakov-g, line 320: Inference per framework tests are quite weird...
It works in the following way:
This test module is run in in every environment container (framework env)  in https://github.com/datarobot/datarobot-user-models/tree/master/public_dropin_environments. (technically, not for every, but for each we manually defined)
When tests are started we pass an argument --framework-env.
In the beginning of the test we check if `framework` from parametrization belongs to a specific environment(framework),
```
 skip_if_framework_not_in_env(framework, framework_env)
 ```
 if yes, we run test case in that specific env.

So you need to add: `PYTHON_AGENTIC_WORKFLOW` https://github.com/datarobot/datarobot-user-models/blob/9e50a3e554ecd0a5c2676d88800da1307dc7defc/tests/conftest.py#L200

Now I'm going to check if we even trigger any tests in PYTHON311_GENAI env.

eric-s-s (approved):

tests/functional/test_inference_per_framework.py
yakov-g: Looks like it all good then.

automactic: @yakov-g Would you mind approving this PR? It needs two approvals to merge. 😄

automactic: @yakov-g Hi, could you please approve my PR if everything looks good? Thanks!

yakov-g (approved):

PR #1405, opened 2025-04-29
[RAPTOR-12324] Using logging in DRUM server outputs

peterzdeb:

## Summary

Timestamps are already provided in the log format from logging module. Problem is that DRUM is a cli script and when running locally it provides outputs to the console.
There were many places with `print()` outputs on server side. That works as a bad practice to follow in custom model scripts - only logging module can be in use there. 

This PR only replaces `print` in DRUM modules. All the usage of `print()` actually happens in the `model_templates`
In a followup PR I'm planning to update `model_templates`.

## Rationale
Timestamps are needed in all logs from DRUM server. The Runtime logs and events on UI are collected from the containers running this server. Without proper and consistent log formatting it won't be possible to track what is happening with requests to DRUM server and when.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/adapters/cli/shared/drum_class_label_adapter.py
+++ b/custom_model_runner/datarobot_drum/drum/adapters/cli/shared/drum_class_label_adapter.py
@@ -4,16 +4,13 @@ All rights reserved.
 This is proprietary source code of DataRobot, Inc. and its affiliates.
 Released under the terms of DataRobot Tool and Utility Agreement.
 """
-import logging
-import sys
-
 import numpy as np
 import pandas as pd
-from datarobot_drum.drum.enum import LOGGER_NAME_PREFIX
+from datarobot_drum.drum.common import get_drum_logger
 from datarobot_drum.drum.enum import TargetType
 from datarobot_drum.drum.exceptions import DrumCommonException
 
-logger = logging.getLogger(LOGGER_NAME_PREFIX + "." + __name__)
+logger = get_drum_logger(__name__)
 
 
 def needs_class_labels(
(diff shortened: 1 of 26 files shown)


public_dropin_environments/r_lang/env_info.json
```diff
@@ -3,7 +3,7 @@
   "name": "[DataRobot] R 4.4.3 Drop-In",
   "description": "This template environment can be used to create artifact-only R custom models that use the caret library. Your custom model archive need only contain your model artifacts if you use the environment correctly.",
   "programmingLanguage": "r",
-  "environmentVersionId": "680a835568a555e652af1f9a",
+  "environmentVersionId": "6810fa73791e1c002d2b2645",
```
yakov-g, line 6: I guess this update is typo?

yakov-g (approved):
1. I think default log level is WARNING, do we want to bump it to INFO?
2. Recently there was a case that we really wanted to know how much time we spend in the `score` hook. Do you plan to add logging before and after hooks? Ping me if help is needed.

Otherwise great!

public_dropin_environments/r_lang/env_info.json
peterzdeb: I dont't know, some Harness job did that commit to reconcile dependencies.

peterzdeb:
> 1. I think default log level is WARNING, do we want to bump it to INFO?
> 2. Recently there was a case that we really wanted to know how much time we spend in the `score` hook. Do you plan to add logging before and after hooks? Ping me if help is needed.
> 
> Otherwise great!

@yakov-g I see the logging for score start and finish were added recently (if I'm watching correct entrypoint): https://github.com/datarobot/datarobot-user-models/blob/master/custom_model_runner/datarobot_drum/drum/drum.py#L819. 
That's an INFO level so I think I'd prefer to enable INFO level by default. I'd like to discuss it with you, maybe later today? 

Thanks for the feedback.

rabih-datarobot (approved):
LGTM

custom_model_runner/datarobot_drum/drum/adapters/cli/shared/drum_class_label_adapter.py
```diff
@@ -60,7 +56,7 @@ def possibly_intuit_order(
 def infer_class_labels(target_type, input_filename, target_filename=None, target_name=None):
     # No class label information was supplied, but we may be able to infer the labels
     if target_type.is_classification():
-        print("WARNING: class list not supplied. Using unique target values.")
+        logger.warning("WARNING: class list not supplied. Using unique target values.")
```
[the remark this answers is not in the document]
peterzdeb: yeah, good point. I'll update this log message

PR #1407, opened 2025-04-30
[RAPTOR-12324] Using logging as best practive in model_templates

peterzdeb:


yakov-g:
You will not be able to bump DRUM version and set it in the envs in the same  PR.

The sequence should be like:
1. Bump DRUM version. Merge. Tag this commit with `v.1.6.13` Release new drum version.
https://datarobot.atlassian.net/wiki/spaces/RAPTOR/pages/1419247790/Publishing+DRUM+to+PyPi

2. Then update envs. You can do it immediately if you think it is very important or later.
https://datarobot.atlassian.net/wiki/spaces/RAPTOR/pages/1416691967/Install+drop-in+environments+on+staging+production+sts

Ping me for help with DRUM release and envs.

And I see those pipelines, that update IDs and requirements are super noisy. I'll try to refactor them.

yakov-g (approved):

model_templates/python3_sklearn_runtime_params/custom.py
```diff
@@ -4,47 +4,55 @@
...
+            },
         )
     else:
-        print("No credential data set")
+        logger.info("No credential data set")
 
     # boolean runtime param
     bool_var = RuntimeParameters.get("bool_var")
-    print(f"\tbool_var: {bool_var}")
+    logger.info("\tbool_var: %s", bool_var)
```
piratovskyi, line 47: I guess, these `\t` chars are not needed in case of logging? Or we still want to format strings with them?

piratovskyi (approved):
Looks good to me 👍

model_templates/python3_sklearn_runtime_params/custom.py
peterzdeb: This log will be also shown up when using cli. logging supports tab whitespaces so it should be good to preserve original formatting

piratovskyi: Sounds fair 👍

PR #1413, opened 2025-05-05
[RAPTOR-12446] Fix invalid test -nz

nullspoon:
The test builtin can't take multiple options in a single call, so this yields a syntax error. Additionally, `-n` (is set) and `-z` (is not set) presents a fun logic issue. :)

The intent appears to be to check if the related variable is set before proceeding, so this removes the `z` from `-nz` to accomplish that.


yakov-g (approved):
Approve and start build on your branch:
https://ci1.devinfra.drdev.io/job/Custom_Models_drop_in_environment_tests/6486/

klichukb (approved):

PR #1415, opened 2025-05-06
[RAPTOR-12416] add new reconciliation pipeline

yakov-g:

## Summary
Add a new pipeline based on older reconciliation and env version id update pipelines.

Improvements are:
* this pipeline has only one trigger for the parent envs folder. Currently it is configured to run on only one: `public_dropin_environments` 
* When there are changes in any env, pipeline is triggered. Script goes over all the `CHANGED` `public_dropin_environments` subfolders, runs `pip-compile` to generate requirements, then updates environment version id and commits changes into PR.

Then all the tests are automatically retriggered.
Here is an example of running this pipeline: https://github.com/datarobot/datarobot-user-models/pull/1414/files

Later I'll expand functionality on all the environments folders

### DIFF
--- a/.harness/reconcile_dependencies_for_all_envs_based_on_parent_folder.yaml
+++ b/.harness/reconcile_dependencies_for_all_envs_based_on_parent_folder.yaml
@@ -149,6 +149,7 @@ pipeline:
       value: <+input>.allowedValues(public_dropin_environments)
   identifier: reconcile_dependencies_for_all_envs_based_on_parent_folder
   description: |-
-    On PR, for every drop in env defined in input set, check reconcile dependencies.
-    Get requirements.in and generate requirements.txt.
+    On PR, for every drop in env in parent envs folder, e.g. public_dropin_environments:
+    * Enforce getting requirements.in and generate requirements.txt.
+    * Update environmentVersionId in env_info.json
   name: reconcile dependencies for all envs based on parent
--- a/.harness/reconcile_dependencies_for_all_envs_based_on_parent_folder.yaml
+++ b/.harness/reconcile_dependencies_for_all_envs_based_on_parent_folder.yaml
@@ -1,0 +1,163 @@
+pipeline:
+  projectIdentifier: datarobotusermodels
+  orgIdentifier: Custom_Models
+  tags: {}
+  stages:
+    - stage:
+        name: Reconcile Python Dependencies
+        identifier: Reconcile_Python_Dependencies
+        description: ""
+        type: CI
+        spec:
+          cloneCodebase: false
+          caching:
+            enabled: true
+          buildIntelligence:
+            enabled: true
+          execution:
+            steps:
+              - step:
+                  type: Run
+                  name: reconcile_dependencies
+                  identifier: reconcile_dependencies
+                  spec:
+                    connectorRef: account.dockerhub_datarobot_read
+                    image: datarobotdev/mirror_chainguard_datarobot.com_python-fips:3.11-dev
+                    shell: Bash
+                    command: |-
+                      # Ensure Git is installed
+                      if ! command -v git &> /dev/null; then
+                          echo "Git is not installed. Installing..."
+                          apt-get update && apt-get install -y git
+                      fi
+
+                      # Configure Git
+                      git config --global user.name "svc-harness-git2"
+                      git config --global user.email "svc-harness-git2@datarobot.com"
+                      git config --global url."https://${GITHUB_ACCESS_TOKEN}@github.com/".insteadOf "https://github.com/"
+                      git config --global --add safe.directory /harness  # Mark /harness as a safe directory
+
+                      pip install bson
+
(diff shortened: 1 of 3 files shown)


tools/env_version_update.py
```diff
@@ -62,6 +66,15 @@ def main(dir_to_scan, env=None):
         default=None,
         help="Name of the environment to update",
     )
+    parser.add_argument(
+        "-f",
+        "--file",
+        default=None,
+        help="Path of the env_info.json to update. If provided, other ",
```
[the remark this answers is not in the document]
yakov-g: Thanks, fixed

.harness/reconcile_dependencies_for_all_envs_based_on_parent_folder.yaml
```diff
@@ -0,0 +1,163 @@
...
+                      done
+
+                      # Run dependency reconciliation if there are file changes
+                      #if [ -n "$(echo $(git diff --name-only master -- ./requirements.*))" ]; then
+                      #  echo "Requirements changes detected in $ENV_DIR. Running reconciliation..."
+                      #  bash ${ROOT_DIR}/tools/reconcile_dependencies.sh
+                      #else
+                      #  echo "No requirements changes detected. Skipping reconciliation."
+                      #  exit 0  # Exit early if no changes
+                      #fi
```
peterzdeb, line 126: As I understand the changes are checked only for version id. The requirements (.in?) files are always auto generated?

.harness/reconcile_dependencies_for_all_envs_based_on_parent_folder.yaml
```diff
@@ -0,0 +1,154 @@
...
+            spec: {}
+  variables:
+    - name: env_folder
+      type: String
+      description: The folder containing the environment to reconcile, e.g. "public_dropin_environments"
+      required: true
+      value: <+input>.allowedValues(public_dropin_environments)
+  identifier: reconcile_dependencies_for_all_envs_based_on_parent_folder
+  description: |-
+    On PR, for every drop in env defined in input set, check reconcile dependencies.
```
peterzdeb, line 152: PR check trigger and the inputset should be created? perhaps in followup PR?

.harness/reconcile_dependencies_for_all_envs_based_on_parent_folder.yaml
yakov-g: I removed this commented out code.
Now we always run pip-compile if there are any changes in the environment. 
E.g. you publish new drum, then do any change(for example: change env version id or do some change in requirements.txt) to trigger reconciliation.

.harness/reconcile_dependencies_for_all_envs_based_on_parent_folder.yaml
yakov-g: I didn't clean this part when cloned pipeline, I'll clean

yakov-g:
> Looks ok, but the version update workflow seem to be very complicated.
> 
> Environment versions update should be either part of the PR, and being propagated on merge. Or, maybe it should be a separate Harness pipeline that we trigger manually to release a new version. Something to consider for tech design review.

Currently, when user updates environment his task is to bump environmentVersionId. Then later environment may be (or may be not ) published.

Now we are moving towards CI: on every env update image is created and pushed into registry, and then deployed with helm.  That's why we have this ID update every time env is changed.

.harness/reconcile_dependencies_for_all_envs_based_on_parent_folder.yaml
```diff
@@ -0,0 +1,155 @@
...
+            steps:
+              - step:
+                  type: Run
+                  name: reconcile_dependencies
+                  identifier: reconcile_dependencies
+                  spec:
+                    connectorRef: account.dockerhub_datarobot_read
+                    image: datarobotdev/mirror_chainguard_datarobot.com_python-fips:3.11-dev
+                    shell: Bash
+                    command: |-
```
nullspoon, line 27: This is a bit of a personal preference, so take this comment with a grain of salt. :)
But this looks rather complex for a yaml block. In the past, we've run into issues with debugging yaml embedded shell scripts, because there is no syntax highlighting and no good way to extract and test outside of the harness job.

Could we instead turn this block into a standalone script in the tools directory and call that script from this block?

nullspoon (approved):
Left a comment about complexity, but otherwise looks good to me.

.harness/reconcile_dependencies_for_all_envs_based_on_parent_folder.yaml
yakov-g: I will do it in the next iteration.
It doesn't directly related to our project, so for now I'll switch to higher priority things.

nullspoon: Works for me!

PR #1419, opened 2025-05-08
[BUZZOK-25745] Add support for streaming drum server run outputs

mjnitz02:

## Summary
This adds a new functionality to the `DrumRunServer` that allows it to "stream the output to the terminal" instead of waiting for the execution to complete and then displaying the entire set of logic.

## Rationale
**Why would I want this?**
We need to use the `DrumRunServer` to execute `custom-models` for development of Agentic workflows. Agents can become confused during the process and the process can appear to hang for an extended period of time. Because of the complex nature of `custom-models` we would like to be able to run them within the context of a drum server. Due to the way codespaces and other environments work with `scripts/` based execution we **MUST** execute all of this logic from within the context of a `python` program.

The problem here is that we then have a `synchronous` call being made to drum which runs the process in a `pOpen` context and waits for completion. To the user this appears as if the program is hung or no longer running. To resolve this I am proposing a completely optional variation of the `DrumServerRun` by adding the `stream_output` flag (default: False).

This uses a `queue` attached to the `subprocess` that consumes the `stdout` and `stderr` components of the `subprocess.pOpen` continuously. This queue can then be inspected externally while the process is executing asynchronously. When messages are added to the queue, they are sent to the terminal immediately instead of waiting for the process to complete.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/root_predictors/utils.py
+++ b/custom_model_runner/datarobot_drum/drum/root_predictors/utils.py
@@ -10,6 +10,8 @@ import shutil
 import shlex
 import subprocess
 import time
+from queue import Queue, Empty
+from threading import Thread
 
 from datarobot_drum.drum.common import get_drum_logger
 from datarobot_drum.drum.enum import (
@@ -102,6 +104,15 @@ def _create_custom_model_dir(
     return custom_model_dir
 
 
+def _queue_output(stdout, stderr, queue):
+    for line in iter(stdout.readline, b''):
+        queue.put(line)
+    for line in iter(stderr.readline, b''):
+        queue.put(line)
+    stdout.close()
+    stderr.close()
+
+
 def _exec_shell_cmd(
     cmd,
     err_msg,
@@ -110,6 +121,7 @@ def _exec_shell_cmd(
     env=os.environ,
     verbose=True,
     capture_output=True,
+    stream_output=False,
 ):
     """
     Wrapper used by tests and validation to run shell command.
@@ -132,7 +144,26 @@ def _exec_shell_cmd(
         process_obj_holder.process = p
 
     if capture_output:
-        (stdout, stderr) = p.communicate()
+        if stream_output:
+            q = Queue()
+            t = Thread(target=_queue_output, args=(p.stdout, p.stderr, q))
+            t.daemon = True  # thread dies with the program
+            t.start()
+            while True:
+                try:
+                    line = q.get_nowait()
+                    if len(line) > 0:
+                        print(line)
+                except Empty:
+                    if p.poll() is not None:
+                        break
+                    time.sleep(1)
+                except Exception:
+                    break
+            t.join()
+            stdout, stderr = "", ""
+        else:
+            (stdout, stderr) = p.communicate()
     else:
         stdout, stderr = None, None
 
--- a/custom_model_runner/datarobot_drum/drum/root_predictors/utils.py
+++ b/custom_model_runner/datarobot_drum/drum/root_predictors/utils.py
@@ -123,7 +123,11 @@ def _stream_p_open(subprocess_popen: subprocess.Popen):
     SEE: https://datarobot.atlassian.net/browse/RAPTOR-12510
     """
     logger_queue = Queue()
-    logger_thread = Thread(target=_queue_output, daemon=True, args=(subprocess_popen.stdout, subprocess_popen.stderr, logger_queue))
+    logger_thread = Thread(
+        target=_queue_output,
+        daemon=True,
+        args=(subprocess_popen.stdout, subprocess_popen.stderr, logger_queue),
+    )
     logger_thread.start()
     while True:
         try:
--- a/custom_model_runner/datarobot_drum/drum/root_predictors/utils.py
+++ b/custom_model_runner/datarobot_drum/drum/root_predictors/utils.py
@@ -127,7 +127,7 @@ def _stream_p_open(subprocess_popen: subprocess.Popen):
         try:
             # Stream output if available
             line = q.get_nowait()
-            print(line.strip()) if len(line.strip()) > 0 else None
+            logger.info(line.strip()) if len(line.strip()) > 0 else None
         except Empty:
             # Check if the process has terminated
             if subprocess_popen.poll() is not None:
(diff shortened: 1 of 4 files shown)


tsdaemon (approved):
Awesome! I think thats very nice, and I think we have not discussed enough topic of streaming api for chat/completions of agents, and streaming of logs. This might be later improvements, I think it will be highly valuable.

mbromanowsky (approved):
LGTM for Pred AI

custom_model_runner/datarobot_drum/drum/root_predictors/utils.py
```diff
@@ -102,6 +104,41 @@ def _create_custom_model_dir(
...
+    """Helper function to stream output from subprocess to a queue."""
+    for line in iter(stdout.readline, b""):
+        queue.put(line)
+    for line in iter(stderr.readline, b""):
+        queue.put(line)
+    stdout.close()
+    stderr.close()
+
+
+def _stream_p_open(p: subprocess.Popen):
```
eric-s-s, line 117: please do not use one-letter variable names.  it just adds cognitive load.  please describe well what each variable represents

mjnitz02: Sure, I can just call this `subprocess_popen`?

custom_model_runner/datarobot_drum/drum/root_predictors/utils.py
```diff
@@ -110,6 +147,7 @@ def _exec_shell_cmd(
     env=os.environ,
     verbose=True,
     capture_output=True,
+    stream_output=False,
```
eric-s-s, line 150: why not just set `capture_output` to False?  it does basically the same thing.

eric-s-s: but in a less roundabout manner

tests/unit/datarobot_drum/drum/root_predictors/test_drum_server_utils.py
```diff
@@ -141,3 +152,161 @@ def test_calls_thread_correctly(self, mock_get_command, runner):
...
+
+        # Configure stderr mock to return a sequence of lines and then stop
+        stderr_lines = [b"stderr line 1\n", b"stderr line 2\n", b""]
+        mock_stderr.readline.side_effect = stderr_lines
+
+        # Call the function
+        _queue_output(mock_stdout, mock_stderr, test_queue)
+
+        # Verify the queue contains all the expected output lines
+        expected_lines = [
```
eric-s-s, line 176: nit.  the ordering in this test feels suspicious.  that's _very_ much an implementation detail that all of stdout is streamed first and then all of stderr.  perhaps this would be more accurate as a `set` ?

custom_model_runner/datarobot_drum/drum/root_predictors/utils.py
```diff
@@ -102,6 +104,41 @@ def _create_custom_model_dir(
     return custom_model_dir
 
 
+def _queue_output(stdout, stderr, queue):
+    """Helper function to stream output from subprocess to a queue."""
+    for line in iter(stdout.readline, b""):
```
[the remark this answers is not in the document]
mjnitz02: It appears to stream both out when testing, but the error usually crashes the process so `stderr` may be being caught by something else?

custom_model_runner/datarobot_drum/drum/root_predictors/utils.py
mjnitz02: So I tried this on mattn/BUZZOK-25745-handle-streaming-drum-2 and this seems to still be a synchronous type of output where the execution needs to fully complete before the logs appear

mjnitz02: So this can sort of be done, but it isn't possible to log the outputs of `stdout` to a file if this is the case. So this "partially works" but not quite

tests/unit/datarobot_drum/drum/root_predictors/test_drum_server_utils.py
```diff
@@ -141,3 +152,161 @@ def test_calls_thread_correctly(self, mock_get_command, runner):
...
+        stdout, stderr = _stream_p_open(mock_process)
+
+        # Verify results
+        assert stdout == ""
+        assert stderr == ""
+        mock_print.assert_not_called()  # No output to print
+        mock_process.poll.assert_called_once()  # Process termination was checked
+
+    @patch("builtins.print")
+    def test_stream_p_open_handles_delayed_output(self, mock_print):
```
eric-s-s: I am _very_ suspicious of this test.  

1. it seems very very delicate
2. it relies on timing and sleeps
3. it takes at least a tenth of a second to run which is way way too long for a unit test (longer because you have a one-seocond wait for the process to close).

this test points to a larger testability issue here.

custom_model_runner/datarobot_drum/drum/root_predictors/utils.py
mjnitz02: Basically we need to be able to do both:
- `display the outputs to the terminal in realtime`
- Save the outputs to a file for browsing after the fact

peterzdeb (approved):

custom_model_runner/datarobot_drum/drum/root_predictors/utils.py
```diff
@@ -102,6 +104,41 @@ def _create_custom_model_dir(
...
+    """
+    q = Queue()
+    t = Thread(target=_queue_output, args=(subprocess_popen.stdout, subprocess_popen.stderr, q))
+    t.daemon = True  # thread dies with the program
+    t.start()
+    while True:
+        try:
+            # Stream output if available
+            line = q.get_nowait()
+            logger.info(line.strip()) if len(line.strip()) > 0 else None
```
yakov-g, line 132: Do I understand correct that it only shows logs to tell user that process is alive. 
But the actual output of  popen(drum .. .. )  is used

eric-s-s (approved):
after long discussion and taking into account deadlines, approving

custom_model_runner/datarobot_drum/drum/root_predictors/utils.py
yakov-g: Note.
THere is also a `StdoutFlusher` thread used to flush :)
Maybe you could simply add logging into it?

PR #1423, opened 2025-05-12
SemGrep detected `root` user in dronin environments

peterzdeb:

## Summary

All the reported images were updated with non-root user. For Changuard images we have `nonroot` and `nobody` for other Debian-type images.

## Rationale

SemGrep security scanner is configured for all DataRobot repositories. 
For the environment images with DRUM it reports two issues:
* `last-user-is-root`: The last user in the container is root. This is a security hazard because if an attacker gains control of the container they will have root access. Switch back to another user after running commands as root.
* `missing-user-entrypoint`: By not specifying a USER, a program in the container may run as root. This is a security hazard. If an attacker can control a process running as root, they may have control over the container. Ensure that the last USER in a Dockerfile is a USER other than root.

### DIFF
--- a/example_dropin_environments/julia_mlj/Dockerfile
+++ b/example_dropin_environments/julia_mlj/Dockerfile
@@ -24,4 +24,6 @@ ENV WITH_ERROR_SERVER=1
 #Uncomment the following line to switch to production mode
 #ENV PRODUCTION=1 MAX_WORKERS=4 SHOW_STACKTRACE=1
 
+USER nobody
+
 ENTRYPOINT ["/opt/code/start_server.sh"]
--- a/public_dropin_environments/java_codegen/Dockerfile
+++ b/public_dropin_environments/java_codegen/Dockerfile
@@ -68,4 +68,6 @@ ENV WITH_ERROR_SERVER=1 \
 COPY ./*.sh ${CODE_DIR}/
 WORKDIR ${CODE_DIR}
 
+USER nonroot
+
 ENTRYPOINT ["sh", "-c", "exec ${CODE_DIR}/start_server.sh \"$@\"", "--"]
--- a/custom_model_runner/CHANGELOG.md
+++ b/custom_model_runner/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ##### Changed
 - Using logs output for all server operations. `print` command is discouraged in user modules.
+- Non-root user is configured for drop-in environments 
 
 #### [1.16.12] - 2025-04-23
 ##### Changed
--- a/example_dropin_environments/python311_genai-gpu/Dockerfile
+++ b/example_dropin_environments/python311_genai-gpu/Dockerfile
@@ -25,4 +25,6 @@ COPY ./*.sh ${CODE_DIR}/
 
 ENV WITH_ERROR_SERVER=1
 
+USER nobody
+
 ENTRYPOINT ["/opt/code/start_server.sh"]
--- a/example_dropin_environments/python39_genai/Dockerfile
+++ b/example_dropin_environments/python39_genai/Dockerfile
@@ -27,4 +27,6 @@ ENV WITH_ERROR_SERVER=1
 # Uncomment the following line to switch from Flask to production mode
 #ENV PRODUCTION=1 MAX_WORKERS=4 SHOW_STACKTRACE=1
 
+USER nobody
+
 ENTRYPOINT ["/opt/code/start_server.sh"]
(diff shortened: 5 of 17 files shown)


public_dropin_environments/java_codegen/Dockerfile
```diff
@@ -32,25 +40,13 @@ COPY --from=build /bin/mkdir /bin/mkdir
...
-
 ENV VIRTUAL_ENV=/opt/venv
-
-RUN sh -c "python -m venv ${VIRTUAL_ENV} && \
-    . ${VIRTUAL_ENV}/bin/activate && \
-    python -m ensurepip --default-pip && \
-    python -m pip install --upgrade pip && \
-    python -m pip install --no-cache-dir -r requirements.txt && \
-    find ${VIRTUAL_ENV} -type d -name '__pycache__' -exec rm -rf {} + && \
-    rm -rf /usr/bin/rm /usr/bin/find"
```
peterzdeb, line 53: All that commands do not need a separate shell when building Docker image. I moved the dependencies installation step into `build` stage.

public_dropin_environments/java_codegen/Dockerfile
```diff
@@ -32,25 +40,13 @@ COPY --from=build /bin/mkdir /bin/mkdir
 # Required for custom-models to install dependencies
 COPY --from=build /usr/bin/pip /usr/bin/pip
 
-# Cleanup '__pycache__' directories. It solves an AsymmetricPrivateKey scanning error.
-COPY --from=build /usr/bin/rm /usr/bin/rm
-COPY --from=build /usr/bin/find /usr/bin/find
```
peterzdeb, line 37: No cleanup needed as we remove `__pycache__` in the `build` stage. The `rm` and `find` commands won't be needed in the runtime image

tsdaemon (approved):
LGTM Buzok

example_dropin_environments/julia_mlj/Dockerfile
```diff
@@ -24,4 +24,6 @@ ENV WITH_ERROR_SERVER=1
 #Uncomment the following line to switch to production mode
 #ENV PRODUCTION=1 MAX_WORKERS=4 SHOW_STACKTRACE=1
 
+USER nobody
```
klichukb, line 27: This might not matter too much. In LRS we run the container with a specific set of user id / group id (defaults are 1000:1000), also we force chown on the /opt/code to these.
I'm not sure if setting USER to something else would do it.

klichukb: can you test building such image and testing on staging and/or locally?

public_dropin_environments/java_codegen/Dockerfile
```diff
@@ -32,25 +40,13 @@ COPY --from=build /bin/mkdir /bin/mkdir
...
-
-RUN sh -c "python -m venv ${VIRTUAL_ENV} && \
-    . ${VIRTUAL_ENV}/bin/activate && \
-    python -m ensurepip --default-pip && \
-    python -m pip install --upgrade pip && \
-    python -m pip install --no-cache-dir -r requirements.txt && \
-    find ${VIRTUAL_ENV} -type d -name '__pycache__' -exec rm -rf {} + && \
-    rm -rf /usr/bin/rm /usr/bin/find"
+# Copy preinstalled virtualenv
+COPY --from=build --chown=nonroot ${VIRTUAL_ENV} ${VIRTUAL_ENV}
```
klichukb, line 49: this definitely needs testing with actual datarobot. we use 1000:1000 to run by default.

example_dropin_environments/julia_mlj/Dockerfile
peterzdeb: not sure I can test this on staging before merging the PR. Will try

peterzdeb: @klichukb I tested this image on Staging. It really lacks permissions if running for any non-root user (when tried to install model dependencies in the exec env).
The only solution to this is to grant permissions to all users for the `$HOME` directory which is "/opt". That directory keeps pip cache, virtualenvs the the code.

We can't know which user will run this container but we definitely can't run the containers from `root`. Chainguard will never let us to use `1000` as a user - there is no way to create custom users, it only gives us `nonroot` user that we can use when running the container - that's requirement from STIG/FIPS compliance.

peterzdeb: Right, for that reason I updated all the Dockerfiles in this PR.

peterzdeb:
Drop-in environments are running with security context by LRS. Whatever we set in the Dockerfile can be overriden on runtime. Users who request that custom environments may wish to reinstall packages, including DRUM

Planning to close this PR as won't do.
cc @Mint-Martini

PR #1439, opened 2025-05-15
[MMM-19441] Missing chat hook error message for agentic workflow

automactic:

## Summary

Update error message logic so that when `chat` is missing for agentic workflow models, the returning message say not implemented, instead of unsupported


piratovskyi (approved):

lkanggithub (approved):
LGTM

PR #1443, opened 2025-05-16
[RAPTOR-12536] Add busybox to the ChainGuard drop-in envs

elatt:

## Summary
Add `busybox` binary to ChainGuard based images

## Rationale
Many of the _standard_ binaries are copying into the final images are actually just symlinks to `busybox`. Docker appears to follow the symlink during the `COPY` operation but it seems this isn't part of the OCI spec and `Kaniko` doesn't always do this so it is safest to just also have the `busybox` binary in there too.

### DIFF
--- a/public_dropin_environments/java_codegen/Dockerfile
+++ b/public_dropin_environments/java_codegen/Dockerfile
@@ -17,6 +17,10 @@ USER root
 # Copy the JRE from the development stage
 COPY --from=build /usr/lib/jvm/java-11-openjdk /usr/lib/jvm/java-11-openjdk
 
+# Most of the binaries below are just symlinks to busybox and some OCI build tools follow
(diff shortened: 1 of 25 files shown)


klichukb (approved):

klichukb:
@piratovskyi its as minimal as possible by default. Then users add what they need, making security decisions on whats acceptable or not. So far not having basic tools made debugging containers, but in case of busybox I think Erik is doing that since PPS used in Bosun is likely using it for something.

Can a hacker do everything the same without any of these tools? Yes, because there's python interpreter, you can simulate any of these.

elatt:
@piratovskyi can you approve as well so I can get my two approvers so I can merge as this is to fix a live customer issue?

elatt:
I'll try to bundle the version bump into this PR as well.

tsdaemon (approved):

PR #1444, opened 2025-05-16
Move chainguard base to datarobot org

klichukb:
Overall we should not be relying on a private organization.
We're moving to a public org (repositories are still private).


yakov-g:
Let me create all the missing repos in `datarobot` and properly setup promotion pipeline before merging this

PR #1479, opened 2025-05-27
[MMM-19553] Fix issue with not returning _mod_pipeline.chat() result, release DRUM v1.16.14

rickwporter-datarobot:

## Summary

Fix issue with the call to the moderations pipeline chat was not returning a value. Enhanced tests to verify that the moderation pipelines return the a value.

## Rationale

This fix is needed for proper functioning of DRUM with moderations.


amarmudrankit (approved):
LGTM - Let us also bump the version of DRUM, because it will be required to be used in the buzzok and agentic images

PR #1481, opened 2025-05-27
[YOLO] enforce packages reconcile for all the (non GPU) envs

yakov-g:


ahamino (approved):
In the spirit of breaking production =D

PR #1492, opened 2025-05-30
[MMM-19600] release DRUM v1.16.15

nickolai-dr:
For more nim testing I need most recent drum in relevant environments, here is PR that bumps version requirements files.

PR #1493, opened 2025-05-30
[MMM-19604] Bump drum version in the environments.

nickolai-dr:
Follow up for https://github.com/datarobot/datarobot-user-models/pull/1492 with env bump.
new version on PyPI https://pypi.org/project/datarobot-drum/1.16.15/


yakov-g:
just heads up.
We don't functional tests for nim environment.
We might have them, but they are not enabled.

PR #1495, opened 2025-06-01
[RAPTOR-12427] Less verbose 422 when reporting monitoring data from chat completions

amarmudrankit:
## Summary

When Text Gen deployment is created, the feature drift and predictions data collections are off by default.  When /chat/completions request is received DRUM will report monitoring data using API Spooler and will fail with 422 - creating big stack trace. If, by default both flags are off, particularly for this use case of DRUM + chat completions, DRUM shouldn’t be this verbose.

This patch will put the warning - but not the stack trace

Manually tested it to ensure warning is printed but not the stack trace.

### DIFF
--- a/tests/unit/datarobot_drum/drum/language_predictors/test_base_language_predictor.py
+++ b/tests/unit/datarobot_drum/drum/language_predictors/test_base_language_predictor.py
@@ -1,7 +1,9 @@
 import os
 from unittest.mock import patch, Mock, ANY
 
+import datarobot_drum
 import pytest
+from datarobot_mlops.common.connected_exception import DRMLOpsConnectedException
 from openai.types.model import Model
 import pandas as pd
 import numpy as np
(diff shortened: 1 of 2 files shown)


ScottOglesby (approved):

tests/unit/datarobot_drum/drum/language_predictors/test_base_language_predictor.py
```diff
@@ -436,6 +438,41 @@ def chat_hook(completion_request):
...
+                        "model": "any",
+                        "messages": [
+                            {"role": "system", "content": "You are a helpful assistant."},
+                            {"role": "user", "content": "Hi"},
+                        ],
+                    }
+                )
+                mock_warning.assert_called_once()
+                args, _ = mock_warning.call_args
+                assert exception_string in args[0]
```
mbromanowsky: This works fine, but did you know about the magic pytest builtin fixture `caplog`? This fixture is always available and basically does what you want here, I believe.

Pytest docs about it: https://docs.pytest.org/en/stable/how-to/logging.html
An example of use in the monorepo: https://github.com/datarobot/DataRobot/blob/06dc3995c35c2bcd52c476eac1257dad0faa6b8a/MLDev_Tests/autopilot/tests/backend/integration/stages/test_bias_mitigation_stage.py#L351-L358

tests/unit/datarobot_drum/drum/language_predictors/test_base_language_predictor.py
```diff
@@ -436,6 +438,41 @@ def chat_hook(completion_request):
...
+            assert True
```
mbromanowsky: Nit: this clearly does nothing.

mbromanowsky (approved):
LGTM for PredAI, but consider using builtin `caplog` instead of having to patch the logger yourself.

tests/unit/datarobot_drum/drum/language_predictors/test_base_language_predictor.py
amarmudrankit: Fixed it

tests/unit/datarobot_drum/drum/language_predictors/test_base_language_predictor.py
amarmudrankit: Fixed it

PR #1504, opened 2025-06-06
Demo

klichukb:

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/root_predictors/generic_predictor.py
+++ b/custom_model_runner/datarobot_drum/drum/root_predictors/generic_predictor.py
@@ -4,6 +4,7 @@ All rights reserved.
 This is proprietary source code of DataRobot, Inc. and its affiliates.
 Released under the terms of DataRobot Tool and Utility Agreement.
 """
+import json
 import urllib
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


yakov-g (approved):
Very nice.
Maybe implement that `inline.py` as a test

custom_model_runner/datarobot_drum/drum/main.py
```diff
@@ -92,7 +91,12 @@ def signal_handler(sig, frame):
 
         options = arg_parser.parse_args()
         CMRunnerArgsRegistry.verify_options(options)
-        _setup_required_environment_variables(options)
+
+        try:
+            setup_required_environment_variables(options)
+        except Exception as exc:
+            print(str(exc))
```
nullspoon, line 103: Might be a python thing I don't know about (and this probably doesn't really matter that much), but should we send this to stderr?

nullspoon: Oh nevermind. This just moves the code up.

klichukb: Yeah I just lift-and-shifted it, did not want to alter but it should use logger.exception() instead of this.

yakov-g:
Looks great!

klichukb:
@yakov-g @mjnitz02 I tried an e2e test with moderations to make sure that the inline runner executes the moderations pipeline. Its crazy because it triples the duration of tests even with bumping 8G to 16G for the functional test suite. Installing moderations is adding more than 30 minutes, I'm not sure why this is so bad.

I'm skipping the test to get things in for the release, but I think we have to figure out how to get the test in.

PR #1507, opened 2025-06-09
[YOLO] Move Triton env to example

klichukb:
As discussed with @ahamino , we move Triton environment to examples, as opposed to shipping it as a maintained image.
It's an Nvidia image and for now we have no estimates on when such image (along with others) will be FIPS-compliant.

vLLM on the other hand is something under negotiation with Chainguard but also is the one typically used for non-NIM workloads.


yakov-g:
I fixed and restarted pipeline and so bot committed reconciliation.
I have a PR to add a whitelist on paths to run reconcile on.

nullspoon (approved):

PR #1517, opened 2025-06-17
[RAPTOR-12838] add streaming to chat example

ScottOglesby:

## Summary
Small change to add streaming support to the chat example for `custom.py`

## Rationale
* a reference implementation for OpenAI chat streaming
* a convenient streaming stub model for dev and test

## Changes
* `chat()` hook now respects the `stream` field in `completion_create_params`
* each chunk is one word of the response
* default 0.25 second delay between chunks
* response uses OpenAI Python classes
* cleaned up file to satisfy pylint and black

## Testing
```
export TARGET_NAME=completions
drum server --code-dir model_templates/python3_dummy_chat/ --target-type textgeneration --address localhost:6789
```

```
from openai import OpenAI

openai_client = OpenAI(
    base_url="http://localhost:6789",
    api_key="not used",
)

# non-streaming
completion = openai_client.chat.completions.create(
    model="datarobot-deployed-llm",
    messages=[{"role": "user", "content": "What is the meaning of life?"}],
)
print(completion.choices[0].message.content)

# streaming
completion = openai_client.chat.completions.create(
    model="datarobot-deployed-llm",
    messages=[{"role": "user", "content": "What is the meaning of life?"}],
    stream=True,
)
for chunk in completion:
    print(chunk)
```

### DIFF
--- a/model_templates/python3_dummy_chat/custom.py
+++ b/model_templates/python3_dummy_chat/custom.py
@@ -6,22 +6,19 @@ Released under the terms of DataRobot Tool and Utility Agreement.
 """
 import calendar
 import time
-from typing import Iterator
+from typing import Any, Iterator
+import uuid
 
 from openai.types.chat import ChatCompletion
 from openai.types.chat import ChatCompletionChunk
 from openai.types.chat import ChatCompletionMessage
 from openai.types.chat import CompletionCreateParams
 from openai.types.chat.chat_completion import Choice
+from openai.types.chat.chat_completion_chunk import Choice as ChunkChoice
+from openai.types.chat.chat_completion_chunk import ChoiceDelta
 from openai.types.model import Model
 
-from datarobot_drum import RuntimeParameters
-
-"""
-This example shows how to create a text generation model supporting OpenAI chat
-"""
-
-from typing import Any, Dict
+# This example shows how to create a text generation model supporting OpenAI chat
 
 
 def get_supported_llm_models(model: Any):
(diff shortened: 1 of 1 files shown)


model_templates/python3_dummy_chat/custom.py
```diff
@@ -62,29 +60,63 @@ def load_model(code_dir: str) -> Any:
...
-    model = completion_create_params["model"]
+    _ = model
+    inter_token_latency_seconds = 0.25
+    model_id = completion_create_params["model"]
     message_content = "Echo: " + completion_create_params["messages"][0]["content"]
+    stream = completion_create_params.get("stream", False)
+
+    if stream:
+        # Mimic OpenAI streaming: yield one chunk at a time, split by whitespace
+        def gen_chunks():
```
peterzdeb, line 85: Do we have tests for this streaming mode?

yakov-g (approved):

model_templates/python3_dummy_chat/custom.py
ScottOglesby: @peterzdeb There generally is "can the custom model be created" testing (`test_inference_model_templates.py`) and some negative tests (`test_other_cases.py`).

However, I created a more direct test in `test_model_templates.py`. It's unit instead of full e2e, but exercises the python model adapter using a Flask server, with an OpenAI client driving the test. It adds some value, and we can add other model template tests later.

ScottOglesby:
@yakov-g I added a test for the chat hook based on Petro's review comment. Also moved a few fixtures into `tests/unit/datarobot_drum/drum/conftest.py` since they are now used in 2 test modules.

The PR already has 2 approvals, but the changes are large enough that "sneaking them in" seems wrong. I'll ping for a re-review to make sure the test changes are OK.

yakov-g (approved):

PR #1523, opened 2025-06-20
Yakov/bump all deps

yakov-g:


kijimoshib (approved):

public_dropin_environments/java_codegen/requirements.txt
```diff
@@ -10,31 +10,30 @@ azure-core==1.34.0
...
+certifi==2025.6.15
 cffi==1.17.1
 charset-normalizer==3.4.2
 click==8.2.1
-cryptography==45.0.3
+cryptography==45.0.4
 datarobot==3.7.1
 datarobot-drum==1.16.17
-datarobot-mlops==11.1.0a3
+datarobot-mlops==11.0.0
```
mbromanowsky: This seems to be a downgrade, 11.1 to 11.0 -- is that intentional?

mbromanowsky (approved):
Generally LGTM for Predictive AI.

I also reviewed a PR from dependabot, bumping just protobuf in just the xgboost drop-in environment. That PR https://github.com/datarobot/datarobot-user-models/pull/1521 would conflict with this one -- but yours is essentially a superset of that one. Just FYI -- if you can get this green and merged before that one goes in, I think that's OK.

public_dropin_environments/java_codegen/requirements.txt
yakov-g: Looks like pip-compile always take the latest released version, ignoring pre-releases.
Pinning it in the requirements.in

mbromanowsky: Oh, I see, it was a pre-release. I could see it either way, then. Do you know, is the stable release 11.1 of `datarobot-mlops` coming soon?

yakov-g: I'm trying to figure out, but no luck yet

klichukb:
Overall I'm not sure the intention of bumping everything to latest is good when signoffs and testing are getting to the end. Such a move is probably good but earlier in the cycle.

yakov-g:
> Overall I'm not sure the intention of bumping everything to latest is good when signoffs and testing are getting to the end. Such a move is probably good but earlier in the cycle.

Yeah, I tried to do it earlier. But more important tasks hopped in.
I'll update only DRUM and mlops then

PR #1525, opened 2025-06-23
[BUZZOK-26535] Update genai agents environment for tracing and moderations

mjnitz02:

## Summary
Small update to the environment to ensure that libraries are up to date.

### DIFF
--- a/public_dropin_environments/python311_genai_agents/requirements.in
+++ b/public_dropin_environments/python311_genai_agents/requirements.in
@@ -28,7 +28,7 @@ llama-index-llms-langchain
 llama-index-llms-litellm
 openai
 ragas @ git+https://github.com/explodinggradients/ragas@5d59549ad5ef511f621502c563bc55ac5aeb9188#subdirectory=ragas
-datarobot-moderations>=11.1.15
+datarobot-moderations>=11.1.17
 
 # Generative AI Telemetry
 traceloop-sdk


amarmudrankit (commented):
Changes are good, but:
1. Moderations library 11.1.17 version is available so that can be used.  I will need to bump it one more time to get in one critical fix, but will create a separate PR to bump it.  Don't want to hold this PR for that
2. I am surprised the agent env is in public_dropin_env.  Then where do we use [this env](https://github.com/datarobot/af-component-agent/tree/main/template/%7B%7Bagent_app_name%7D%7D/docker_context)?  Should we update this one too?

ltraub (approved):
:+1:

mjnitz02:
@amarmudrankit So this is for users to have a pre-selectable environment that is pre-built. This allows a user to define this in their `.env` and get the agent up quickly without needing to build from the `docker_context`. https://github.com/datarobot/recipe-datarobot-agent-templates/blob/main/.env.sample#L21

Due to a variety of reasons the user may not want to do that, and then they can leave the `.env` var blank and it will build from the `docker_context` folder.

The ultimate plan is probably to remove the `docker_context` from the component at some point, but with continued changes and application templates themselves often implementing custom `docker_context` changes, exactly how we do this is a bit confusing. So for right now the `docker_context` is the "most up to date version that can be built during a pulumi up" or if the user doesn't want any custom updates they can simply put this name as a pre-built environment and it saves ~10 minutes on the first deploy.

public_dropin_environments/python311_genai_agents/requirements.txt
```diff
@@ -62,7 +62,7 @@ dataclasses-json==0.6.7
 datarobot==3.7.1
 datarobot-drum==1.16.16
 datarobot-mlops==11.1.0a3
-datarobot-moderations==11.1.14
+datarobot-moderations==11.1.17
```
mjnitz02, line 65: @amarmudrankit I'll update the `requirements.in` It looks like this selected .17 by default here

amarmudrankit:
Ok, thanks @mjnitz02 so if we need to bump moderation lib version is it sufficient to do it in this env or we should also bump it in `af-component` as well? I guess its later?

mjnitz02:
@amarmudrankit We probably need to do it in both. I can look at trying to make some form of automation that can help do this. I've been trying to see if I a change to the `pyproject.toml` could propagate everywhere somehow.

PR #1528, opened 2025-06-24
[YOLO] refactor to remove distutils usage

yakov-g:


custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -1211,35 +1231,74 @@ def _generate_runtime_report_file(self, fit_mem_usage: float, pred_mem_usage: fl
...
+def _output_in_code_dir(code_dir: Path, output_dir: Path) -> bool:
+    """Return True if output_dir is inside code_dir."""
+    try:
+        output_dir.resolve().relative_to(code_dir.resolve())
+        return True
+    except ValueError:
+        return False
+
+
+def _copy_tree(src_dir: Path, dst_dir: Path) -> set[Path]:
```
nullspoon, line 1221: Why do we need this one if we have the global `copy_tree` above? IMO, we should have one or the other, but not both (unless I'm missing something?).

From a quick search, looks like the global `copy_tree` isn't referenced here, so that one seems the easiest to not have.

yakov-g: THanks!

yakov-g: removed

nullspoon: Faaaaaannnntastic! :shipit:

PR #1530, opened 2025-06-24
[RAPTOR-13924] Remove python311_genai environment

amarmudrankit:
This environment was specifically used for global models work.  We are now maintaining the global models environment in its own repository and not here.  Removing all references to this environment from here


yakov-g:
Looks good.
SO you are not enabling any tests for genai_agents environments?
We just keep an eye on building it, right?

pipelineController.yaml
```diff
@@ -66,13 +66,6 @@ test_functional_python_pytorch: &test_functional_python_pytorch
     FRAMEWORK: python3_pytorch
   phase: 3
 
-test_functional_python311_genai: &test_functional_python311_genai
```
eric-s-s, line 69: are we still using jenkins???

eric-s-s (approved):

amarmudrankit:
> Looks good. SO you are not enabling any tests for genai_agents environments? We just keep an eye on building it, right?

Thats what my confusion is - what is framework and framework env in the context of these tests?

If you look at this line https://github.com/datarobot/datarobot-user-models/pull/1530/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128L202 why is agentic workflow using python311_genai or how is it using / being tested? The environment being removed has no agentic support not sure why it was added to use for any agentic testing. @automactic do you know?

amarmudrankit:
ok yeah, makes sense.  @yakov-g does my change mean there were some tests for genai_agents environment and I removed it? I guess not - because I only removed the references to `python311_genai` and not touched `python311_genai_agents`.  So, if there were tests intended for `python311_genai_agents` they should remain as is, unless I am completely misunderstanding how these tests are run

yakov-g:
> ooks good.
> SO you are not enabling any tests for genai_agents environments?
> We just keep an eye on building it, right?



> ok yeah, makes sense. @yakov-g does my change mean there were some tests for genai_agents environment and I removed it? I guess not - because I only removed the references to `python311_genai` and not touched `python311_genai_agents`. So, if there were tests intended for `python311_genai_agents` they should remain as is, unless I am completely misunderstanding how these tests are run

THere are no tests for genai_agents env. Thats why I asked, if we want not to remove tests, but switch them to run on `genai_agents. 
If you decide to do this, you only need to partially touch affected files. You don't need to create pipeline inputs, etc.
I'll go over changes and comment in some places

I won't be able to help with this today. Busy on another customer issues.

jenkins/test_functional_per_framework.sh
```diff
@@ -25,8 +25,6 @@ elif [ "$1" = "python3_pmml" ]; then
     DOCKER_IMAGE="python3_pmml"
 elif [ "$1" = "python3_pytorch" ]; then
     DOCKER_IMAGE="python3_pytorch"
-elif [ "$1" = "python311_genai" ]; then
```
yakov-g, line 28: not sure it is used and if it is needed to be defined to run tests on genai_agents

tests/e2e/conftest.py
```diff
@@ -139,13 +139,6 @@ def pytorch_drop_in_env():
     )
 
 
-@pytest.fixture(scope="session")
```
yakov-g, line 142: these(E2E) tests are not run at all

tests/functional/test_per_framework.json
```diff
@@ -8,7 +8,6 @@
     {"env_folder":  "public_dropin_environments", "framework":  "python3_sklearn", "with_local":  true},
     {"env_folder":  "public_dropin_environments", "framework":  "python3_xgboost", "with_local":  true},
     {"env_folder":  "public_dropin_environments", "framework":  "python311", "with_local":  true},
-    {"env_folder":  "public_dropin_environments", "framework":  "python311_genai", "with_local":  true},
```
yakov-g, line 11: Yes, define it here

tests/conftest.py
```diff
@@ -199,7 +198,6 @@
         CUSTOM_TASK_INTERFACE_PYTORCH_BINARY,
         CUSTOM_TASK_INTERFACE_PYTORCH_MULTICLASS,
     ],
-    PYTHON311_GENAI: [PYTHON_TEXT_GENERATION, PYTHON_AGENTIC_WORKFLOW],
```
yakov-g, line 202: Removing these was ok as these test case was never triggered in this env as you removed it from test_per_framewor.json

tests/functional/test_per_framework.json
yakov-g: framework would be "python311_genai_agents"

tests/constants.py
```diff
@@ -31,7 +31,6 @@
 PYTHON_XGBOOST = "python3_xgboost"
 PYTHON_KERAS = "python3_keras"
 PYTHON_PYTORCH = "python3_pytorch"
-PYTHON311_GENAI = "python311_genai"
```
yakov-g, line 34: rename constant and value

jenkins/test_functional_per_framework.sh
yakov-g: This one not needed

amarmudrankit:
@yakov-g if we want to add tests for `python311_genai_agents`, then it is not in the scope of this ticket.  I just want to cleanly remove `python311_genai` environment.  I am neither expert  nor owner of the environment `python311_genai_agents` and the corresponding team / owner should add those tests.  However, I will consider adding these tests that are removed in global-envs-models repository - because the environment is being hosted there.

yakov-g (approved):

PR #1535, opened 2025-06-25
[RAPTOR-14024] set PYTHONUNBUFFERED=1 in vllm env

yakov-g:

## Summary
This allows to see `print` right away.
We had it in all or envs before, and have it now in chainguard based envs

### DIFF
--- a/public_dropin_gpu_environments/vllm/env_info.json
+++ b/public_dropin_gpu_environments/vllm/env_info.json
@@ -4,7 +4,7 @@
   "description": "A high-throughput and memory-efficient inference and serving engine for LLMs.",
   "programmingLanguage": "python",
   "label": "v0.8.3+dr.1",
-  "environmentVersionId": "6848b64a2081a81aa6129277",
+  "environmentVersionId": "685c32fd6796f40f96d7c546",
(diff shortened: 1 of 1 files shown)


public_dropin_gpu_environments/vllm/Dockerfile
```diff
@@ -37,5 +37,7 @@ WORKDIR ${CODE_DIR}
 COPY --chown=1000:0 ./*.sh ./*.py ${CODE_DIR}/
 
 ENV WITH_ERROR_SERVER=1
+ #This makes print statements show up in the logs API
+ENV PYTHONUNBUFFERED=1
```
peterzdeb, line 41: Is there a misuse of `print()` instead of `logging` in that deployment? I'd not expect this issue to occur with logging to stream output - that should be the only supported output method for DRUM running in containers.

yakov-g: This is user's code in custom.py.
We will recommend using logging, but we can not enforce.

klichukb: other envs might also need this if we dont do that there

yakov-g: we have it everywhere else

PR #1536, opened 2025-06-25
[RAPTOR-14025] Allow override nim server details

klichukb:

## Summary
Applying changes to the latest DRUM version and release.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/gpu_predictors/nim_predictor.py
+++ b/custom_model_runner/datarobot_drum/drum/gpu_predictors/nim_predictor.py
@@ -29,7 +29,7 @@ class NIMPredictor(BaseOpenAiGpuPredictor):
     ENGINE_CONFIG_FILE = "engine_config.json"
     LEGACY_START_SERVER_SCRIPT = Path("/opt/nim/start-server.sh")
     START_SERVER_SCRIPT = Path("/opt/nim/start_server.sh")
-    HEALTH_ROUTE = "/v1/health/ready"
+    DEFAULT_HEALTH_ROUTE = "/v1/health/ready"
 
     def __init__(self):
         super().__init__()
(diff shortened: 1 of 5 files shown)


yakov-g (approved):

custom_model_runner/datarobot_drum/drum/gpu_predictors/nim_predictor.py
```diff
@@ -41,6 +41,17 @@ def __init__(self):
         self.max_model_len = self.get_optional_parameter("NIM_MAX_MODEL_LEN")
         self.log_level = self.get_optional_parameter("NIM_LOG_LEVEL")
 
+        # DR prefix to denote that this is not some standard NIM variable of any sort.
+        self.health_route = self.get_optional_parameter(
+            "DR_NIM_HEALTH_ROUTE", default_value=self.DEFAULT_HEALTH_ROUTE
+        )
+        self.openai_port = self.get_optional_parameter(
+            "DR_NIM_SERVER_PORT", default_value=self.openai_port
```
klichukb, line 49: There is an attempt.. to have a standard NIM_SERVER_PORT, which should work on the NIM side, but as always its inconsistent.
Anyway, these vars are for DRUM side, not NIM

custom_model_runner/datarobot_drum/drum/gpu_predictors/nim_predictor.py
```diff
@@ -41,6 +41,17 @@ def __init__(self):
...
+        )
+        self.openai_host = self.get_optional_parameter(
+            "DR_NIM_SERVER_HOST", default_value=self.openai_host
```
klichukb, line 52: its already read from env var, but runtime param is a different beast ⭐ as it has a prefix, so for backwards compat we will support both ways.

PR #1538, opened 2025-06-25
[cherry-pick] Migrate opensource envs to chainguard (#1526)

yakov-g:


klichukb (approved):
they were chainguard, just in a different repo

PR #1571, opened 2025-07-11
[RAPTOR-14178] improve harness images build pipeline

yakov-g:

## Summary
Make stage node name unique by adding matrix iteration number so it will be shown as 
image

Though stage name is unique even without iteration number, it seems that it is trimmed for github PR status, so:
`image_build_env_python_1234` 
and 
`image_build_env_python_1234.local`
will be smth like `image_build_env_python`, so only one status is shown in PR.

I'll change pipeline name later.


yakov-g:
jarvis please cherrypick release/11.1

nullspoon (approved):

PR #1576, opened 2025-07-11
[BUZZOK-26787] Force remove outdated asyncio package from container, it breaks codespaces

mjnitz02:

## Summary
There is an outdated `asyncio` package (https://pypi.org/project/asyncio/) being installed as part of `datarobot-moderations`. This package is unnecessary in python > 3.4 and it also breaks compatibility with a bunch of other packages. We need to ensure this package is not installed. I will open a parallel PR to remove this from moderations, but we need to have a fix faster then that turnaround for these environments.

⚠️ **Including this package is a critical failure as a codespace GUI cannot be started if this package is present.**

RELEASE 11.1 BACKPORT: https://github.com/datarobot/datarobot-user-models/pull/1575


mjnitz02:
cc @amarmudrankit

PR #1594, opened 2025-07-16
[RAPTOR-14249] Remove python3 PMML dropin env

nullspoon:
This deletes the python3_pmml public dropin environment. The upstream project isn't frequently maintained, and there are CVEs for it now that have yet to be fixed. As this is not commonly used, we're just dropping it now.


custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
```diff
@@ -1,5 +1,5 @@
 """
-Copyright 2021 DataRobot, Inc. and its affiliates.
+Copyright 2025 DataRobot, Inc. and its affiliates.
```
mbromanowsky: I don't think copyright years should be updated when changing existing files.

nullspoon: Well look at that! I wasn't aware that had changed. I learned many years ago that any major releases of software merited a copyright year update. Seems that is no longer the case these days. It should reflect original date of publish only.
Good call!

nullspoon: Annnnd they're all back the way they were. Thanks for the review!

mbromanowsky: Thanks! This is based on slack conversations with our former director of IP and patents. I'll send you a link.

PR #1615, opened 2025-08-06
[BUZOK-27100] Fix Drum Inline Runner and streamline DRUM options generation

mjnitz02:


tests/unit/datarobot_drum/drum/utils/test_setup.py
mjnitz02, line 1: There weren't any tests I could find really related to this code flowing through, so here is a small suite of new tests

kijimoshib:
LGTM
@yakov-g could you please double check this?

PR #1673, opened 2025-09-25
[RAPTOR-14422] Drum will get terminated if it fail to load and parse config file

reza-mahdiani-m:
## Summary
This pull request introduces "fast fail" mechanism to the DRUM runner when failing to parse a configuration file.

The implementation includes:
A new UnrecoverableError exception hierarchy to signal fatal errors.
An update to the main DRUM runner to catch these specific exceptions and terminate the process.
A new try-except block for loading config-file

## Rationale
A model could fail to initialize correctly due to a simple typo in a config file, but the DRUM server would continue running. Users would only discover the problem after the deployment timed out or when prediction requests failed, wasting time and resources.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/exceptions.py
+++ b/custom_model_runner/datarobot_drum/drum/exceptions.py
@@ -44,3 +44,11 @@ class DrumSerializationError(DrumException):
 
 class DrumRootComponentException(DrumException):
     """Raised when there is an issue specific to root components."""
+
+
+class UnrecoverableError(DrumException):
+    """A base exception for any error that is considered fatal and main runner should terminate immediately."""
+
+
+class UnrecoverableConfigurationError(UnrecoverableError):
+    """Raised when failure in parsing or validating configuration file."""
--- a/custom_model_runner/datarobot_drum/drum/drum.py
+++ b/custom_model_runner/datarobot_drum/drum/drum.py
@@ -22,7 +22,6 @@ from typing import Callable, Optional
 from typing import Dict
 from typing import Union
 
-
 import docker.errors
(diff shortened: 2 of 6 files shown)


custom_model_runner/datarobot_drum/drum/exceptions.py
```diff
@@ -44,3 +44,11 @@ class DrumSerializationError(DrumException):
 
 class DrumRootComponentException(DrumException):
     """Raised when there is an issue specific to root components."""
+
+
+class UnrecoverableError(DrumException):
```
yakov-g, line 49: I think we don't need a separate subclass. 
When configuration yaml can no be parsed we use:
class DrumFormatSchemaException(DrumException):
    """Raised when supplied model_metadata cannot be parsed using strict yaml"""
    
with DrumException as the base

klichukb: @yakov-g do you have any other feedback?

yakov-g:
@klichukb 
Bogdan, 
Reza and I tested this approach more and turns out it doesn't work.
We reviewed the ways to test and debug drum and Reza will try to find a better solution.

custom_model_runner/datarobot_drum/drum/exceptions.py
reza-mahdiani-m: I used UnrecoverableError for the case that we want the system crash immediately, now for configuration later maybe for some other cases. However we can define some other behaviour for DrumFormatSchemaException(DrumException) such as graceful failure or even not failing at all and considering some default values. I used another subclass to separate different error handling polices.

PR #1677, opened 2025-09-26
[BUZZOK-27771] Optimize DRUM to import heavy stuff conditionally

tsdaemon:

BEFORE MERGING THIS I AM GOING TO TEST IT IN AGENTIC TEMPLATES

## Summary

This optmizes DRUM a bit by moving inline imports of heavy libraries to places where they are used instead of at the top of the file. This improves runtime of CLI invocation.

This reduces time (when no moderations in the agent) from 8.95s to 3s.

## Rationale

CLI call before optimization:
image
[out.html](https://github.com/user-attachments/files/22564126/out.html)

CLI call after optimization:
image
[out_new.html](https://github.com/user-attachments/files/22564251/out_new.html)

Primarily this avoids loading moderations if they are not used (which should also speed up all textgen and vdb models). Besides that it moves several imports inline. I was not able to move inline another offender `pandas`, because it requires heavy refactoring of `PythonModelAdapter`. But there def were few noticeable improvements.

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/description.py
+++ b/custom_model_runner/datarobot_drum/drum/description.py
@@ -4,6 +4,6 @@ All rights reserved.
 This is proprietary source code of DataRobot, Inc. and its affiliates.
 Released under the terms of DataRobot Tool and Utility Agreement.
 """
-version = "1.17.3a1"
+version = "1.17.3"
 __version__ = version
 project_name = "datarobot-drum"


tsdaemon:
@amarmudrankit @klichukb is it possible to build a dev version I can use in templates to verify it?

GunnerGlory (approved):

amarmudrankit (approved):
Changes look good to me

amarmudrankit:
> @amarmudrankit @klichukb is it possible to build a dev version I can use in templates to verify it?

Not that I am aware of.  I typically would build an image with the wheel file created from the branch and test it there.

README.md
```diff
@@ -135,8 +154,10 @@ To get more information, search for `custom models` and `datarobot user models`
 #### Testing changes to drum in DR app
 There is a script called `create-drum-dev-image.sh` which will build and save an image with your latest local changes to the DRUM codebase. You can test new changes to drum in the DR app by running this script with an argument for which dropin env to modify, and uploading the image which gets built as an execution environment. 
 
+You can also [publish a development version](https://datarobot.atlassian.net/wiki/spaces/RAPTOR/pages/1419247790/Publishing+DRUM+to+PyPi#I-Just-Want-To-Test-My-Thing) which you can pin in your dependencies.
```
tsdaemon: @klichukb here I added the link you have shared. I also made dev prerequisites a little bit more structured, leaving some space to explain how to work with in on MacOS. Its not critical now, but there might be need eventually.

PR #1678, opened 2025-09-30
[RAPTOR-14734] update drum builder image

yakov-g:

## Summary
Tested pipeline manually

### DIFF
--- a/docker/drum_builder/build_image.sh
+++ b/docker/drum_builder/build_image.sh
@@ -22,15 +22,7 @@ done
 GIT_ROOT=$(git rev-parse --show-toplevel)
 BUILD_DOCKER_DIR=$GIT_ROOT/docker/drum_builder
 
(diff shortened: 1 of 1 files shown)


docker/drum_builder/build_image.sh
```diff
@@ -22,14 +22,14 @@ done
 GIT_ROOT=$(git rev-parse --show-toplevel)
 BUILD_DOCKER_DIR=$GIT_ROOT/docker/drum_builder
 
-IMAGE_ORG_DATAROBOT=datarobot
+#IMAGE_ORG_DATAROBOT=datarobot
 IMAGE_ORG_DATAROBOTDEV=datarobotdev
```
nullspoon: Do we even need this variable now since this is purely a datarobotdev image? Wondering if we should just hardcode datarobotdev rather than variablizing this.

PR #1684, opened 2025-10-02
[BUZZOK-27659] Migrate genai agents images to use UV instead of pip

mjnitz02:


public_dropin_environments/python311_genai_agents/pyproject.toml
```diff
@@ -0,0 +1,214 @@
+[project]
+
+name = "DockerContext"
+version = "0.1.0"
+description = "Implementation of DockerContext"
```
akshoop, line 5: should we update the name and description to be more specific?

mjnitz02: It's probably ok for now. We maybe need to reconsider what should be part of the `pyproject.toml` in the future to remove some linter stuff too

PR #1686, opened 2025-10-02
[RAPTOR-14742] python311: Use system site packages for pip CVE 

nullspoon:
Pip currently has a security vulnerability, CVE-2025-8869, which has been fixed
but not released. The release date is undetermined as there are apparently
resource allocation issues on the pip maintainer team. However, chainguard has 
applied this patch on their builds of pip.

That said, when built the virtual env, we pulled in a virutalenv copy of pip,
which reintroduced the CVE to the image.

As such, we now use system site packages and do not pull in an updated pip.
That way we can will get chainguard fixes that have been applied to python
resources.

### DIFF
--- a/public_dropin_environments/python311/Dockerfile
+++ b/public_dropin_environments/python311/Dockerfile
@@ -30,6 +30,7 @@ COPY --from=build /bin/mkdir /bin/mkdir
 
 # Required for custom-models to install dependencies
 COPY --from=build /usr/bin/pip /usr/bin/pip
+COPY --from=build /usr/lib/python3.11/site-packages/pip /usr/lib/python3.11/site-packages/pip
 
 # Cleanup '__pycache__' directories. It solves an AsymmetricPrivateKey scanning error.
 COPY --from=build /usr/bin/rm /usr/bin/rm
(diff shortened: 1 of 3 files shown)


yakov-g (approved):

nullspoon:
Hey @klichukb and @yakov-g This is really just an example PR of a change I wonder if we can make.
Essentially, we have a CVE open right now and the upstream pip devs are not sure when they will have time to cut a release to fix it (the fix is already in place, we just can't get it from pypi). However, chainguard has applied the fix out of band. The docker image build process creates the virtualenv, installing its own latest pip and reintroducing the vulnerability.

This forces the use of system site packages (and system pip) so the vulnerability doesn't get reintroduced when the virtualenv is set up. Are we okay with this and/or do you two know of any use cases where this might be problematic? If it's fine still, I'd like to apply it to all of the other public dropin envs.

yakov-g:
I built the image and tried to do pip install.
It seems pip still works.

nullspoon:
Updated to drop the virtualenv completely.

nullspoon:
Alllrighty. Had a conversation with Bogdan. Looks like we can't remove the virtualenv, so this latest update goes back to using system site packages when creating the venv so we don't re-introduce the cve at build time.

PR #1688, opened 2025-10-03
[RAPTOR-14422] Drum fast fail when config file has issues

reza-mahdiani-m:
## Summary
This pull request introduces "fast fail" mechanism to the DRUM runner when failing to parse a configuration file.

The implementation includes:
- A new UnrecoverableError exception hierarchy to signal fatal errors. This is implemented for better maintenance.
- This change introduces a global `threading.excepthook` in `drum.py` to act as a final safety net. The hook catches any unhandled exception from any thread, logs a critical error with a full traceback, and immediately terminates the process with a non-zero exit code.
- A new try-except block for loading config-file and wrapping exception.

Closes: [RAPTOR-14422]

## Rationale
A model could fail to initialize correctly due to a simple typo in a config file, but the DRUM server would continue running. Users would only discover the problem after the deployment timed out or when prediction requests failed, wasting time and resources.

[RAPTOR-14422]: https://datarobot.atlassian.net/browse/RAPTOR-14422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/drum.py
+++ b/custom_model_runner/datarobot_drum/drum/drum.py
@@ -16,6 +16,7 @@ import shutil
 import subprocess
 import sys
 import tempfile
+import threading
 import time
 from pathlib import Path
 from tempfile import NamedTemporaryFile
@@ -72,6 +73,20 @@ from scipy.io import mmwrite
 
 SERVER_PIPELINE = "prediction_server_pipeline.json.j2"
 PREDICTOR_PIPELINE = "prediction_pipeline.json.j2"
+
+
+def _handle_thread_exception(args):
+    """
+    This global hook is called for any unhandled exception in any thread.
+    """
+    logging.critical(
+        f"CRITICAL: An unrecoverable error occurred in thread '{args.thread.name}': {args.exc_value}. Terminating process immediately.",
+        exc_info=(args.exc_type, args.exc_value, args.exc_traceback),
+    )
+    os._exit(1)
+
+
+threading.excepthook = _handle_thread_exception
 
 
 class CMRunner:
--- a/custom_model_runner/datarobot_drum/drum/drum.py
+++ b/custom_model_runner/datarobot_drum/drum/drum.py
@@ -16,7 +16,6 @@ import shutil
 import subprocess
 import sys
 import tempfile
-import threading
 import time
 from pathlib import Path
 from tempfile import NamedTemporaryFile
@@ -73,21 +72,6 @@ from scipy.io import mmwrite
 
 SERVER_PIPELINE = "prediction_server_pipeline.json.j2"
 PREDICTOR_PIPELINE = "prediction_pipeline.json.j2"
-
-
-def _handle_thread_exception(args):
-    """
-    This global hook is called for any unhandled exception in any thread.
-    """
-    logging.critical(
-        f"CRITICAL: An unrecoverable error occurred in thread '{args.thread.name}': {args.exc_value}. Terminating process immediately.",
-        exc_info=(args.exc_type, args.exc_value, args.exc_traceback),
-    )
-    os._exit(1)
-
-
-threading.excepthook = _handle_thread_exception
-
 
 class CMRunner:
     def __init__(self, runtime, flask_app=None, worker_ctx=None):
--- a/custom_model_runner/datarobot_drum/drum/drum.py
+++ b/custom_model_runner/datarobot_drum/drum/drum.py
@@ -73,6 +73,7 @@ from scipy.io import mmwrite
 SERVER_PIPELINE = "prediction_server_pipeline.json.j2"
 PREDICTOR_PIPELINE = "prediction_pipeline.json.j2"
 
+
 class CMRunner:
     def __init__(self, runtime, flask_app=None, worker_ctx=None):
         self.runtime = runtime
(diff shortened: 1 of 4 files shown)


custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -73,8 +75,29 @@
...
+    crashes the application, remove the 'if issubclass(...)' block and
+    always call os._exit(1) after logging.
+    """
+    # Check if the exception is a designed-for fatal error.
+    if issubclass(args.exc_type, UnrecoverableError):
+        logging.critical(
+            f"CRITICAL: An unrecoverable error occurred in thread '{args.thread.name}': {args.exc_value}. Terminating process immediately.",
+            exc_info=(args.exc_type, args.exc_value, args.exc_traceback),
+        )
+        os._exit(1)
```
yakov-g: Very cool!
I guess we can use sys.exit() now?

yakov-g:
Very cool!
Can you add tests?
I can think of the following cases:
* test vllm logic
* start thread in custom.py, throw exception make sure it is caught
* start thread in custom.py, throw exception and catch in custom.py. - make sure excepthook is not called

custom_model_runner/datarobot_drum/drum/drum.py
reza-mahdiani-m: Sounds good. Thank you for the comment.

custom_model_runner/datarobot_drum/drum/gpu_predictors/vllm_predictor.py
```diff
@@ -60,8 +61,19 @@ def download_and_serve_model(self):
...
-            config = json.loads(engine_config_file.read_text())
+            try:
+                config = json.loads(engine_config_file.read_text())
+            except Exception as e:
+                # Catch any other file-related errors
+                raise UnrecoverableConfigurationError(
+                    f"Failed to read or parse critical config file engine_config.json: {e}"
+                ) from e
             if "args" in config:
+                cli_args = config["args"]
```
adamalpi, line 72: no sure if it makes sense, if the config is not dynamically changing, you may consider some json schema validator to validate the json? Or maybe load it into a pydantic model? This way you may be catching more error earlier?

custom_model_runner/datarobot_drum/drum/drum.py
```diff
@@ -73,8 +75,29 @@
...
+
+
 class CMRunner:
     def __init__(self, runtime, flask_app=None, worker_ctx=None):
+        threading.excepthook = _handle_thread_exception
```
adamalpi: As far as I understood `threading.excepthook` is a global setting, so we don't need to assign it in the constructor. In theory, it should be okay to add it above this class? We may avoid running into side effects in the future.

adamalpi: Also, you might want to move this close to where the __main__ is? so that if anybody reads this, it will be easy to find.

adamalpi (changes requested):
Looks mostly good, I left just one comment that shall improve the logic of the code.

custom_model_runner/datarobot_drum/drum/drum.py
reza-mahdiani-m: Thank you for the comment. That makes sense. I will move that to main to make the code cleaner.

custom_model_runner/datarobot_drum/drum/drum.py
reza-mahdiani-m: I appreciate your comments and help. I agree that sys.exit() is a graceful process. I used os._exit() because we may have different crashes that may cause corrupt sate. Assume that one thread is failed while holding a like a lock. If we were to use sys.exit() it may cause deadlock instead of exiting the process.

custom_model_runner/datarobot_drum/drum/gpu_predictors/vllm_predictor.py
reza-mahdiani-m: That's a great point, thank you for the suggestion. The only point is that for this particular PR, the primary goal is to fix the "slow failure" bug by ensuring that unhandled exceptions during startup lead to an immediate crash. 
We can create a new ticket to refactor the configuration parsing with Pydantic. 
I used unrecoverable exception there to make further debugging easier. However, after receiving feedbacks I am thinking to make that more robust for different crashes. I definitely agree that it would be a better practice to catch exception earlier and fix them (maybe using some defaults) if possible

adamalpi: I honestly think handling the args properly and establishing sort of a contract is in the scope of the work you described. It will save future work from future headackes and you already have the context to make the change with low additional efforts.

yakov-g: BTW,
we could have been able to parse engine_config_file just in the init, and not having thread issues just from the very beginning.

reza-mahdiani-m: to provide some context. config file format is as the following:
```
{
  "args": [
    "--model", "meta-llama/Llama-3.1-8B-Instruct"
  ]
}
```
While testing the faulty one had just on extra comma.

adamalpi: got it!

adamalpi (approved):

custom_model_runner/datarobot_drum/drum/drum.py
reza-mahdiani-m: I moved this part of the code to `main.py `to make it more readable.

PR #1693, opened 2025-10-06
[RAPTOR-14742] PDIE: All without pip

nullspoon:
This is for all relevant public dropin environments. This drops the pip install and update that happens, preferring to rely on the system site packages and system pip. This is currently to address CVE-2025-8869, which upstream pip has fixed but has not released yet, and its release date is undetermined at this time.

Chainguard however has fixed this in their images, so pip installing from pypi is re-introducing the vulnerability into our images. Relying on system site packages, especially for pip, allows us to use the fixes from Chainguard without re-introducing them.

### DIFF
--- a/public_dropin_environments/java_codegen/Dockerfile
+++ b/public_dropin_environments/java_codegen/Dockerfile
@@ -38,6 +38,7 @@ COPY --from=build /bin/mkdir /bin/mkdir
 
 # Required for custom-models to install dependencies
 COPY --from=build /usr/bin/pip /usr/bin/pip
+COPY --from=build /usr/lib/python3.11/site-packages/pip /usr/lib/python3.11/site-packages/pip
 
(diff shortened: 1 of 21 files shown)


public_dropin_environments/java_codegen/Dockerfile
```diff
@@ -51,11 +52,9 @@ COPY requirements.txt requirements.txt
 
 ENV VIRTUAL_ENV=/opt/venv
 
-RUN sh -c "python -m venv ${VIRTUAL_ENV} && \
+RUN sh -c "/usr/bin/python -m venv --without-pip ${VIRTUAL_ENV} && \
```
peterzdeb, line 55: No chances to stop using virtualenv in the exec environments?

nullspoon: I discussed with Bogdan and he pointed out that there are some external processes that explicitly reference the virtualenv path, so sadly no, not at this time at least.  :(

PR #1954, opened 2026-03-03
[RAPTOR-16267] fix retry storm in drum sidecar for nim. 

s-gavrenkov:
in this PR. 
- Reduced timeout from 1 hour for chat completion to 10 minutes (600 seconds)
- removed internal retry (OpenAI) for chat completion, left only retry on network issue. 

## Problem description

```
DRUM with chat completion endpoint has Retry storm problem.
The issue is more apparent in Nim deployments (Gunicorn + Gevent), especially when users explicitly set an OpenAI client timeout in their Python scripts.

How the Retry Amplification Happens
1. Python Client → DRUM

 The Python client calls DRUM using the OpenAI client with:
timeout=120s
2 retries by default

This means up to 3 total attempts per request.

2. DRUM → NIM

DRUM calls NIM using its own OpenAI client with:
default_timeout=3600s
2 retries

Again, up to 3 total attempts per request.

Request Multiplication

If a request is invalid for NIM, the retry behavior stacks:
3 attempts from the Python client
3 attempts from DRUM
Result: up to 9 total requests (3 × 3) reaching NIM.



Slow NIM Scenario
If NIM is slow:

The Python client times out after 120 seconds.
It sends a new request to DRUM.
However, the original request is still being processed by NIM (due to DRUM’s 3600s timeout).
This can result in multiple concurrent requests (up to 3 instead of 1) for the same logical operation.
Capacity Impact

Each request consumes DRUM capacity (limited to 500 concurrent slots).
Connections are not released until processing completes.
Slow requests therefore occupy capacity for extended periods.
As retries accumulate, available capacity is quickly exhausted.
```

## Rationale

---

> [!NOTE]
> **Medium Risk**
> Changes request timeout and retry behavior for chat/completions, which can impact latency and failure modes under load. Risk is moderate because it alters production inference error-handling paths but is scoped to OpenAI client calls.
> 
> **Overview**
> Reduces retry amplification when DRUM proxies chat/completions to an OpenAI-compatible sidecar by **lowering the default OpenAI client timeout from 3600s to 600s** and disabling OpenAI SDK internal retries (`max_retries=0`).
> 
> Adds a `retry_on_network_errors` wrapper and applies it to chat completion calls, retrying only on connection errors with small exponential backoff (and **not** retrying on HTTP errors like 5xx/429 or on timeouts).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 52d18af9d1d57c8fde5d28d40ee341684dca1323. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


yakov-g (approved):

aremishevsky (approved):
Can't see any of CFX related changes. Approving to unblock this PR propagation...

PR #2081, opened 2026-04-21
[PRED-12653] Added custom model error

yshovkun:
## Summary
New CustomHTTPError that users can raise from custom model code to return a specific HTTP status and a clean JSON message to the customer.

## Rationale
- Added CustomHTTPError and BaseCustomUserError as a base user error class. 
Defaults: status_code=400, message "User's HTTP error in custom model".
- Possibility to applies to every hook — load_model, score, predict, transform, etc.
- Added unit, regression, and functional tests

---

> [!NOTE]
> **Medium Risk**
> Changes error propagation from user model code through the Python adapter and Flask prediction server, affecting returned HTTP status codes and response bodies for failed requests. Risk is moderate because it touches request-level error handling paths but is gated to explicit user-raised exceptions and covered by new tests.
> 
> **Overview**
> Adds a user-facing `CustomHTTPError` (via new `BaseCustomUserError`) that custom model code can raise to return a *caller-controlled* HTTP status code and a clean JSON `{ "message": ... }` response.
> 
> Updates the Python model execution path to **propagate user-raised errors unwrapped** (instead of logging/wrapping them as generic DRUM errors), including during custom task class `load()` and prediction hooks. The Flask `PredictionServer` now intercepts these user errors and returns their `status_code` rather than always falling back to `500`.
> 
> Expands unit/integration/functional coverage to validate custom-status behavior across adapter predict/load, predictor `configure()`, and end-to-end `/predict/` server responses.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dba9e1427a86b646a9b48b2bda3940fb5dd76ee5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
+++ b/custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
@@ -53,7 +53,7 @@ from datarobot_drum.drum.exceptions import (
     DrumException,
     DrumTransformException,
     DrumSerializationError,
-    ModelError,
+    CustomPredictionError,
 )
 from datarobot_drum.drum.utils.dataframe import extract_additional_columns
 from datarobot_drum.drum.utils.structured_input_read_utils import StructuredInputReadUtils
--- a/custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
+++ b/custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
@@ -315,11 +315,14 @@ class PythonModelAdapter(AbstractModelAdapter):
         DrumCommonException if model loading failed.
         """
         if self.is_custom_task_class:
-            self._custom_task_class_instance = self._custom_task_class.load(self._model_dir)
-            secrets = load_secrets(user_secrets_mount_path, user_secrets_prefix)
-            patch_outputs_to_scrub_secrets(secrets.values())
-            self._custom_task_class_instance.secrets = secrets
-            return self._custom_task_class_instance
+            try:
+                self._custom_task_class_instance = self._custom_task_class.load(self._model_dir)
+                secrets = load_secrets(user_secrets_mount_path, user_secrets_prefix)
+                patch_outputs_to_scrub_secrets(secrets.values())
+                self._custom_task_class_instance.secrets = secrets
+                return self._custom_task_class_instance
+            except Exception as exc:
+                self._log_and_raise_final_error(exc, "Failed to load custom task class.")
 
         else:
             return self._load_model_from_artifact_for_legacy_drum(skip_predictor_lookup)
--- a/custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
+++ b/custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
@@ -327,8 +327,6 @@ class PythonModelAdapter(AbstractModelAdapter):
 
         try:
             model = self._custom_hooks[CustomHooks.LOAD_MODEL](self._model_dir)
-        except ModelError:
-            raise
         except Exception as exc:
             self._log_and_raise_final_error(exc, "'load_model' hook failed to load the model.")
 
@@ -471,8 +469,6 @@ class PythonModelAdapter(AbstractModelAdapter):
         if self._custom_hooks.get(CustomHooks.READ_INPUT_DATA) and try_hook:
             try:
                 data = self._custom_hooks[CustomHooks.READ_INPUT_DATA](binary_data)
-            except ModelError:
-                raise
             except Exception as exc:
                 self._log_and_raise_final_error(
                     exc, "Model 'read_input_data' hook failed to read input data."
@@ -607,7 +603,7 @@ class PythonModelAdapter(AbstractModelAdapter):
                 predictions_df = self._custom_task_class_instance.predict(data, **kwargs)
             self._validate_data(predictions_df, CustomHooks.SCORE)
             return RawPredictResponse(predictions_df.values, predictions_df.columns)
-        except ModelError:
+        except CustomPredictionError:
             raise
         except Exception as exc:
             self._log_and_raise_final_error(exc, "Model 'score' hook failed to make predictions.")
@@ -639,7 +635,7 @@ class PythonModelAdapter(AbstractModelAdapter):
                 else:
                     # noinspection PyCallingNonCallable
                     predictions_df = score_fn(data, model, **kwargs)
-            except ModelError:
+            except CustomPredictionError:
                 raise
             except Exception as exc:
                 self._log_and_raise_final_error(
@@ -654,7 +650,7 @@ class PythonModelAdapter(AbstractModelAdapter):
         else:
             try:
                 predictions, model_labels = self._predictor_to_use.predict(data, model, **kwargs)
-            except ModelError:
+            except CustomPredictionError:
                 raise
             except Exception as exc:
                 self._log_and_raise_final_error(exc, "Failure when making predictions.")
@@ -671,7 +667,7 @@ class PythonModelAdapter(AbstractModelAdapter):
             try:
                 # noinspection PyCallingNonCallable
                 predictions_df = self._custom_hooks[CustomHooks.POST_PROCESS](predictions, model)
-            except ModelError:
+            except CustomPredictionError:
                 raise
             except Exception as exc:
                 self._log_and_raise_final_error(
--- a/custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
+++ b/custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
@@ -53,6 +53,7 @@ from datarobot_drum.drum.exceptions import (
     DrumException,
     DrumTransformException,
     DrumSerializationError,
+    ModelError,
 )
 from datarobot_drum.drum.utils.dataframe import extract_additional_columns
 from datarobot_drum.drum.utils.structured_input_read_utils import StructuredInputReadUtils
@@ -171,6 +172,11 @@ class PythonModelAdapter(AbstractModelAdapter):
             # Just log that no moderation info present
 
     def _log_and_raise_final_error(self, exc: Exception, message: str) -> NoReturn:
+        if isinstance(exc, ModelError):
+            self._logger.error(f"{message} Exception: {exc!r}")
+            # Only allow ModelError to pass through unwrapped
+            raise exc
+
         self._logger.exception(f"{message} Exception: {exc!r}")
         raise DrumPythonModelAdapterError(f"{message} Exception: {exc!r}")
 
--- a/custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
+++ b/custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
@@ -173,6 +173,7 @@ class PythonModelAdapter(AbstractModelAdapter):
 
     def _log_and_raise_final_error(self, exc: Exception, message: str) -> NoReturn:
         if isinstance(exc, BaseCustomUserError):
+            self._logger.info(f"{message} User-raised error: {exc!r}")
             raise exc
         self._logger.exception(f"{message} Exception: {exc!r}")
         raise DrumPythonModelAdapterError(f"{message} Exception: {exc!r}")
--- a/custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
+++ b/custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
@@ -172,11 +172,6 @@ class PythonModelAdapter(AbstractModelAdapter):
             # Just log that no moderation info present
 
     def _log_and_raise_final_error(self, exc: Exception, message: str) -> NoReturn:
-        if isinstance(exc, ModelError):
-            self._logger.error(f"{message} Exception: {exc!r}")
-            # Only allow ModelError to pass through unwrapped
-            raise exc
-
         self._logger.exception(f"{message} Exception: {exc!r}")
         raise DrumPythonModelAdapterError(f"{message} Exception: {exc!r}")
 
@@ -332,6 +327,8 @@ class PythonModelAdapter(AbstractModelAdapter):
 
         try:
             model = self._custom_hooks[CustomHooks.LOAD_MODEL](self._model_dir)
+        except ModelError:
+            raise
         except Exception as exc:
             self._log_and_raise_final_error(exc, "'load_model' hook failed to load the model.")
 
@@ -474,6 +471,8 @@ class PythonModelAdapter(AbstractModelAdapter):
         if self._custom_hooks.get(CustomHooks.READ_INPUT_DATA) and try_hook:
             try:
                 data = self._custom_hooks[CustomHooks.READ_INPUT_DATA](binary_data)
+            except ModelError:
+                raise
             except Exception as exc:
                 self._log_and_raise_final_error(
                     exc, "Model 'read_input_data' hook failed to read input data."
@@ -608,6 +607,8 @@ class PythonModelAdapter(AbstractModelAdapter):
                 predictions_df = self._custom_task_class_instance.predict(data, **kwargs)
             self._validate_data(predictions_df, CustomHooks.SCORE)
             return RawPredictResponse(predictions_df.values, predictions_df.columns)
+        except ModelError:
+            raise
         except Exception as exc:
             self._log_and_raise_final_error(exc, "Model 'score' hook failed to make predictions.")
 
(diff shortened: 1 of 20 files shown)


custom_model_runner/datarobot_drum/drum/exceptions.py
```diff
@@ -52,3 +52,15 @@ class UnrecoverableError(DrumException):
...
+
+class ModelError(DrumException):
+    """Raise this exception in your custom model to return a specific HTTP status code (400-499) to the user with custom message."""
+
+    def __init__(self, message: str = "User error in custom model", status_code: int = 400):
+        super().__init__(message)
+        if not (400 <= status_code <= 499):
+            raise ValueError(
+                f"ModelError status_code must be between 400 and 499, got {status_code}"
+            )
```
ikalnytskyi: So this would be converted into `500 Internal Server Error`. Is it really what we want?

custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
```diff
@@ -171,6 +172,11 @@ def _load_moderation_hooks(self, model_dir):
             # Just log that no moderation info present
 
     def _log_and_raise_final_error(self, exc: Exception, message: str) -> NoReturn:
+        if isinstance(exc, ModelError):
```
ikalnytskyi: nit: You have an absolutely different branch that shared zero code with what we used to have. Instead of making this `isinstance()` check, better catch `ModelError` where applicable.

custom_model_runner/datarobot_drum/drum/root_predictors/prediction_server.py
```diff
@@ -304,6 +307,9 @@ def stats():
         def handle_exception(e):
             logger.exception(e)
 
+            if isinstance(e, ModelError):
+                return {"message": "{}".format(e)}, getattr(e, "status_code", 400)
```
ikalnytskyi: nit: `e.status_code`. You already checked that it's `ModelError`, and we know that `ModelError` has `.status_code`. By calling `getattr` you're communicating that this property may or may not exist in `ModelError`, and this is not the case.

ikalnytskyi: nit: `str(e)` is more idiomatic then going through format.

custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
```diff
@@ -171,6 +172,11 @@ def _load_moderation_hooks(self, model_dir):
...
+            self._logger.error(f"{message} Exception: {exc!r}")
+            # Only allow ModelError to pass through unwrapped
```
ikalnytskyi: nit: This is a zero-value code comment. It literally duplicates what people see before their eyes: `if isnstance(exc, ModelError)` -> `raise exc`.

custom_model_runner/datarobot_drum/drum/runtime.py
```diff
@@ -108,6 +111,9 @@ def health():
     @model_api.route("/predictionsUnstructured/", methods=["POST"])
     @model_api.route("/invocations", methods=["POST"])
     def predict():
+        if isinstance(exc_value, ModelError):
+            return {"message": "{}".format(exc_value)}, getattr(exc_value, "status_code", 400)
```
ikalnytskyi: Ditto on `getattr`.

ikalnytskyi: What's the point of this change? it seems this always returns a hardcoded error. I don't think you need to touch it.

custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
yshovkun: Just to clarify.
Leave this method as it is, but instead try to catch `ModelError` in prediction methods right?
Like:
```
  except ModelError:
      raise
  except Exception as exc:
      self._log_and_raise_final_error(
          exc, "Model 'score' hook failed to make predictions."
      )
```

custom_model_runner/datarobot_drum/drum/exceptions.py
yshovkun: I was thinking to block user from triggering Internal Server Errors in their code. Since it will affect our monitoring and will rise alerts, in case someone will spam 500 in custom model code.
But you are right, it will spam 500 error anyway.
So would it be ok to raise custom ModelError instead with such error message (status could be between 400 and 499)?

yshovkun: thoughts
```
@model_api.errorhandler(Exception)
def handle_exception(e):
    logger.exception(e)

    if isinstance(e, ModelError):
        status_code = e.status_code
        if not (400 <= status_code <= 499):
            logger.warning(f"Invalid ModelError status_code {status_code}. Forcing 422.")
            status_code = 422
        return {"message": str(e)}, status_code

    if isinstance(e, HTTPException) and e.code == HTTP_400_BAD_REQUEST:
        return jsonify(error=e.description), e.code

    return {"message": "ERROR: {}".format(e)}, HTTP_500_INTERNAL_SERVER_ERROR
```

custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
ikalnytskyi: Yes, correct.

custom_model_runner/datarobot_drum/drum/exceptions.py
ikalnytskyi: I don't think I'm the right person to ask. I just mentioned a problem in the code. I still think that this kind of protection is over-engineering. Nobody in their mind is going to explicitly return 500. And if they really want to, they would just raise an unknown exception and drum would return 500 anyway. I'd just let the status code to be passed through with not extra checks.

ikalnytskyi: but I'm neither the one who gave you the task nor drum maintainer

custom_model_runner/datarobot_drum/drum/adapters/model_adapters/python_model_adapter.py
```diff
@@ -468,6 +471,8 @@ def load_data(self, binary_data, mimetype, try_hook=True, sparse_colnames=None):
         if self._custom_hooks.get(CustomHooks.READ_INPUT_DATA) and try_hook:
             try:
                 data = self._custom_hooks[CustomHooks.READ_INPUT_DATA](binary_data)
+            except ModelError:
+                raise
```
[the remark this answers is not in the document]
yshovkun: We need this for prediction methods only

custom_model_runner/datarobot_drum/drum/exceptions.py
```diff
@@ -52,3 +52,11 @@ class UnrecoverableError(DrumException):
...
 class UnrecoverableConfigurationError(UnrecoverableError):
     """Raised when failure in parsing or validating configuration file."""
+
+
+class ModelError(DrumException):
+    """Raise this exception in your custom model to return a specific HTTP status code (400-499) to the user with custom message."""
+
+    def __init__(self, message: str = "User error in custom model", status_code: int = 400):
+        super().__init__(message)
+        self.status_code = status_code
```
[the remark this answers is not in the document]
yshovkun: Fixed

custom_model_runner/datarobot_drum/drum/runtime.py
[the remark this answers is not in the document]
arturops: I think this is relevant to the PR because the `ModelError` with custom code seems to be wrapped around a catch it all and it will never be seen by users if they expect it

custom_model_runner/datarobot_drum/drum/root_predictors/prediction_server.py
```diff
@@ -304,6 +307,9 @@ def stats():
         def handle_exception(e):
             logger.exception(e)
 
+            if isinstance(e, ModelError):
+                return {"message": str(e)}, e.status_code
+
```
arturops, line 329: I do not know how drum works, but why can't we just do this check here, I see the `@model_api.errorhandler(Exception)` and the method is `handle_exception` so we probably can check here the exception instead an remove all other `except ModelError` and also the `except Exception` so that it all comes here and only one place needs change instead of all other places.

Something similar for the `runtime.py`. 
My point being `Exception` is to general and that is why the need of adding `except ModelError: raise`, so maybe this needs to have a `CustomException` instead of `except Exception` to handle most of expected exceptions and let pass `ModelError` so you can catch it here and avoid re-raising everywhere else.

Once some new code gets added by someone unaware of ModelError , meaning the `except ModelError` is not added, you will get a bug or customer bug because the `ModelError` was not raised

custom_model_runner/datarobot_drum/drum/runtime.py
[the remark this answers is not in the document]
yshovkun: @arturops code is outdated a bit.
I was thinking about adding ModelError specificaly for prediction methods only. Do you think we should allow this for other method like load, etc?

custom_model_runner/datarobot_drum/drum/root_predictors/prediction_server.py
yshovkun: First I added `ModelError` check inside `_log_and_raise_final_error` methods, but then we decided separately itiniate it for each method call separately and leave general `rumPythonModelAdapterError` as it is now

custom_model_runner/datarobot_drum/drum/exceptions.py
yshovkun: I removed limitation by statuses. 
User can add any status to their custom errors

custom_model_runner/datarobot_drum/drum/root_predictors/prediction_server.py
yshovkun: Fixed

custom_model_runner/datarobot_drum/drum/runtime.py
yshovkun: Code was modified
```
    def predict():
        if isinstance(exc_value, ModelError):
            return {"message": str(exc_value)}, exc_value.status_code

        return {"message": "ERROR: {}".format(exc_value)}, HTTP_513_DRUM_PIPELINE_ERROR
```
So I just check if this is Error from model and propagete it with status and message instead of hardcoded HTTP_513_DRUM_PIPELINE_ERROR

custom_model_runner/datarobot_drum/drum/root_predictors/prediction_server.py
piratovskyi: Why did you decide to not use _log_and_raise_final_error for this purpose?

custom_model_runner/datarobot_drum/drum/exceptions.py
```diff
@@ -52,3 +52,11 @@ class UnrecoverableError(DrumException):
 
 class UnrecoverableConfigurationError(UnrecoverableError):
     """Raised when failure in parsing or validating configuration file."""
+
+
+class ModelError(DrumException):
```
piratovskyi: I'd name it a bit different because ModelError sounds too generic. What about CustomHttpException or something similar?

tests/unit/datarobot_drum/drum/adapters/model_adapters/test_python_model_adapter.py
```diff
@@ -373,6 +379,27 @@ def test_multiple_artifacts_detection_negative(self):
             with pytest.raises(DrumCommonException, match="Multiple serialized model files found."):
                 adapter._detect_model_artifact_file()
 
+    def test_predict_new_drum_custom_status_code(self):
```
piratovskyi, line 410: I'd say you need to add tests for each hook you've changed with this exeception

custom_model_runner/datarobot_drum/drum/root_predictors/prediction_server.py
yshovkun: We want to add this to prediction methods only, so that user will see their custom response. And I saw that I added raise of ModelError to load model method by mistake. Removed.

custom_model_runner/datarobot_drum/drum/exceptions.py
yshovkun: Changed this to `CustomPredictionError`

custom_model_runner/datarobot_drum/drum/runtime.py
yshovkun: After investigation, I drop these changes.

yshovkun:
jarvis test base_checks-test_unit

tests/unit/datarobot_drum/drum/adapters/model_adapters/test_python_model_adapter.py
yshovkun: Added tests

custom_model_runner/datarobot_drum/drum/root_predictors/prediction_server.py
arturops: Ok with that change, if the `ModelError` is only for predictions, the question is where in the code the users want to raise this exception and for what reasons??  or are we saying every exception will become a `ModelError` if it happens within the model predict method?

ikalnytskyi (approved):

custom_model_runner/datarobot_drum/drum/root_predictors/prediction_server.py
arturops: I can share an example of my suggestion about using a custom exception and the general Exception. In the example code, we have an `Exception` to catch everything and send a webhook no matter what when we fail. And our custom exception `UmapJobError` that is the parent of our custom exceptions in the job, this way we can catch those custom exceptions and do our custom logic like sending webhook, sending uI a user firnedly error message, etc.
https://github.com/datarobot/insights-service/blob/e06ee0de4a9eee1b76fb01a3cd087de61023338d/insights_service/insights/umap/job.py#L316-L341
The key is to know where you want to raise your exceptions, and in this case you might need a custom exception like `CustomerUserError...` and inherit it to `ModelError` then you can raise ModelError wherever you need to and let it propagate all the way to your handler location. 

If you have `except Exception` in the middle of that bubble up `ModelError` exception you need to refactor the code so they become a `DataRobotPredictionError...` so your `ModelError` doesn't get catch and you can actually handle `except Exception` in one single place where all things are catch. 
This gives you a framework to keep adding more custom exception inheritted from `DataRobotpredictionError...` or `CustomUsererror...` and catching unexpected Exception if you must otherwise ignore/remove `except Exception` (too broad) if you need the code to fail spectacular when unexpected things happen which is usually desired instead of defensive programming. If that will bring the predictions erver down or custom model lrs then sure you can keep it to avoid service disruption.

Again, not sure how this really works, but bubbling an exception adding `except ... raise`  will be hard to maintain (saying it from painful experience)

arturops (approved):
Looks good, thanks for cleaning some things :) and left some future work you could do if you want to minimize adding `except CustomPredictionError` 👍

custom_model_runner/datarobot_drum/drum/exceptions.py
```diff
@@ -52,3 +52,15 @@ class UnrecoverableError(DrumException):
...
+
+
+class BaseCustomUserError(Exception):
+    """Base class for errors that originate in user-provided custom model code."""
+
+
+class CustomHTTPError(BaseCustomUserError):
+    """Raise this exception in your custom model to return a specific HTTP status code with custom message."""
+
+    def __init__(self, message: str = "User's HTTP error in custom model", status_code: int = 422):
```
piratovskyi: nit: please use `HTTP_422_UNPROCESSABLE_ENTITY` instead of hardcoded 422

custom_model_runner/datarobot_drum/drum/server.py
```diff
@@ -28,6 +28,7 @@
 
 HEADER_REQUEST_ID = "X_Request_ID"
 HEADER_DRUM_VERSION = "X-Drum-Version"
+HEADER_DRUM_USER_ERROR = "X-Drum-User-Error"
```
piratovskyi: How is this gonna be used later on?

piratovskyi:
Looks like PR description is different from what PR implements. Please, update

yshovkun:
> Looks like PR description is different from what PR implements. Please, update

Thanks for note. I updated this accotdingly.

custom_model_runner/datarobot_drum/drum/server.py
piratovskyi: Resolved in DM

piratovskyi (approved):

custom_model_runner/datarobot_drum/drum/server.py
yshovkun: We will have controll for these errors on prediction side.
Like here we do not retry them:
https://github.com/datarobot/DataRobot/blob/ef4cb16fbc87d6b8126960a5dfe803a31c43738f/raptor_lake/k8s/service_request_utils.py#L195

custom_model_runner/datarobot_drum/drum/exceptions.py
yshovkun: Fixed

PR #2137, opened 2026-05-20
[MODEL-23532] Add dr_mcp_execute_sandbox_minimal sandbox runner env

j1z0:
## Summary

Adds a minimal Python 3.12 container image used by the `execute_code` MCP tool in
[datarobot-oss/datarobot-genai](https://github.com/datarobot-oss/datarobot-genai)
(`src/datarobot_genai/drtools/sandbox/`). The image is submitted to the DataRobot
Workload API as a single-container, short-lived sandbox for executing untrusted
Python code on behalf of MCP clients.

Unlike the rest of `public_dropin_environments/`, this is **not** a custom-model
drop-in env and is **not** registered in the DataRobot environment catalog — it's
purely a runtime image consumed by genai-side sandbox tooling.

## What's in here

- `public_dropin_environments/dr_mcp_execute_sandbox_minimal/`
  - `Dockerfile` — `python:3.12-slim` + `polars`/`pyarrow`/`datarobot`/`requests`/`httpx`, runs as UID 65534
  - `runner.py` — standalone runner that decodes `DR_SANDBOX_CODE_B64`, executes it, prints `__DR_SANDBOX_RESULT__:<json>` for the caller to parse
  - `README.md` — points back to the genai-side source of truth
- `.harness/dr_mcp_execute_sandbox_minimal_image_build_default_pr_input.yaml` — master-push image-build input set (publishes `_latest`)
- `.harness/dr_mcp_execute_sandbox_minimal_image_build_pr_input.yaml` — PR-build input set (publishes `pr-<N>` for testing)
- `.harness/dr_mcp_execute_sandbox_minimal_on_pr.yaml` — webhook trigger: builds on PR open/sync against this folder
- `.harness/dr_mcp_execute_sandbox_minimal_on_push_master.yaml` — webhook trigger: publishes `_latest` on master push

## Published image (after merge)

```
datarobotdev/datarobot-user-models:public_dropin_environments_dr_mcp_execute_sandbox_minimal_latest
```

PR builds produce `datarobotdev/datarobot-user-models:pr-<PR-number>`.

## Tracking

- **Jira**: [MODEL-23532](https://datarobot.atlassian.net/browse/MODEL-23532) — child of epic [MODEL-22666](https://datarobot.atlassian.net/browse/MODEL-22666) (Use Panels as MCP Resources & Tools Enhancement)
- **Consumer (genai)**: datarobot-oss/datarobot-genai#350 — updates `DEFAULT_SANDBOX_IMAGE` to the URI above. Both PRs should merge together.
- **Related Jira**: [MODEL-23530](https://datarobot.atlassian.net/browse/MODEL-23530) (MCP_SANDBOX entitlement gating), [MODEL-23531](https://datarobot.atlassian.net/browse/MODEL-23531) (workload-api securityContext gating)

## Test plan

- [ ] PR-open Harness trigger fires and builds `pr-<N>` image successfully
- [ ] `docker pull datarobotdev/datarobot-user-models:pr-<N>` works
- [ ] Local end-to-end run of `scripts/demo_sandbox_e2e.py` (genai side) against a DR endpoint succeeds using the `pr-<N>` image
- [ ] After merge, master-push trigger publishes `_latest` tag

[MODEL-23532]: https://datarobot.atlassian.net/browse/MODEL-23532


public_dropin_environments/dr_mcp_execute_sandbox_minimal/runner.py
```diff
@@ -0,0 +1,68 @@
...
+    if not code_b64:
+        print("DR_SANDBOX_CODE_B64 not set", file=sys.stderr)
+        return 1
+    code = base64.b64decode(code_b64).decode("utf-8")
+    inputs = _decode_env("DR_SANDBOX_INPUTS_B64", "{}")
+
+    namespace: dict[str, Any] = {"inputs": inputs, "_return": None}
+    try:
+        exec(compile(code, "<sandbox>", "exec"), namespace)  # noqa: S102
+    except Exception:
```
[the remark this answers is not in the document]
j1z0: Fixed in d5b8c227. The runner now catches BaseException around exec() — SystemExit is re-mapped to its exit code, other BaseException subclasses (including KeyboardInterrupt) go to exit code 1, and the result marker is emitted in every path so the caller's output contract is never silently broken.

mjnitz02 (approved):
The harness related changes I'm not super familiar with, but the `dr_mcp` stuff looks good. I wonder if, to speed things up, if its worth doing some sort of `pyproject.toml` synchronization between the MCP repo and the minimal sandbox image? That might be outside the scope, but then you'd have less "building" during a live deployment.

.harness/dr_mcp_execute_sandbox_minimal_on_pr.yaml
mjnitz02, line 1: Do you want to create/build images when the PR is opened?

tools/create-and-source-venv.sh
```diff
@@ -15,6 +15,9 @@ echo "== Creating and sourcing a new virtual environment for the tests: '$tmp_ve
 rm -rf $tmp_venv_path
 python3 -m venv $tmp_venv_path
 . ${tmp_venv_path}/bin/activate
-pip install -U pip
-pip install -U setuptools wheel
+# pip 26.1.1 + setuptools>=82 + wheel>=0.47 introduce regressions in
+# datarobot-drum's wheel build (dist-info collision in setuptools 82, and
+# pip 26's ephemeral wheel cache fails cleanup with "Directory not empty").
+# Pin to last-known-good majors until upstream fixes propagate.
+pip install -U 'pip<26' 'setuptools<82' 'wheel<0.47'
```
rabih-datarobot, line 22: This looks like an existing code, not related to this pr, is it a tactical fix for CI to unblock this PR?

public_dropin_environments/dr_mcp_execute_sandbox_minimal/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Copyright 2026 DataRobot, Inc.
+#
+# Minimal sandbox runner image used by DataRobotWorkloadSandbox.
+#
+# Ships a standalone runner script (no datarobot-genai install needed
+# inside the container) plus the data libraries user code commonly needs.
+
+FROM python:3.12-slim
```
rabih-datarobot: Are we only going to support python env? would it be better to make the sandbox generic and run the code in a child process instead of runner interpreter?

public_dropin_environments/dr_mcp_execute_sandbox_minimal/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+RUN pip install --no-cache-dir \
+    "polars>=1.0.0,<2.0.0" \
+    "pyarrow>=21.0.0,<22.0.0" \
+    "datarobot>=3.10.0,<4.0.0" \
```
rabih-datarobot: Do we want to restrict ourselves to certain datarobot release versions?

rabih-datarobot (commented):
Overall looks good and minimal. 
I left few question, none of them is a blocker, would like some clarification before approval.
It would be great to enforce timeouts and use a child processes instead if possible. 
This runner is not a sandbox security boundary. It relies on the container restrictions to make that safe. The security boundary is the workload execution environment. @j1z0 could you please share the related pr in the workload api for this work?

tools/create-and-source-venv.sh
j1z0: This is a global problem with this repo. New version of the packages just came in a couple of days ago and broke all the builds for this repo.  This fixes those version forcing the compatible ones.

public_dropin_environments/dr_mcp_execute_sandbox_minimal/Dockerfile
j1z0: Sorry not entirely sure what you mean here.  But I think we need to create a separate image for each Sandboxed environment we want to support.  This one is meant to be the minimal one that supports the filtering requirements and basic python coding needed for the panels execution....   Is there a better way to manage / organize these images?

public_dropin_environments/dr_mcp_execute_sandbox_minimal/Dockerfile
j1z0: We need some minimal version to support, but maybe not a maximum version.  Are you suggesting we remove the maximum version or rather pin datarobot to the current released version?

j1z0:
related PR is here: https://github.com/datarobot-oss/datarobot-genai/pull/350

.harness/dr_mcp_execute_sandbox_minimal_on_pr.yaml
j1z0: Keeping `on_pr` to match the convention of the other dropin envs (e.g. `python3_keras_local_on_pr.yaml`). Note the trigger is new in this PR so it won't actually fire here — it'll start running on PRs after this merges.

public_dropin_environments/dr_mcp_execute_sandbox_minimal/Dockerfile
mjnitz02: I think in a lot of the genai packages we stay within the minor version in case there are breaking changes in a more version change. So like `"datarobot>=3.10.0,<3.11.0"`?

public_dropin_environments/dr_mcp_execute_sandbox_minimal/Dockerfile
rabih-datarobot: I was under the impression that this is targeting a generic sandbox but given its only for panels for now, that will do, and yes we should discuss that when we start expand on the MCP sandbox env overall

public_dropin_environments/dr_mcp_execute_sandbox_minimal/Dockerfile
j1z0: Done in f943a9b1 — pinned to `datarobot>=3.10.0,<3.11.0`.

j1z0:
Thanks @rabih-datarobot — addressed both:

- **CODEOWNERS**: added `.github/CODEOWNERS` in d0d2b6e1 pointing this directory at `@datarobot/mcp`. (No CODEOWNERS file existed in the repo before; this only covers our new directory and leaves the rest of the repo's ownership unchanged.)
- **Execution timeout**: added in b5e5ff88. `runner.py` now wraps the user-code `exec()` with a SIGALRM-based wall-clock, defaulting to **1 hour** (configurable via `DR_SANDBOX_TIMEOUT_SECS`). On timeout the runner exits with code 124 and still emits the `__DR_SANDBOX_RESULT__:null` marker so callers don't hang. The cap is called out prominently in the Dockerfile header and the README's "Execution timeout" section. Worth noting: this is defense-in-depth for accidental hangs (infinite loops, runaway ops) and not a security boundary — malicious code can reset the SIGALRM handler. The real wall-clock enforcement still belongs caller-side (`DataRobotWorkloadSandbox` / Workload API), and I'll wire the matching caller-side cap in [datarobot-oss/datarobot-genai#350](https://github.com/datarobot-oss/datarobot-genai/pull/350).

PR #2142, opened 2026-05-25
[BUZZOK-30877] Use CUSTOM_MODEL_WORKERS for dragent to set gunicorn workers

tsdaemon:

## Summary

`dragent` was set up to run as a single worker which is really not the way to do it in production. This enables `gunicorn`, and sets workers to `CUSTOM_MODEL_WORKERS`, same as for `drum`

## Rationale

---

> [!NOTE]
> **Medium Risk**
> Changes production entrypoint and process model (new NAT CLI, multi-worker Gunicorn); misconfigured worker counts could affect capacity or resource use in deployments.
> 
> **Overview**
> The **python311_genai_agents** startup script now launches the NAT dragent API with **`nat dragent serve`** (replacing **`nat start dragent_fastapi`**) and runs it behind **Gunicorn** with a configurable worker count.
> 
> When **`ENABLE_DRAGENT_SERVER`** is enabled, the script reads **`CUSTOM_MODEL_WORKERS`** from MLOps runtime parameters (default **1**) and passes **`--use_gunicorn true`** and **`--workers`** to the serve command, matching how other drop-in environments expose Gunicorn worker settings.
> 
> **`uv sync`** is allowed to fail without aborting startup (**`|| true`**), so a dependency sync error no longer blocks the server from starting.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d34e2e2212e9802706c39e470f425cbe9e95ee44. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/public_dropin_environments/python311_genai_agents/env_info.json
+++ b/public_dropin_environments/python311_genai_agents/env_info.json
@@ -4,7 +4,7 @@
   "description": "This template environment can be used to create GenAI-powered agents using CrewAI, LangGraph, or Llama-Index. Similar to other drop-in environments, you can either include a .pth artifact or any other code needed to deserialize your model, and optionally a custom.py file. You can also use this environment in codespaces.",
   "programmingLanguage": "python",
   "label": "",
-  "environmentVersionId": "6a0eefa1f2ab47076d4cf4c3",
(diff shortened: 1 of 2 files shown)


rmidianyi (approved):
lgtm

PR #2146, opened 2026-05-28
[RAPTOR-17573] moving java dependencies into optional extras in DRUM

peterzdeb:

## Summary

Make py4j and Log4j JARs optional in datarobot-drum by introducing a `[java]` extra, and strip the JARs from non-Java dropin environment images at build time.

## Rationale

`datarobot-drum` has **two independent Java CVE sources** that affect every dropin environment, even those with no Java runtime:

| CVE source | Where it lives | Fixed by |
|---|---|---|
| `py4j` Python package | `requirements.txt` core dep — installed in every env | Moved to `extras_require["java"]` |
| Log4j-bearing JARs (`drum-py4j-entrypoint.jar`, `drum-predictors.jar`) | Bundled in the wheel via `setup.py` glob | Deleted from venv in non-Java Dockerfiles |

Both sources need to be addressed independently:
- Removing py4j from core deps eliminates py4j CVE hits on Python-only images and prevents future py4j vulnerabilities from re-appearing across the fleet.
- Stripping the JARs at Docker build time removes Log4j from non-Java images.

Only `java_codegen` legitimately needs both — it now declares `datarobot-drum[java]` to retain py4j, and its Dockerfile is not patched so the JARs remain.

Full design context: [DRUM + Java dependency maintenance](https://datarobot.atlassian.net/wiki/spaces/RAPTOR/pages/7691436035/DRUM+Java+dependency+maintenance)
Jira: [RAPTOR-17573](https://datarobot.atlassian.net/browse/RAPTOR-17573)

## Changes

**`custom_model_runner/`**
- `requirements.txt` — removed `py4j` as a core dep
- `setup.py` — added `extras_require["java"] = ["py4j~=0.10.9.0"]`
- `java_predictor.py` — guarded py4j imports with `try/except ImportError`; `configure()` raises `DrumCommonException` if `[java]` extra is not installed
- `description.py` — bumped to 1.17.18

**`java_codegen` environment**
- `requirements.in` — `datarobot-drum` → `datarobot-drum[java]`

**Non-Java Dockerfiles** (`python3_sklearn`, `python3_keras`, `python3_onnx`, `python3_pytorch`, `python3_xgboost`, `python312`, `r_lang`)
- Added `find ${VIRTUAL_ENV} -path '*/datarobot_drum/*.jar' -delete` after pip install

## Related CVE tickets

RAPTOR-16822, RAPTOR-16824, RAPTOR-16825, RAPTOR-16826, RAPTOR-16827, RAPTOR-16828, RAPTOR-16829, RAPTOR-16830, RAPTOR-16831

[RAPTOR-17573]: https://datarobot.atlassian.net/browse/RAPTOR-17573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shopin32 (commented):
I would expect that py4j imports don't cause any issues unless you try to start a java process, so I wonder why do we want to make py4j an optional dependency

nullspoon (approved):
🥲

peterzdeb:
> I would expect that py4j imports don't cause any issues unless you try to start a java process, so I wonder why do we want to make py4j an optional dependency

@shopin32 the idea is to reduce number of envs with CVEs related to Java dependencies - in the images where we don't even install openjdk

shopin32:
> > I would expect that py4j imports don't cause any issues unless you try to start a java process, so I wonder why do we want to make py4j an optional dependency
> 
> @shopin32 the idea is to reduce number of envs with CVEs related to Java dependencies - in the images where we don't even install openjdk

@peterzdeb but from what I've understood CVEs are not inside py4j.jar that is installed with py4j package, but from our jars that are built in this project... so I would expect creating a separate package with those jars and opting out it, not py4j itself.

peterzdeb:
> > > I would expect that py4j imports don't cause any issues unless you try to start a java process, so I wonder why do we want to make py4j an optional dependency
> > 
> > 
> > @shopin32 the idea is to reduce number of envs with CVEs related to Java dependencies - in the images where we don't even install openjdk
> 
> @peterzdeb but from what I've understood CVEs are not inside py4j.jar that is installed with py4j package, but from our jars that are built in this project... so I would expect creating a separate package with those jars and opting out it, not py4j itself.

That's right, this PR doesn't resolve the CVE itself. It's just reduces noise from other envs, future proof. Next step for me is to look for the CVE fix

shopin32:
> > > > I would expect that py4j imports don't cause any issues unless you try to start a java process, so I wonder why do we want to make py4j an optional dependency
> > > 
> > > 
> > > @shopin32 the idea is to reduce number of envs with CVEs related to Java dependencies - in the images where we don't even install openjdk
> > 
> > 
> > @peterzdeb but from what I've understood CVEs are not inside py4j.jar that is installed with py4j package, but from our jars that are built in this project... so I would expect creating a separate package with those jars and opting out it, not py4j itself.
> 
> That's right, this PR doesn't resolve the CVE itself. It's just reduces noise from other envs, future proof. Next step for me is to look for the CVE fix

I just have a feeling that introducing this `datarobot-drum[java]` in this PR doesn't make sense as it still non functional, as it is implicitly hints you that having a regular environment you can do pip install datarobot-drum[java] and it will enable using scoring code jars but it is not the case. 

I would limit this PR to cleaning up those .jars only in Dockerfile

mariusvilkas (approved):
I do not think there are files to review for core modeling.
LGTM

amarmudrankit:
> > > > > I would expect that py4j imports don't cause any issues unless you try to start a java process, so I wonder why do we want to make py4j an optional dependency
> > > > 
> > > > 
> > > > @shopin32 the idea is to reduce number of envs with CVEs related to Java dependencies - in the images where we don't even install openjdk
> > > 
> > > 
> > > @peterzdeb but from what I've understood CVEs are not inside py4j.jar that is installed with py4j package, but from our jars that are built in this project... so I would expect creating a separate package with those jars and opting out it, not py4j itself.
> > 
> > 
> > That's right, this PR doesn't resolve the CVE itself. It's just reduces noise from other envs, future proof. Next step for me is to look for the CVE fix
> 
> I just have a feeling that introducing this `datarobot-drum[java]` in this PR doesn't make sense as it still non functional, as it is implicitly hints you that having a regular environment you can do pip install datarobot-drum[java] and it will enable using scoring code jars but it is not the case.
> 
> I would limit this PR to cleaning up those .jars only in Dockerfile

I agree with @shopin32 here — datarobot-drum[java] only gates py4j itself, not the JARs that are causing the CVEs.  We should scope this PR to just the Dockerfile cleanup — deleting the JARs that ship with the base datarobot-drum package from non-Java images. 

The `[java]` extra could come back as a follow-up if we decide to restructure JAR distribution, but that's a bigger design question and probably not needed right now.

peterzdeb:
Thanks for the feedback, @amarmudrankit @shopin32 . To clarify the scope — the Jira ticket RAPTOR-17573 identifies two separate root causes:
1. JAR files (drum-py4j-entrypoint.jar, drum-predictors.jar) embedding Log4j → addressed by the Dockerfile cleanup
2. py4j Python package itself → also generates CVE hits on non-Java images

Making py4j optional via [java] ensures that future py4j CVEs don't affect Python-only environments. Without it, every new py4j vulnerability would again require manual suppression across the whole fleet.

The [java] extra isn't misleading — it correctly gates py4j for environments that actually need Java. The JARs being absent (from the Dockerfile find ... -delete) makes the extra functionally consistent: datarobot-drum[java] + a Java runtime is what java_codegen uses; nothing else should have either.

Happy to discuss further, but the design intent covers both sources of CVE noise, not just the JARs.

custom_model_runner/datarobot_drum/drum/language_predictors/java_predictor/java_predictor.py
```diff
@@ -74,6 +79,10 @@ def __init__(self):
         )
 
     def configure(self, params):
+        if py4j is None:
+            raise DrumCommonException(
+                "Java support requires the 'java' extra: pip install 'datarobot-drum[java]'"
+            )
```
shopin32, line 85: If we want to address two different situations, it would be nice then to distinguish them here. 

Scenarios: 
1. if there are all jars in place, then this exception makes sense.
2. If there are no jars we should print something like: "Wrong env, please use java based env. "

peterzdeb: added this as a separate message

shopin32 (approved):

PR #2154, opened 2026-06-03
[RAPTOR-17634] Make gunicorn the default server in drum

piratovskyi:
# RATIONALE
Gunicorn is a production-grade WSGI server and should be the default when running drum in server mode, rather than requiring an explicit opt-in via the `DRUM_SERVER_TYPE` environment variable. This improves out-of-the-box production readiness while still allowing users to explicitly select `werkzeug` when needed.

## CHANGES
- `entry_point.py`: Inverted the server type selection logic — gunicorn is now the default when `DRUM_SERVER_TYPE` is not set. If an unsupported server type is provided, a warning is logged and gunicorn is used as the fallback.
- `enum.py`: Added `DrumServerType` class with constants (`GUNICORN`, `WERKZEUG`, `ALL`, `DEFAULT`) to avoid magic strings in the server selection logic.
- `tests/unit/datarobot_drum/drum/test_entry_point.py`: Added 6 unit tests covering all server type selection scenarios: default (no env var), explicit gunicorn, explicit werkzeug, unsupported type fallback with warning, and non-server mode.

---

> [!NOTE]
> **Medium Risk**
> Changing the default HTTP server affects all `drum server` deployments without `DRUM_SERVER_TYPE` set; mitigations exist (env var, docker fallback, warnings) but behavior and multi-process error handling differ from Werkzeug.
> 
> **Overview**
> **`drum server` now defaults to Gunicorn** instead of Werkzeug when `DRUM_SERVER_TYPE` is unset. Users can set `DRUM_SERVER_TYPE=werkzeug` for the old behavior; a startup warning explains the change. Invalid server types log a warning and fall back to Gunicorn. **`--docker` always uses Werkzeug** because Gunicorn’s host bind conflicts with container port publishing.
> 
> Server selection is centralized via new **`DrumServerType`** constants in `enum.py`, with **`config_logging()`** at entry and again in `main()` for Gunicorn worker processes.
> 
> **Gunicorn-specific fixes:** startup failures with `--with-error-server` no longer tear down the master when a Flask app is already serving (`runtime.py`). Perf tests kill the whole process group so Gunicorn workers don’t survive after the master dies (`perf_testing.py`).
> 
> Release **1.17.19** is documented in the changelog; unit tests cover default, explicit, unsupported, docker override, and non-server modes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fbfe4e8d6d4684bd54eee208a0f77e2f3e38dd34. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/entry_point.py
+++ b/custom_model_runner/datarobot_drum/drum/entry_point.py
@@ -33,6 +33,16 @@ def run_drum_server():
                 DrumServerType.GUNICORN,
                 DrumServerType.WERKZEUG,
             )
+
+        if server_type == DrumServerType.GUNICORN and getattr(options, "docker", None):
+            logger.warning(
+                "Gunicorn is not supported with --docker: gunicorn binds the host address "
+                "before the container starts, which conflicts with the container's own port "
+                "publish. Falling back to %s.",
+                DrumServerType.WERKZEUG,
+            )
+            server_type = DrumServerType.WERKZEUG
+
         if server_type == DrumServerType.GUNICORN:
             main_gunicorn()
         else:
--- a/custom_model_runner/datarobot_drum/drum/entry_point.py
+++ b/custom_model_runner/datarobot_drum/drum/entry_point.py
@@ -1,14 +1,16 @@
-import logging
-
+from datarobot_drum.drum.common import config_logging, get_drum_logger
 from datarobot_drum.drum.gunicorn.run_gunicorn import main_gunicorn
 from datarobot_drum.drum.main import main
 from datarobot_drum import RuntimeParameters
 from datarobot_drum.drum.utils.setup import setup_options
 
-from datarobot_drum.drum.enum import ArgumentsOptions, DrumServerType
+from datarobot_drum.drum.enum import ArgumentsOptions, DrumServerType, LOGGER_NAME_PREFIX
+
+logger = get_drum_logger(LOGGER_NAME_PREFIX)
 
 
 def run_drum_server():
+    config_logging()
     options = setup_options()
     if options.subparser_name == ArgumentsOptions.SERVER:
         server_type = DrumServerType.DEFAULT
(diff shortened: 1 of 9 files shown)


piratovskyi:
@cursor review

GunnerGlory (approved):

custom_model_runner/datarobot_drum/drum/entry_point.py
```diff
@@ -1,19 +1,32 @@
+import logging
```
peterzdeb: Would it require a logger setup first, before producing any logs?

custom_model_runner/datarobot_drum/drum/entry_point.py
```diff
@@ -1,19 +1,32 @@
...
     options = setup_options()
-    if (
-        options.subparser_name == ArgumentsOptions.SERVER
-        and RuntimeParameters.has("DRUM_SERVER_TYPE")
-        and str(RuntimeParameters.get("DRUM_SERVER_TYPE")).lower() == "gunicorn"
-    ):
-        main_gunicorn()
+    if options.subparser_name == ArgumentsOptions.SERVER:
+        server_type = DrumServerType.DEFAULT
+        if RuntimeParameters.has("DRUM_SERVER_TYPE"):
```
peterzdeb, line 17: Any deployment that relied on the implicit default (werkzeug) without setting DRUM_SERVER_TYPE will silently switch to gunicorn. The PR description acknowledges this as medium risk. A migration note or deprecation warning for a version would be safer

peterzdeb (approved):
few comments to warn and reconsider. But looks good overall

custom_model_runner/datarobot_drum/drum/entry_point.py
piratovskyi: Fair point, I can add a warning message about the change

custom_model_runner/datarobot_drum/drum/entry_point.py
piratovskyi: yes, you are right

piratovskyi:
@peterzdeb, I have updated this PR significantly. Can you please check this again?

custom_model_runner/datarobot_drum/drum/entry_point.py
```diff
@@ -1,19 +1,54 @@
...
+            )
+
+        if server_type == DrumServerType.GUNICORN and getattr(options, "docker", None):
+            logger.warning(
+                "Gunicorn is not supported with --docker: gunicorn binds the host address "
+                "before the container starts, which conflicts with the container's own port "
+                "publish. Falling back to %s.",
+                DrumServerType.WERKZEUG,
+            )
+            server_type = DrumServerType.WERKZEUG
```
peterzdeb, line 46: Only a logger.warning marks there is no error, there is no  non-zero exit. Anyone relying on gunicorn's multi-worker behavior in a docker deployment gets werkzeug instead, silently. We might need to confider a hard fail on this.

peterzdeb (approved):

custom_model_runner/datarobot_drum/drum/entry_point.py
piratovskyi: @peterzdeb, I wouldn't do this. Usually, updating DRUM or execution environment is a manual action. If people do that they will see the problems during the rollout. It will never happen as a silent update. On the other hand, if folks used to use default server and they dont care about the type of the server - they will see an error after the upgrade. That will be confusing. I wouldn't do that

PR #2291, opened 2026-08-05
[CFX-7426] History should persist even after we shutdown and then start a session again for Codespaces/Notebooks

c-h-russell-walker:

## Summary

We have history persistence already in our other NBX (3.11 etc) EEs but was missed when we created 3.13 EE recently

## Rationale

This is expected behavior in our Codespaces

Seen here tested with this version (after confirming existing packaged 3.13 EE does _not_ have this after starting a session after shutdown)

Screenshot 2026-08-05 at 9 37 36 AM

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 6862b19fdc0dcef70647312d08516e726ccc9e16. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/public_dropin_notebook_environments/python313_notebook/env_info.json
+++ b/public_dropin_notebook_environments/python313_notebook/env_info.json
@@ -4,7 +4,7 @@
   "description": "This environment is the Python 3.13 Execution Environment that is present in DR Codespaces/Notebooks. It can also be used as a template to expand upon and create custom Python 3.13 notebook environments.",
   "programmingLanguage": "python",
   "label": "",
-  "environmentVersionId": "6a7110d5ab532f076d2d7580",
+  "environmentVersionId": "6a733d7482c8bc076d859def",
   "environmentVersionDescription": "",
   "isPublic": true,
   "isDownloadable": true,
@@ -15,8 +15,8 @@
   "contextUrl": "https://github.com/datarobot/datarobot-user-models/tree/master/public_dropin_notebook_environments/python313_notebook",
   "imageRepository": "env-notebook-python313-notebook",
   "tags": [
-    "v11.12.0-6a7110d5ab532f076d2d7580",
-    "6a7110d5ab532f076d2d7580",
+    "v11.12.0-6a733d7482c8bc076d859def",
+    "6a733d7482c8bc076d859def",
     "v11.12.0-latest"
   ]
 }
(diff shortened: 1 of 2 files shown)


c-h-russell-walker:
**UPDATE ON BELOW:**

I was able to update the Harness trigger for `base_checks` to ignore the `public_dropin_notebook_environments` directory since those unit tests have nothing to do with our NBX EEs

<hr>

Unit tests failures are unrelated - they're a failure in python 3.9:
`TypeError: unsupported operand type(s) for |: 'type' and 'type'`

I posted about it in #drum here:
https://datarobot.slack.com/archives/C01BAAV38FR/p1785939587652139

I am going to see what I can do about getting this NBX repo to _not require these unit tests_

c-h-russell-walker:
Got approval/confirmation that GenAi-Systems team does _not_ need/want to approve these NBX EE PRs - updated codeowners here:

https://github.com/datarobot/datarobot-user-models/pull/2292

ajalon1 (commented):
This LGTM, especially if we've already tested this in nbx-kernels. But what about `histappend`?

https://www.gnu.org/software/bash/manual/html_node/Bash-History-Facilities.html

> When a shell with history enabled exits, Bash copies the last $HISTSIZE entries from the history list to the file named by $HISTFILE. If the histappend shell option is set (see [Bash Builtin Commands](https://www.gnu.org/software/bash/manual/html_node/Bash-Builtins.html)), Bash appends the entries to the history file, otherwise it overwrites the history file. If HISTFILE is unset or null, or if the history file is unwritable, the history is not saved. After saving the history, Bash truncates the history file to contain no more than $HISTFILESIZE lines as described above.

c-h-russell-walker:
> This LGTM, especially if we've already tested this in nbx-kernels. But what about `histappend`?
> 
> https://www.gnu.org/software/bash/manual/html_node/Bash-History-Facilities.html
> 

Huh, thank you, interesting.

Since I took this from nbx-kernels I'll merge as-is since I'd prefer to not only keep relatively close parity but also it's been prod-tested for a long time in those EEs

https://github.com/datarobot/nbx-kernels/blob/main/kernels/setup-shell.sh

ajalon1 (approved):

PR #101, opened 2020-07-31
[RAPTOR-3204, RAPTOR-3205] Better errors when errors occur in hooks, model loading or on predict

rvorobii:
# This repository is public. Do not put here any private DataRobot or customer's data: code, datasets, model artifacts, .etc.

## Summary
When a model fails to load - error returned by error server is not informative enough.

This PR wraps exceptions from `custom.py` hooks, model loading and predictors' predict methods to add more context.
The error if the same type is re-raised, only with changed text.
The exceptions are chained to have original traceback and message.

### Example 1
```
{"message":"ERROR: Could not load model from artifact file because an error was raised: UnpicklingError('could not find MARK')"}
```
instead of
```
{"message":"ERROR: could not find MARK"}
```

### Example 2
```
{"message":"ERROR: An error was raised when making predictions: ValueError('X.shape[1] = 5 should be equal to 13, the number of features at training time')"}
```
instead of
```
{"message":"ERROR: X.shape[1] = 5 should be equal to 13, the number of features at training time"}
```

Tracebacks of the original exception is included in the resulting log.

# Example: load_model hook raises ValueError('exception')
### The error on UI
image
### The log with chained exceptions
```
2020-08-03 12:37:02,176 ERROR drum.prediction_server:  Model score hook failed to make predictions because an error was raised: ValueError('exception')
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/datarobot_drum/drum/model_adapter.py", line 284, in predict
    predictions = self._custom_hooks[CustomHooks.SCORE](data, model, **kwargs)
  File "/opt/code/custom.py", line 14, in score
    raise ValueError('exception')
ValueError: exception

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.7/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/local/lib/python3.7/site-packages/datarobot_drum/resource/components/Python/prediction_server/prediction_server.py", line 99, in predict
    out_df = self._predictor.predict(in_df)
  File "/usr/local/lib/python3.7/site-packages/datarobot_drum/drum/language_predictors/python_predictor/python_predictor.py", line 37, in predict
    predictions = self._model_adapter.predict(data=df, model=self._model, **kwargs)
[...]

PR #2308, opened 2026-08-13
[MODEL-24091] sandbox image: install folium for create_chart_panel map charts

j1z0:
## Why

`create_chart_panel` (datarobot-genai / wren-mcp) supports three charting libraries. Only folium needs a package in this image:

| library | user code returns | needs a package here? |
| --- | --- | --- |
| `plotly` | figure JSON dict | no |
| `altair` | Vega-Lite spec dict | no |
| `folium` | rendered leaflet HTML string | **yes** |

plotly and altair charts are plain JSON, so the tool has user code emit the dict directly and never imports either package. Leaflet HTML is not hand-authorable, so folium has to be installed — without it the folium path fails with `ModuleNotFoundError`.

This unblocks the altair/folium regression on [wren-mcp #101](https://github.com/datarobot/wren-mcp/pull/101), where the genai-server rebuild had collapsed `create_chart_panel` to plotly only.

## Why render in the sandbox rather than in the MCP server

Handing a spec out for the calling server to render would mean either:

- dropping `style_function` / `highlight_function` — Python **callables** that no JSON spec can carry, and the whole point of a choropleth; or
- `eval`-ing a code string in a shared, multi-tenant MCP process, which is strictly worse than the isolated container that already exists.

`Map.render()` is also a Jinja2 expansion that materializes the entire feature set into one in-memory HTML string — that belongs somewhere with a resource cap and a short lifetime.

## Cost

Negligible. folium's whole tree is pure Python:

| package | size |
| --- | --- |
| `folium` | 0.11 MB |
| `jinja2` | 0.13 MB |
| `xyzservices` | 0.09 MB |
| `branca` | 0.02 MB |
| `markupsafe` | 0.01 MB |
| **total** | **~0.36 MB** |

Against ~120 MB of numpy/pandas/polars/pyarrow already in the image (numpy and pandas arrive as `datarobot` dependencies). No new compiled extensions. altair was deliberately **not** added — it would pull 8 packages including compiled `rpds-py`, for a format user code can already emit as JSON.

## Verification

- `folium.Map(...).get_root().render()` on 0.20.0 returns a full `<!doctype html>` string containing `</head>` — the hook the consuming frontend's `FoliumChart` uses to inject its own styles — and is JSON-serializable.
- Confirmed `style_function` is accepted as a callable, which is the basis of the rationale above.
- The pins resolve against the existing constraint set on Python 3.12 (28 packages, no conflicts).
- The image build itself is exercised by the `on_pr` Harness trigger for this path. Merging republishes the mutable `_latest` tag, which is what consumers pull — there is no caller-side version to bump.

## Also in this PR

Corrects the README's "Source of truth" section, which claimed the Dockerfile and runner are mirrored from `docker/sandbox/` in datarobot-genai. That path does not exist in that repo — genai deliberately shares only the wire contract in `drtools/core/sandbox/protocol.py` so a duplicated runner can't drift from the image. The old note would have sent the next person editing this folder looking for a file that isn't there.

---

> [!NOTE]
> **Low Risk**
> Adds a small pure-Python dependency to an existing untrusted-code sandbox image with no changes to runner behavior, auth, or the wire protocol.
> 
> **Overview**
> Adds **`folium>=0.20.0,<1.0.0`** to the MCP execute sandbox image so **`create_chart_panel`** map charts can run user code that returns rendered Leaflet HTML (plotly/altair stay JSON-only and are not added).
> 
> The **Dockerfile** and **README** now document why folium is the only charting dependency and why rendering stays in the capped sandbox (choropleth callables, `Map.render()` memory use) instead of the shared MCP server.
> 
> README fixes: **source of truth** is this folder only (genai keeps the wire contract in `protocol.py`), and **published image** notes that merging republishes the mutable `_latest` tag with no caller version bump.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 091d90d65ddbe6ea7cf6e548edef2627537bd327. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/public_dropin_environments/dr_mcp_execute_sandbox_minimal/Dockerfile
+++ b/public_dropin_environments/dr_mcp_execute_sandbox_minimal/Dockerfile
@@ -3,7 +3,9 @@
 # Minimal sandbox runner image used by DataRobotWorkloadSandbox.
 #
 # Ships a standalone runner script (no datarobot-genai install needed
-# inside the container) plus the data libraries user code commonly needs.
+# inside the container) plus the data libraries user code commonly needs and
+# folium, the one charting library that cannot be replaced by emitting JSON
+# (see the install step below).
 #
 # ============================================================================
 #   *** USER CODE INSIDE THIS IMAGE IS HARD-CAPPED AT 1 HOUR (3600s) ***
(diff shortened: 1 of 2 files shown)


lkanggithub (approved):
LGTM

PR #2315, opened 2026-08-17
[RAPTOR-19443] Exec into gunicorn so its master inherits PID 1

piratovskyi:
# RATIONALE
Starting gunicorn via `subprocess.Popen` left drum as PID 1, forcing a drum-side signal forwarder and an extra process hop for exit-status propagation. Replacing that with `exec` makes the gunicorn master PID 1 directly, so signal handling and OOM-triggered master exits reach the container runtime natively.

## CHANGES
- `run_gunicorn.py` now `os.execve`s into `python -m gunicorn` instead of spawning it as a child and waiting.
- Dropped the `_install_signal_forwarder` relay and its forwarded-signal set — no longer needed once gunicorn owns PID 1.
- Rewrote `test_run_gunicorn.py` to cover the exec path (command/env/PYTHONPATH, missing-config, missing-gunicorn).
- Werkzeug path unchanged: it already runs in-process via `app.run()`, so drum is already PID 1 there — no subprocess to exec.

---

> [!NOTE]
> **Medium Risk**
> Changes container process model and shutdown/signal handling for gunicorn `drum server`; misbehavior could affect graceful worker cleanup, though it aligns with standard PID-1 semantics.
> 
> **Overview**
> **Gunicorn `drum server` startup** no longer spawns gunicorn as a child process. `main_gunicorn()` now **`os.execve`s** into `python -m gunicorn` after setting `DRUM_GUNICORN_DRUM_ARGS` and prepending the gunicorn package dir to `PYTHONPATH`, so the **gunicorn master is the container entry process** and gets runtime signals and exit status directly.
> 
> The **drum-side signal forwarder** (`_install_signal_forwarder` / `_FORWARDED_SIGNALS`) and **subprocess wait/exit-code relay** are removed as unnecessary. **Unit tests** were rewritten around the exec path (argv, env, missing config/gunicorn); the old signal-forwarder and end-to-end SIGTERM propagation tests are dropped.
> 
> **Werkzeug server mode is unchanged** (still in-process).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ef8028acebe10968824ee4bdbaf6e073941925e2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/custom_model_runner/datarobot_drum/drum/gunicorn/run_gunicorn.py
+++ b/custom_model_runner/datarobot_drum/drum/gunicorn/run_gunicorn.py
@@ -1,6 +1,4 @@
 import logging
-import signal
-import subprocess
 from pathlib import Path
 import sys
 import os
@@ -9,42 +7,6 @@ import shlex
 from datarobot_drum.drum.enum import LOGGER_NAME_PREFIX
 
 logger = logging.getLogger(LOGGER_NAME_PREFIX + "." + __name__)
-
-
-# Signals the orchestrator/runtime may send to the drum entry-point that must
-# reach the gunicorn master so it can run its graceful shutdown (which is what
-# fires each worker's worker_exit hook → ctx.stop() / ctx.cleanup()).
-_FORWARDED_SIGNALS = (
-    signal.SIGTERM,
-    signal.SIGINT,
-    signal.SIGHUP,
-    signal.SIGQUIT,
-    signal.SIGUSR1,
-    signal.SIGUSR2,
-)
-
-
-def _install_signal_forwarder(proc):
-    def _forward(sig, _frame):
-        # Relay first: logging is not async-signal-safe (a signal landing
-        # mid-write can raise a reentrancy error), so nothing may run before
-        # send_signal that could fail and swallow the relay.
-        try:
-            proc.send_signal(sig)
-        except ProcessLookupError:
-            pass
-        try:
-            logger.info("Forwarded signal %s to gunicorn master (pid=%s)", sig, proc.pid)
-        except Exception:
-            pass
-
-    for s in _FORWARDED_SIGNALS:
-        try:
-            signal.signal(s, _forward)
-        except (ValueError, OSError):
-            # ValueError: not the main thread; OSError: signal not available
-            # on this platform. Skip and let the default disposition stand.
-            pass
 
 
 def main_gunicorn():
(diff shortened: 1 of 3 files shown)


custom_model_runner/datarobot_drum/drum/gunicorn/run_gunicorn.py
```diff
@@ -81,22 +43,13 @@ def main_gunicorn():
         "app:app",  # module:variable; app.py sits next to this script
     ]
 
+    # Replace this process with gunicorn so its master runs as PID 1, receiving container signals and owning the exit status directly.
     try:
-        proc = subprocess.Popen(gunicorn_command, env=env)
+        os.execve(sys.executable, gunicorn_command, os.environ)
```
mariusvilkas, line 48: we have some tickets on 
"uWSGI intentionally overwrites sys.executable. When your application runs inside uWSGI, sys.executable points to the path of the uWSGI binary itself rather than the standard Python interpreter (like /usr/bin/python3)"
are you checking that this is a correct one?

mariusvilkas (approved):
Please recheck sys.executable. Otherwise
LGTM

custom_model_runner/datarobot_drum/drum/gunicorn/run_gunicorn.py
piratovskyi: @mariusvilkas, I lack the context here - why are you referring to uwsgi here? afaik, we don't use uwsgi in DRUM. Or am I missing something?

mariusvilkas: So the answer is - we do not check since we do not need?

piratovskyi: I guess so. We don't support uwsgi

custom_model_runner/datarobot_drum/drum/gunicorn/run_gunicorn.py
```diff
@@ -63,12 +25,12 @@ def main_gunicorn():
         except AttributeError:
             os.environ["DRUM_GUNICORN_DRUM_ARGS"] = " ".join(shlex.quote(a) for a in extra_args)
 
-    env = os.environ.copy()
-    current_pythonpath = env.get("PYTHONPATH", "")
+    package_dir_str = str(package_dir)
```
adamalpi, line 28: any reason why not using pathlib for these operations? it might be safer.

adamalpi (approved):

peterzdeb (approved):

custom_model_runner/datarobot_drum/drum/gunicorn/run_gunicorn.py
piratovskyi: @adamalpi, what do you mean? if not using a variable for stringtified package_dir then yeah it might be redundant. package_dir is already a Path instance
```package_dir = Path(__file__).resolve().parent```

adamalpi: Ahh sorry, didn't see it is a Path instance above. 

Then why not using:

`os.environ["PYTHONPATH"] = str(package_dir_str / current_pythonpath)`

instead of 

`os.environ["PYTHONPATH"] = f"{package_dir_str}{os.pathsep}{current_pythonpath}"`

adamalpi: ahh okay, PYTHONPATH is a path list, nvm, ignore this comment, I was wrong with this suggestion.

PR #372, opened 2021-06-29
[RAPTOR-5491] fix datatypes in

scottp-ml:
# This repository is public. Do not put here any private DataRobot or customer's data: code, datasets, model artifacts, .etc.

## Summary
This fixes the use of IN to match DR

PR #2378, opened 2026-09-08
Open a bump PR when pulumi-datarobot releases

cdevent:
## Summary

Adds a `repository_dispatch` receiver that opens a PR bumping the `pulumi-datarobot` pin in the Python 3.13 notebook environment whenever the provider publishes a release.

- `scripts/bump_pulumi_datarobot.sh` — verifies the version, bumps the pin, opens the PR
- `.github/workflows/bump-pulumi-datarobot.yaml` — the receiver, plus `workflow_dispatch` with an optional version input for re-runs and backfills

Provider side: datarobot-community/pulumi-datarobot#372. Sibling receivers: datarobot/nbx-kernels#382, datarobot-community/af-component-base#122.

## Rationale

`public_dropin_notebook_environments/python313_notebook/requirements.txt:27` pins `0.10.42`. PyPI is at `0.11.1`, released September 1st — four releases back. `python313_notebook` is the only environment in the repo that pins this package.

Reacting to the release beats polling for it, and notably Dependabot *cannot* poll fast enough anyway: it has no release webhook and enforces a hard 24-hour minimum interval, rejecting anything shorter outright (`Cronjob expression '0 */4 * * *' has a minimum interval of 4 hours which is less than the minimum allowed interval of 24 hours`).

Three properties worth calling out:

**The PR includes a fresh `environmentVersionId`.** Changing an environment's contents without one leaves the platform treating it as the same version. Harness has `update_env_version` input sets for the seven `public_dropin_environments` envs but **none for the notebook environments**, so nothing else does this for `python313_notebook` — the workflow runs the repo's own `tools/env_version_update.py`, the same script that pipeline runs.

**The dispatched version is verified, not trusted.** It must be installable from PyPI *and* cut as a non-draft, non-prerelease GitHub release before the pin is touched. The dispatch fires the moment the provider's SDK publish step finishes, and PyPI's index lags its own publish API, so the check retries briefly.

**Only the pin file and `env_info.json` are staged, by explicit path.** An incidental change elsewhere in the tree cannot ride along in a bump PR — verified below.

## Notes

- **Whether GitHub Actions is enabled on this repo is unverified.** It has no `.github/workflows/` directory today, and reading the Actions policy needs admin scope I don't have. If Actions is disabled, this workflow simply never runs — a visible no-op, not a broken build. Worth confirming before relying on it.
- Actions is used **only to open the PR**; builds stay on Jenkins/Harness. The bump PR's own checks come from the usual PR pipelines, which are webhook-driven and therefore unaffected by the rule that a PR opened with `GITHUB_TOKEN` doesn't trigger *Actions* workflows. The first PR it opens is the test of that.
- Independent of #2377 (the Dependabot config); either can land first. Keeping both is deliberate — the provider-side dispatch depends on a PAT, and a PAT expiring is silent, so the daily Dependabot check is the backstop. The de-dup here matches Dependabot's branch shape as well as this script's, so the two will never both open a PR for the same version.
- **The first PR it opens will be `0.10.42` → `0.11.1`.** On a `0.x` Pulumi provider a minor bump can carry breaking resource changes, so expect that one to need real review.

## Testing

`shellcheck` (via `uvx --from shellcheck-py`) and `actionlint` are both clean. The script was exercised against live PyPI and GitHub:

| Scenario | Expected | Result |
|---|---|---|
| `--check 0.11.1` (pin `0.10.42`) | verify + bump | pin rewritten, exit 0 |
| `--check` with no argument | resolve latest, bump | same, exit 0 |
| `--check '0.11.1 && id'` | reject | exit 1 |
| Pin line deleted | shape error | exit 1 |
| `--open-pr` with nothing pending | usage error | exit 1 |

`tools/env_version_update.py --file …` was run for real and produces a clean one-line diff — only `environmentVersionId`, nothing else in the JSON reformatted.

The staging behavior was tested against a deliberately dirtied tree: with an unrelated `uv.lock` modification present, staging the way `--open-pr` does picked up only the two intended files and left `uv.lock` behind. (That stray change came from running the version script via `uv run` during testing; the workflow uses plain `pip install pymongo`, which doesn't touch the lockfile.)

The provider-side dispatch payload was separately confirmed against the live API (`HTTP 204`) — see #372.

---

> [!NOTE]
> **Low Risk**
> This PR only adds CI automation and a script; it does not change the pin or notebook image contents until a separate auto-opened PR is merged.
> 
> **Overview**
> Adds **release-driven automation** so the Python 3.13 notebook pin for `pulumi-datarobot` stays current without waiting on Dependabot’s polling limits.
> 
> A new GitHub Actions workflow (`.github/workflows/bump-pulumi-datarobot.yaml`) listens for `repository_dispatch` type `pulumi-datarobot-released` from the provider repo and supports manual `workflow_dispatch` with an optional version. It runs `scripts/bump_pulumi_datarobot.sh`, which validates the target on **PyPI and GitHub** (with retries for index lag), updates the single pin in `public_dropin_notebook_environments/python313_notebook/requirements.txt`, and opens a **review-only PR**—never merges.
> 
> The workflow also regenerates `environmentVersionId` in `python313_notebook/env_info.json` via `tools/env_version_update.py`, matching what Harness does for other drop-in envs but not notebook envs. PR creation stages only those two paths and **de-dupes** against existing open PRs (including Dependabot branch naming) or closed-PR branches left on the remote.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63466fb997dcbb87c48362b9ac7d18c1a990f77c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/bump-pulumi-datarobot.yaml
+++ b/.github/workflows/bump-pulumi-datarobot.yaml
@@ -1,0 +1,91 @@
+---
+name: Bump pulumi-datarobot
+
+# Opens a PR bumping the pinned pulumi-datarobot version in the Python 3.13
+# notebook environment. Driven by the provider's release, not by polling:
+# pulumi-datarobot's release.yml dispatches `pulumi-datarobot-released` here
+# once its SDK is on PyPI. See scripts/bump_pulumi_datarobot.sh for the
+# skip/fail rules.
+#
+# This repo builds on Jenkins/Harness, not GitHub Actions; this workflow exists
+# only to open the bump PR. The PR's own checks come from the usual PR
+# pipelines, which are webhook-driven and so are unaffected by the rule that a
+# PR opened with GITHUB_TOKEN doesn't trigger *Actions* workflows. The first PR
+# this opens is the test of that.
+
+on:
+  # Sent by the release job in datarobot-community/pulumi-datarobot, with the
+  # released version in client_payload.version. Note that repository_dispatch
+  # only ever runs the copy of this workflow on the default branch.
+  repository_dispatch:
+    types: [pulumi-datarobot-released]
+
+  # Manual runs: re-run a dispatch that failed, or backfill a missed release.
+  # Leave the version empty to bump to whatever is currently newest.
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Version to bump to, e.g. 0.11.2 (empty = latest published)"
+        required: false
+        type: string
+
+permissions:
+  contents: write
+  pull-requests: write
+
+concurrency:
+  group: bump-pulumi-datarobot
+  cancel-in-progress: false
+
+jobs:
+  bump:
+    runs-on: ubuntu-latest
+    name: Bump pulumi-datarobot
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+
+      - name: Configure git identity
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+      # The version comes from the dispatch payload, or from the manual input,
+      # or is empty -- in which case the script resolves the newest release
+      # itself. It is verified against PyPI and GitHub either way.
+      - name: Check the released version and apply the bump
+        id: check
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RELEASED_VERSION: ${{ github.event.client_payload.version || github.event.inputs.version }}
+        run: scripts/bump_pulumi_datarobot.sh --check "$RELEASED_VERSION"
+
+      - name: Setup Python
+        if: steps.check.outputs.should_bump == 'true'
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      # Changing an environment's contents without a new environmentVersionId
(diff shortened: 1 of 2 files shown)


cdevent:
Closing in favour of datarobot-community/pulumi-datarobot#372, which now opens the bump PRs directly instead of dispatching an event for a workflow here to react to.

Why the change: opening the PRs from the provider repo with a PAT means one credential to configure and no bump workflow to keep in step across three repos. It also gives the bump PRs better CI — a PAT-opened PR is not subject to the rule that `GITHUB_TOKEN`-opened PRs do not trigger workflows, so each one runs this repo's normal checks.

Nothing is lost from this PR: the pin locations, version verification (installable on PyPI *and* cut as a non-draft GitHub release, with retries), de-dup against both its own and Dependabot's branch names, and the never-auto-merge behaviour all moved into `scripts/bump_downstream_pins.sh` in #372.
The `environmentVersionId` regeneration this PR added is preserved in #372: it runs this repo's own `tools/env_version_update.py` and commits `env_info.json` alongside the pin, since Harness has `update_env_version` input sets for the `public_dropin_environments` envs but none for the notebook environments.

The daily Dependabot config in #2377 is unaffected and still stands as a backstop.

